PR #283, opened 2015-04-15
Update prebuildStep.bat

ApertureSecurity:
The files when built have an underscore . I think that this is because despite having the file target name set to x32bridge.lib in MSC++2013 , it will place an underscore in the name. So I altered the  prebuild script when I realized the files weren't being copped over.


mrexodia:
hi,

thanks for the pull, but this is desired behavior, changed in 1a0f112b66d919bf22051823e59c4333f8c90fb4 (rename `x64_dbg` to `x64dbg`)

PR #344, opened 2015-07-18
afxres.h doesnt always exist on system (if you dont have MFC installed)

fileoffset:
using windows.h instead works


fileoffset:
Also fixed issues I had with pre/post build steps not being able to copy files. Wrapping path in quotes is also necessary if the project directory contains any spaces.

PR #346, opened 2015-07-19
Adding const-ness to various, and fixing potential mem leak with BPState

fileoffset:


x64_dbg_gui/Project/Src/Memory/MemoryPage.cpp
```diff
@@ -1,47 +1,44 @@
...
+bool MemoryPage::write(const void* parDest, uint_t parRVA, uint_t parSize) const
 {
-    bool ret = DbgFunctions()->MemPatch(mBase + parRVA, (unsigned char*)parDest, parSize);
+    bool ret = DbgFunctions()->MemPatch(mBase + parRVA, reinterpret_cast<const unsigned char*>(parDest), parSize);
     GuiUpdatePatches();
     return ret;
 }
 
-bool MemoryPage::write(const byte_t* parDest, uint_t parRVA, uint_t parSize)
+bool MemoryPage::write(const byte_t* parDest, uint_t parRVA, uint_t parSize) const
```
mrexodia: this isn't const :)

fileoffset: True! But because the underlying code doesn't modify any MemoryPage class members, it can still be set to const. 

Having the write() function as const or not in this instance is more of a design choice, I guess. It's up to you whether you think having this a non-const makes more sense from an interface perspective.

I'm happy to remove it from write() if you would prefer.

mrexodia: Yea, I think it's better to not allow writes when you have a "const
MemoryPage". Makes more sense to me.

Nukem9:
Personally I'd merge this, but it is @mrexodia 's repo mainly and I don't know what he thinks of the changes.

mrexodia:
I also want to merge, but the memwrite shouldn't be const. If it's updated,
please merge

PR #399, opened 2015-09-21
Fix memory alignment

mrgreywater:
Fix SSE dependent classes not being properly aligned when created on the
heap. 

vc was warning about it, and it can possibly lead to bugs when SSE instructions are executed on unaligned members


mrexodia:
Does this compile on VS13? I cannot test right now...

mrgreywater:
I checked it with both vc12 and vc14 in vs15, so it should be fine.

mrexodia:
Alright, I'll merge it as soon as possible.

PR #449, opened 2015-11-28
Herz3h dev

Herz3h:
Added Multiple Dump window with abiltiy to choose in what to dump to follow (choice available only in multidump window)

Added ability to search within module list in the Symbols tab

Added double click behavior in InfoBox : not the "best" way to do it as it relies on the text shown in the infobox ("Jump to..", last row double clicked = follow EIP in dump).

### DIFF
--- a/src/gui/Src/BasicView/HexDump.cpp
+++ b/src/gui/Src/BasicView/HexDump.cpp
@@ -46,7 +46,7 @@ void HexDump::updateFonts()
     setFont(ConfigFont("HexDump"));
 }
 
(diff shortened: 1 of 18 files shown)


mrexodia:
Hey, the Escape to clear the symbols didn't work for me and is a duplicate
for ctrl+backspace but I will test and see if it works now.

Herz3h:
Both are working here, but imo having to press ctrl + backspace is not intuitive, and not really that convenient/intuitive.

mrexodia:
I will check it out. Escape is really special in Qt

PR #453, opened 2015-12-01
DBG : Added ability to save/load command line arguments into the datab…

Herz3h:
…ase.

GUI : Symbols tab, when searching a module and getting no result, function widget gets cleared.


Nukem9:
You forgot `commandline.cpp/.h`

Herz3h:
Yeh indeed, mistake there.

PR #469, opened 2015-12-23
Drawing stuff should be faster + Fixed WordEdit and Calculator dialogs being unresponsive to inputs (no "conversion" done..)

Herz3h:


Herz3h:
Replacing repaint by update forces to call AbstractTableView::repaint() manually in case of say setting a breakpoint/bookmark/adding func/resizing columns etc...trying to fix that atm

mrexodia:
Quick sidenote I still don't understand this whole callback hell. I thought
the signals can handle that or is there custom behavior somewhere that
needs a special callback?

Herz3h:
Well the ValidateExpressionThread is used for validating expression in Calculator/GoTo Dialog aswell as Assemble Dialog (when Keep Size is checked), the way expression is validated can be different (for assemble dialog i have to check if its a valid instruction to get its size, again only when Keep Size is checked), that's why i use a callback.

That way it can be used to validate expressions on different conditions

mrexodia:
Ah okay, makes a lot of sense now!

mrexodia:
Is this branch ready to merge or are there still issues with the drawing?

Herz3h:
There is no more need to explicitely call repaints, so it's almost like before (except the calls to update instead of repaint() on the viewport which makes it faster than before), so it should be fine now. At least what i've tested which is disassembly widget/breakpoints/memory map and references..I don't think it breaks anything anymore when it comes down to drawing stuff

mrexodia:
Alright, do you want me to merge now or should I wait? Just tell me when to
merge :)

Merry Christmas by the way!

Herz3h:
Don't think i'll add anything atm, will switch to the graph branch and try to fix the crash thing (remove all ptrs stuff)

So you can merge it now if you think it's ok

mrexodia:
Alright, I'll merge this in development and merge development in graph.

mrexodia:
Gimme a few hours, the merge conflicts are real and I have Christmas dinner
:)

Herz3h:
Haha no problem take your time :D

mrexodia:
I merged it, please check it because something in a thread froze and
crashed x64dbg when I tested.

Herz3h:
What part caused the crash, development or graph?

mrexodia:
I don't know, I'm now working a little on the development branch and I
didn't see any crashes. I will check further.

Herz3h:
Oh so you merged dev into graph and you had the crashes kk

mrexodia:
Yea that's what happened :) please check my merge commit and see if I
messed up somewhere.

PR #476, opened 2015-12-25
Graph : Move controlFlowGraph into it's own class to allow for stuff …

Herz3h:
…to properly be deleted when reloading the program/loading a different program

Graph : Fixed two bugs causing crashes + some cleanup/refactor

### DIFF
--- a/src/dbg/_global.h
+++ b/src/dbg/_global.h
@@ -1,7 +1,7 @@
 #ifndef _GLOBAL_H
 #define _GLOBAL_H
 
-#define _WIN32_WINNT 0x0501
+#define _WIN32_WINNT 0x0601
 #define WINVER 0x0501
 #define _WIN32_IE 0x0500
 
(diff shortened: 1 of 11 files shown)


src/dbg/_global.h
```diff
@@ -1,7 +1,7 @@
 #ifndef _GLOBAL_H
 #define _GLOBAL_H
 
-#define _WIN32_WINNT 0x0501
```
mrexodia: yea this, don't do it. we need xp support

Herz3h: Well GetProcessDepPolicy requires it to be 0x600

Nukem9:
@Herz3h 
https://github.com/x64dbg/x64dbg/commit/5582730485f9851ce5a1f2436f3edd034a46468e

aka the hacky-but-compatible-way

mrexodia:
Yea we talked about it in Gitter, come join is 😀

PR #615, opened 2016-03-30
SEH Viewer for 32-bit

Avinm:
cbc6fec8-f366-11e5-8c90-4bc9bb2cbc03
d43aa780-f366-11e5-9bc1-9b4ce7395f6d

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/615)


mrexodia:
Great work!

I just checked, the call to `ExHandlerGetSEH` in `stackcommentget` is too much overhead (especially when there is more than 1 SEH entry). The main reason being that it is called ~20 times on each repaint. A possible solution is an `unordered_map` that gets populated in the `updateSEHChainThread` (making performance not matter) and then do lookups in this map in the `stackcommentget` function.

This requires a new SRW lock (threading.h). If you remove the part in `stackcommentget` I will accept the pull so people can have a dedicated SEH view to start with. Then the performance critical part can be updated later.

Avinm:
I'm wondering why `stackcommentget` is called that often? Also is there a function that's called only once when the stack is modified? I could update the map there then without bothering with the locks.

mrexodia:
You will always need locks because the map can be accessed from multiple
threads. Anyway stackcommentget is called on every redraw (which happens a
lot). See my previous comment for the location that is called when the
stack is modified (in the thread you create). I will try to make a code
example later this evening.

PR #629, opened 2016-04-19
add option to select code page in dump view

polelf:
DBG:
fix modLoad Chinese path bug
add HexDialog utf8 show
fix CPUDump to show Chinese chars

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/629)


mrexodia:
Thanks a lot for your work, but this pull request will not be accepted as is. There are definitely changes I like a lot in this however. Please keep it open until I manually merged those in.

Things in particular that will never be merged are changes to the project files (changing to Qt 5, upgrading to VS 15, changing release scripts, dependencies). Those changes will happen but at a different point in time and in coordination with the build servers and such.

The rest looks great and I will make an effort to get them merged in as soon as possible!

mrexodia:
Just a quick comment. I didn't forget about this pull request but since I disagree with certain changes I have to manually merge the changes I want for which I don't have much time at the moment.

mrexodia:
I added the unicode support for symbol loading, the other parts still have to be added manually.

torusrxxx:
How's the progress? I'm just unable to use x64dbg to debug my bunches of Chinese executables. +1

mrexodia:
Are you unable to load them or are you referring to strings not being recognized and such?

I didn't add the changes related to code pages yet and I'm currently not working on it. I also have some concerns regarding performance of the getFontWidth function. These being called in richTextPainter will probably drastically decrease single step performance.

mrexodia:
Today I implemented code page views in the HexDump. @polelf if this feature is okay for you, could you re-fork the project so I can merge your pull requests again?

polelf:
i pulled your changes,.i have some suggest,  
QChar ch(data.constData()[i]);
            if(ch.isPrint())
                text += ch.toLatin1();
            else
                text += '.'; 
Because the Chinese characters is two byte ,and not range of 0-127 
so ch. IsPrint () judgment cannot display Chinese characters , the utf8 is also such a problem, 
so i use QTextCodec fromUnicode toUnicode do the convert, your can see my code , Not quite right, can refer to

mrexodia:
The name `UTF-16` is confusing and I will change it. The only reason it exists is to print extended ASCII `XX 00` characters. If you want to print real UTF-16 you can use the `Text/Hex -> Codepage...` option and it will print the text per-row and properly.

See [here](https://github.com/x64dbg/x64dbg/blob/development/src/gui/Src/BasicView/HexDump.cpp#L586):

```
//This might produce invalid characters in variables-width encodings. This is currently ignored.
curData.text = desc.textCodec->toUnicode(QByteArray((const char*)wData, wBufferByteCount));
curData.text.replace('\t', "\\t");
curData.text.replace('\f', "\\f");
curData.text.replace('\v', "\\v");
curData.text.replace('\n', "\\n");
curData.text.replace('\r', "\\r");
richText.push_back(curData);
```

The reason I don't want to change the `ASCII/UTF-16` is that certain non-English characters have a variable width (or even different from the English characters), so it makes per-unit printing/selection a complete hell.

I forgot to add the UTF-8 conversion in the `HexEdit` dialog and I will also add your change with the process handles.

torusrxxx:
Maybe we can adopt UCS-2 here. UCS-2 is fixed-length encoding. We use width
that is twice the width of English letters. Most UCS-2 characters will
display well.

polelf:
QT is already a UCS-2, but debug target file coding are not UCS-2

mrexodia:
I do not want to change the current views, but obviously we can add new views that fit your needs.

mrexodia:
Added various features from this pull request in 424c69bc and 749d5262

polelf:
i refork x64dbg and change codepagedialog to a menu , you see if need to merge

mrexodia:
@polelf I like the CodePageDialog as a dialog. It now also saves the last selection so the difference with a menu is minor.

mrexodia:
Closed since all features were merged in. Please open a new pull request if you have new features.

PR #641, opened 2016-05-14
Fix UI bug in Setting dialog

blaquee:
Caused by Hi Resolution and Hi-DPI settings.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/641)


mrexodia:
thanks!

PR #642, opened 2016-05-18
Add the ability to set breakpoints directly on stack

torusrxxx:
It's a common practice to place hardware access breakpoints directly on stack, so that when the breakpoint is hit, the SFX code will stop right before OEP. I've added the breakpoint menu to the CPU stack's context menu so that it will be more convenient.
I'm a new comer, knew about x64dbg today and played with it a whole day. I'm starting from this easy task. I'm learning about x64dbg's architecture and may be contributing more code in future days.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/642)


blaquee:
Nice! looking forward to testing this merge out

mrexodia:
You can already do this through the registers window (right click -> hw break on esp)

Will probably merge it though (after I checked the code).

mrexodia:
thanks!

PR #643, opened 2016-05-20
x64dbg is translated to Simplified Chinese

torusrxxx:
I have translated x64dbg to Simplified Chinese. Batch file for generating translation template is also created.
x64dbg in simplified chinese
**The translations can be found in "src/gui/Translations" directory.** TS files in it can be opened using "Qt Linguist", and you can start translating every strings in it. After you've done, publish it in the form of QM files. I've included my version of x64dbg_zh_CN.qm as part of pull request. 
**If you want to translate x64dbg to another language**, edit "src/gui/x64dbg.pro" to include your target, then run "maketranslatetemplate.bat". Qt will generate TS files for you that you can work on.
**To use the translated QM files**, create a folder named "translations" under the "bin" directory or "release" directory. **It should reside along with "x32" or "x64" directories.** Then you download the QM file for the locale of your system, and copy it to the folder you've just created. x64dbg will automatically use that translation on restart. **There's currently no option to set current language.**

I suggest the x64dbg project create a translation project on sites like "Transiflex" so people who know nothing about Qt can join translating job. Translation templates should be updated frequently. And project contributors should be aware of internationalization from now on, use `tr` whenever he wants to display a string, and avoid internationalization-unfriendly string operations.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/643)


brainblowjob:
Nice PR +1

mrexodia:
Really liking this one! Will try to merge it as soon as possible. Do you
think it's possible to also have strings in the dbg be translated?

torusrxxx:
Not now. The bridge is currently unicode-unaware, preventing me from doing this. I've added a comment noting such problem in `bridgemain.h`. Will come back as soon as the bridge functions be improved.

mrexodia:
That is actually not true. The bridge is fully UTF-8 aware (or at least
almost there might be some places I missed because I don't speak any
languages that work with utf8).

I will check out this pull in a few hours and see if something can be done
with the dbg translation :)

torusrxxx:
Besides the problems in bridge, I use the internationalization mechanism come together with Qt. However the dbg have no interaction with Qt. So all texts in dbg will go through bridge to get their translations, which is problematic.

mrexodia:
merged it, thanks!

PR #645, opened 2016-05-20
Oops! Fix a compilation error introduced recently

torusrxxx:
mWindowMainTitle is set to `tr("x64dbg")` however it is not of type QString. #643

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/645)

### DIFF
--- a/src/gui/Src/Gui/MainWindow.h
+++ b/src/gui/Src/Gui/MainWindow.h
@@ -150,7 +150,7 @@ private:
     UpdateChecker* mUpdateChecker;
     TimeWastedCounter* mTimeWastedCounter;
 
-    const char* mWindowMainTitle;
+    QString mWindowMainTitle;
 
     QStringList mMRUList;
     int mMaxMRU;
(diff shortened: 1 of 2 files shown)


torusrxxx:
Some one have to test the translation branch recently merged. I find it not work after reinstalling snapshots from GitHub. I forgot to upload some files I've changed.

mrexodia:
i didnt test translation yet but i dont have a machine with chinese locale right now also probably the build scripts need to be updated to include the translation data right?

torusrxxx:
No, you don't need a system with Chinese locale. You can rename the x64dbg_zh_CN.qm into x64dbg_(locale).qm . It will trick x64dbg into thinking you've installed the translation for your current locale. I'm wondering if it's even nessesary to provide language in preferences.

mrexodia:
Alright, added some hidden settings to test translations. I'm Dutch myself but I would never use a Dutch version of x64dbg so it's nice to be able to change the default behavior of taking the system locale.

PR #656, opened 2016-05-23
Updated translation, and fixed an disturbing issue

torusrxxx:
In the Chinese version of x64dbg, if you reorder tabs then lots of message boxes will be poped, complaining it can't find config items. I addressed this issue in two ways:
1. Removed any message boxes from configuration.cpp and converted them into logs
2. Add a separate list for original native name, so it's possible to store tab orders in Chinese version of x64dbg.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/656)


src/gui/Src/Utils/Configuration.cpp
```diff
@@ -529,16 +526,10 @@ const bool Configuration::getBool(const QString category, const QString id) cons
...
-        msg.setWindowFlags(msg.windowFlags() & (~Qt::WindowContextHelpButtonHint));
-        msg.exec();
+        GuiAddLogMessage(QString("NOT FOUND IN CONFIG! %1:%2").arg(category).arg(id).toUtf8().constData());
         return false;
     }
-    QMessageBox msg(QMessageBox::Warning, "NOT FOUND IN CONFIG!", category);
-    msg.setWindowIcon(QIcon(":/icons/images/compile-warning.png"));
-    msg.setWindowFlags(msg.windowFlags() & (~Qt::WindowContextHelpButtonHint));
-    msg.exec();
+    GuiAddLogMessage(QString("NOT FOUND IN CONFIG! %1").arg(category).toUtf8().constData());
```
mrexodia: this will crash when an actual config string has not been found before the MainWindow is loaded (this was the reason these QMessageBoxes are here)

torusrxxx:
So what is the correct way of reporting these issues? I just dislike message boxes followed by message boxes.

mrexodia:
The native name should fix the message boxes so there is no need to remove
them. When everything works correctly they will not pop up anyway.

torusrxxx:
I'll replace these message boxes with Retry/Cancel message boxes. If cancel is hit then no message boxes will show and the error is not reported.

mrexodia:
also fine by me, they are just to nag you when you misconfigured something in AppearanceDialog/Configuration

torusrxxx:
Message boxes are equipped with Retry/Cancel button now.

Issue #654, opened 2016-05-23
Trace record


torusrxxx:
Trace record is the next major feature I'd like to introduce. When enabled on a page/module, it will record any location the executable have accessed/executed, and show they with different color in disassambly so you know you've been there before. It works only if you're stepping and no exception or thread switching occurs.

mrexodia:
Sounds cool! I'm interested to see your approach. Tell me if you need any
facilities that I can help with.

torusrxxx:
It was supposed to work by disassembling the current instruction before, though I have changed my mind and decided to implement it by defining a new type of memory breakpoint. It should work with previous restrictions removed.
But I have to fix another issue right now. If you're using a translated version of x64dbg and reordering tabs, it produces dozens of message boxes, which is very disturbing.

mrexodia:
I wouldn't use memory breakpoints for anything that needs performance. Placing a logger before you disassemble is probably much faster.

torusrxxx:
When testing current memory breakpoint capability I found memory breakpoint doesn't work. code is:
resume:
ADD R11,1
MOV RDI,table
MOVZX RAX,R11B
LEA RAX,QWORD PTR DS:[RAX+RDI]
XCHG RAX,RDI
MOV AL,BYTE PTR DS:[RDI]
ADD AL,1
MOV BYTE PTR DS:[RDI],AL
MOV RAX,QWORD PTR DS:[table]
JMP resume
table:
NOP...(256)
apply a memory breakpoint to all these and see it.

mrexodia:
Memory breakpoint works but you can only break on page sized buffers.

torusrxxx:
Oh no! Olly's memory breakpoint can be set to any buffer. Olly will bypass false hits automatically.

torusrxxx:
I'll defer the work until both memory breakpoint solution and disassemble solution are compared further.
Trace record data is useful not only in UI, but also in analysising. You can first run a "coverage" run,labelling every executable instructions and during analysising or next run, simply treat all instructions not executed before as junk bytes. I won't forget it.

mrexodia:
Yes I know. I plan to implement more flexible memory breakpoints in GleeBug.

[...]

PR #665, opened 2016-05-25
Trace record

torusrxxx:
#654

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/665)


torusrxxx:
The hit counter is recorded, however is not displayed in the GUI, nor is there any way to read them out except reading the database. That will be done in subsequent pull requests. My local branch is really outdated.

torusrxxx:
Trace record is crashing when the instruction crosses page boundary.

src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -358,12 +358,30 @@ void CPUDisassembly::setupRightClickContextMenu()
...
         return DbgMemIsValidReadPtr(addr);
     });
     mMenuBuilder->addMenu(makeMenu(QIcon(":/icons/images/label.png"), tr("Label")), labelMenu);
 
+    QAction* traceRecordDisable = makeAction(tr("Disable"), SLOT(ActionTraceRecordDisableSlot()));
+    QAction* traceRecordEnableBit = makeAction(tr("Bit"), SLOT(ActionTraceRecordBitSlot()));
+    QAction* traceRecordEnableByte = makeAction(tr("Byte"), SLOT(ActionTraceRecordByteSlot()));
+    QAction* traceRecordEnableWord = makeAction(tr("Word"), SLOT(ActionTraceRecordWordSlot()));
+    //TODO: I can't get mMenuBuilder work without an icon. Please add an icon for trace record. --torusrxxx
+    mMenuBuilder->addMenu(makeMenu(QIcon(":/icons/images/breakpoint.png"), tr("Trace record")), [ = ](QMenu * menu)
```
mrexodia: if you want menus without icon, you can use:

```
mMenuBuilder->addMenu(makeMenu(tr("Trace record")), [ = ](QMenu * menu)
{
}
```

mrexodia:
I can manually merge it if you want, the only conflict is in DBGFUNCTIONS probably.

torusrxxx:
So I'll add these functions in the middle of `DBGFUNCTIONS` to avoid conflicting. This feature have to be tested. I've just fixed an crashing condition within it.

src/dbg/TraceRecord.h
```diff
@@ -0,0 +1,89 @@
...
+
+    struct TraceRecordPage{
+        void* rawPtr;
+        duint rva;
+        TraceRecordType dataType;
+        unsigned int moduleIndex;
+    };
+
+    //Key := page base, value := trace record raw data
+    std::map<duint, TraceRecordPage> TraceRecord;
```
mrexodia: for this you can use `ModuleHashFromAddr(addr & ~(PAGE_SIZE - 1)) + addr & (PAGE_SIZE - 1)`

torusrxxx:
Leave the icon as is. An icon will be finally added.

src/dbg/TraceRecord.cpp
```diff
@@ -0,0 +1,445 @@
...
+        ModuleNames.push_back(moduleName);
+        return ModuleNames.size() - 1;
+    }
+}
+
+void _dbg_dbgtraceexecute(duint CIP)
+{
+    if(TraceRecord.getTraceRecordType(CIP) != TraceRecordManager::TraceRecordType::TraceRecordNone)
+    {
+        Capstone disassembler;
```
mrexodia: Creating this outside of this function with a lock will speed up a lot (internally `Capstone` allocates memory so this kills performance for tracing).

src/dbg/TraceRecord.cpp
```diff
@@ -0,0 +1,445 @@
...
+        unsigned char buffer[16];
+        duint size;
+        if(MemRead(CIP, buffer, 16))
+        {
+            TraceRecord.increaseInstructionCounter();
+            disassembler.Disassemble(CIP, buffer);
+            size = disassembler.Success() ? disassembler.Size() : 1;
```
mrexodia: You could also use `DbgDisasmFastAt` if you just need the `disassembler.Size()` this should be an optimized function.

src/dbg/TraceRecord.h
mrexodia: Nevermind, seems like you already do something like that :+1:

mrexodia:
You could use the trace icon `:/icons/images/trace.png` :smile:

torusrxxx:
It seems the conflicts may origin from GUI.

torusrxxx:
I don't know what to do now. May be it can only be merged manually. Thanks, @mrexodia.

mrexodia:
You should repull and test if everything is working properly, manual merges sometimes turn out real weird. It compiles and runs and it makes some instructions green when I select trace record -> bit.

mrexodia:
Thanks a lot for your contributions once again by the way! :tada:

torusrxxx:
Can I reinstall the latest branch from development now?

mrexodia:
Yes

src/dbg/TraceRecord.cpp
torusrxxx: Oh no. When I'm digging into the source, it turns out that DbgDisasmFastAt calls into `disasm_fast.cpp:bool disasmfast(const unsigned char* data, duint addr, BASIC_INSTRUCTION_INFO* basicinfo)`, where there is a Capstone object created to disassemble the instruction. Besides, that function doesn't handle the corner case where address is near the end of the page, and finally the returned boolean flag is dropped in the call chain.

mrexodia: I think Capstone wasn't a problem anymore because we used an optimized
disasm function.

torusrxxx: So the current code reads the current instruction, handles the corner case correctly, and without using that buffer it just calls the same routine which is not aware of close-to-page-end condition. A boring code duplication and waste.

mrexodia: I don't quite understand the close to page condition. If you use it
correctly you copy min(16, sizeLeftInPage) to the 16 byte buffer which you
pass to the disasm function. Since disasm length it never bigger than 16
there should never be any problems.

mrexodia: But yeah I think the old code should be restored for that indeed. No need
for DisasmFastAt

PR #670, opened 2016-05-27
GUI: Fixed HexEditDialog behavior

### LINKED ISSUE #367: Wrong behavior in binary edit window
1- Select a sequence of bytes in the disassembly / dump window
2- Right click -> Binary -> Edit
3- In the ASCII or the UNICODE text box type a character (in the middle or beginning)
4- The typing cursor immediately jumps to the end of the text box

If this is not a wrong behavior (this is how x64dbg works) could you please change it
so that it jumps to the next character like any text editor ?
It gets really annoying especially if you want to edit a long string.

EDIT: after some more tests, it also seems to be pushing the characters to the right
1- Edit for example a string of 5 characters / bytes (example: `"Hello"`)
2- Make sure the switch __`Keep size`__ is selected and place the cursor exactly before the
3rd char
3- Type any char (example: `Z`, so the new string would be `"Hezll"`)
(you can already see at this step that the first 'L' was not replaced but instead pushed to the right)
4- Now place the cursor at the beginning of the string and type another char (example: `X`,
so the new string is: `"XHezl"`)
5- You can clearly see that the char 'H' was not replaced with `X`
but instead the whole string got pushed to the right.

Atvaark:
I changed the behavior of the binary edit dialog slightly to make it less painful to edit text with it.

1. Prevented the position cursor from jumping to the end of the line when inserting / deleting text.
2. 'Keep size' will set the ASCII / Unicode text box to overwrite mode. This prevents characters from getting pushed out of the text boxes.

Fixes #367 .

A cleaned solution would be to create a custom QLineEdit widget that displays a QByteArray as ASCII or Unicode and supports overwrite mode.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/670)

### DIFF
--- a/src/gui/Src/Gui/HexEditDialog.cpp
+++ b/src/gui/Src/Gui/HexEditDialog.cpp
@@ -96,6 +96,21 @@ void HexEditDialog::on_btnUnicode2Hex_clicked()
 void HexEditDialog::on_chkKeepSize_toggled(bool checked)
 {
     mHexEdit->setKeepSize(checked);
+    QByteArray data = mHexEdit->data();
+    if(checked)
+    {
+        int dataSize = data.size();
+        QString asciiMask = QString("x").repeated(dataSize / sizeof(char));
+        QString unicodeMask = QString("x").repeated(dataSize / sizeof(wchar_t));
+        ui->lineEditAscii->setInputMask(asciiMask);
+        ui->lineEditUnicode->setInputMask(unicodeMask);
+    }
+    else
+    {
+        ui->lineEditAscii->setInputMask("");
+        ui->lineEditUnicode->setInputMask("");
+        mHexEdit->setData(data);
+    }
 }
 
 void HexEditDialog::dataChangedSlot()
@@ -126,11 +141,15 @@ void HexEditDialog::dataChangedSlot()
 void HexEditDialog::on_lineEditAscii_textEdited(const QString & arg1)
 {
     Q_UNUSED(arg1);
+    int cursorPosition = ui->lineEditAscii->cursorPosition();
     on_btnAscii2Hex_clicked();
+    ui->lineEditAscii->setCursorPosition(cursorPosition);
 }
 
 void HexEditDialog::on_lineEditUnicode_textEdited(const QString & arg1)
 {
     Q_UNUSED(arg1);
+    int cursorPosition = ui->lineEditUnicode->cursorPosition();
     on_btnUnicode2Hex_clicked();
+    ui->lineEditUnicode->setCursorPosition(cursorPosition);
 }


mrexodia:
Cheers!

mrexodia:
reverted because it introduced #675

Atvaark:
Found the issues.

The Qt documentation said that the 'x' character in a mask means **Any character permitted but not required.**. But I must have read over that blank characters are removed from masked inputs after every change. And there was also an off by one error when editing an uneven amount of bytes.

This means that a custom QLineEdit is needed to fix both errors #367. 
(The *jumping to the end of the text box* error can still be fixed with this:)
```
void HexEditDialog::on_lineEditAscii_textEdited(const QString & arg1)
{
    Q_UNUSED(arg1);
    int cursorPosition = ui->lineEditAscii->cursorPosition();
    on_btnAscii2Hex_clicked();
    ui->lineEditAscii->setCursorPosition(cursorPosition);
}

void HexEditDialog::on_lineEditUnicode_textEdited(const QString & arg1)
{
    Q_UNUSED(arg1);
    int cursorPosition = ui->lineEditUnicode->cursorPosition();
    on_btnUnicode2Hex_clicked();
    ui->lineEditUnicode->setCursorPosition(cursorPosition);
}
```

mrexodia:
So there should be a custom `QLineEdit` that takes a `QByteArray` as data and visualizes/edits it?

Atvaark:
Yeah, that should do it.

This is the interface for the QLineEdit I'm working on now. I have also moved all the text handing from the HexEditDialog into the new widget.
See: [HexEditDialog.cpp](https://gist.github.com/Atvaark/19c151f35af261c52986c40f3468b450)

```
class HexLineEdit : public QLineEdit
{
    Q_OBJECT

public:
    enum class Encoding
    {
        Ascii,
        Unicode
    };

    explicit HexLineEdit(QWidget *parent = 0);
    ~HexLineEdit();

    void keyPressEvent(QKeyEvent* event);

    void setData(const QByteArray & data);
    QByteArray data();

    void setEncoding(const Encoding encoding);
    Encoding encoding();

    void setKeepSize(const bool enabled);
    bool keepSize();

    void setOverwriteMode(bool overwriteMode);
    bool overwriteMode();
signals:
    void dataEdited();

private slots:
    void updateStyle();
    void updateData(const QString & arg1);

private:
    Ui::HexLineEdit *ui;

    QByteArray mData;
    Encoding mEncoding;
    bool mKeepSize;
    bool mOverwriteMode;
    int mMaxLength;

};
```

Got any remarks? I haven't worked with Qt before so I might have missed something.

Now I only have to figure out how to overwrite the keyPressEvent handler correctly, so that I can toggle between insert mode and overwrite mode.

Atvaark:
Text editors visualize the mode they are in.
In insert mode a blinking beam will be displayed and an in overwrite mode either a blinking line or a blinking rectangle is used to display which character will be overwritten next.

mrexodia:
Ohh, that shouldn't be a problem I think.

PR #688, opened 2016-05-31
db corruption is greatly reduced

torusrxxx:
In the crash dump I noticed multiple debug threads. This is quite strange since It's assumed only one process can be debuggged at a time. So I added a mutex locking the debug function.
Now if you holds CTRL+F2 down, the debugger will enter a wrong state, but the DB should not corrupt, and the program crashes less often.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/688)


mrexodia:
Thanks! I hope it will work!

PR #698, opened 2016-06-02
Allow resize of goto dialog

torusrxxx:
The dialog can already be resized automatically, but resizing was disabled before. Allowing resizing partially solves HIDPI problem.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/698)


blaquee:
If this works, I'd like to get it done for the Settings Dialog as well. Might help with the hi DPI displays

PR #703, opened 2016-06-03
Make register view larger in x64 mode

### LINKED ISSUE #701: Register view is too small to display YMM8~YMM15
x64dbg-register-window-too-small

torusrxxx:
Fix #701

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/703)


mrexodia:
What does this look like if you compare it with the current view?

torusrxxx:
The only difference is that the register view can scroll down more, revealing the content behind. And some space behind DRX.

torusrxxx:
x64dbg-register-view-extended

mrexodia:
it looks like this for me without this will is not show all the data on your side?

img

torusrxxx:
Different systems have different default font sizes.

mrexodia:
So I guess a proper fix would be to have this size be something like `fontHeight * x`?

mrexodia:
resolved in c4234d3

Issue #675, opened 2016-05-28
Wrong behavior of "Keep Size" button when selection is not a multiple of 4


wk-952:
1. Debug anything and edit a string from the dump window,
but make sure the number of selected bytes/chars is not a multiple of 4 (select 7 chars for example)
2. Without editing the string or doing anything, just press the `Keep Size` button
and the string will be weirdly reconstructed. (btw, the no. of chars in this pic is 31)

before:
![1](https://cloud.githubusercontent.com/assets/11916365/15623919/e4ea61e2-247b-11e6-8ca5-acbccef9795b.png)

after:
![2](https://cloud.githubusercontent.com/assets/11916365/15623924/f4376122-247b-11e6-86d2-1f72bedd8295.png)

If you selected 4, 8, 12, etc... chars, this issue won't be triggered.

wk-952:
This is considered an invalid issue now, since the commit was reverted.

PR #704, opened 2016-06-03
GUI: Implemented a custom QLineEdit

Atvaark:
I'm done with the cleaner version of PR #670.

- Fixes the ascii / unicode line edit behavior (Bug #367).
- Allows overwrite mode by pressing the INSERT key.
- Prevents immediately converting the unicode text back to hex (Bug #675)

Initializing the data the dialog should display is a bit odd right now. It is still done by calling *hexEdit.mHexEdit->setData(...);*, but it can be done just a single time (which should be fine right now, since *setData* is only called 0 or 1 times on each HexEditDialog instance).
I'd have to set the *mHexEdit* field to private and add some of the the QHexEdit api to the HexEditDialog to fix this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/704)


mrexodia:
looking great, thanks!

Issue #669, opened 2016-05-27
Supress all Gui message in plugin


wynick27:
I was trying to write a analyze plugin for vmprotect. The plugin sets hundreds of breakpoints and trace every vm instruction. It only took several seconds for the analyze to finish, but the gui thread takes several minute to respond to all related update view messages. I think it would be nice if there is an option to supress all message when the plugin is running.

mrexodia:
I believe somebody added an option to disable the log, you could try that.

torusrxxx:
You may minimize the window as a workaround, so that all the GUI drawing will not happen, speeding up the plugin.

mrexodia:
Assumed to be fixed, comment if you want to reopen this issue.

PR #710, opened 2016-06-04
Added api to supress gui message + add set breakpoint on every command in reference view

wynick27:
I implemented feature I mentioned in #669. After disabling gui messages, scripts and plugin commands that sets a lot breakpoints should run at least 5x faster.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/710)

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -13,6 +13,11 @@ static HINSTANCE hInst;
 static Utf8Ini settings;
 static wchar_t szIniFile[MAX_PATH] = L"";
 static CRITICAL_SECTION csIni;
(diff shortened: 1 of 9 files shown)


wynick27:
Log should also support grouping like Debug, Script, Plugin and level like Info, Warning and Error so that we can filter log messages. But this requires all call to dputs and dprintf changed.

wynick27:
Added api to supress gui message + add set breakpoint on every command in reference view
This operation is extremely slow (several minutes for a module with 10000 calls). That's again because ui is updated every time bp command is executed.
capture

mrexodia:
Thanks!

Issue #686, opened 2016-05-31
Export patches as C program source


torusrxxx:
Sometimes we need to distribute patches to users who have no access to debuggers, or to patch an executable quickly via a simple double-click.

mrexodia:
There is already an export option that tools can use to generate their own custom formats.

torusrxxx:
I'd like to leave this issue as a reminder. I would like to introduce this feature as a core feature instead of as a plugin.

mrexodia:
Okay :+1: As another idea it might be cool to be able to export patches in a template format, where you have format variables `{module}` `{va}` `{rva}` `{fileoffset}` `{oldbyte}` `{newbyte}` and you specify a template like:

Prefix (empty in my example):
```

```

Patch:
```
{rva}:{oldbyte}->{newbyte}
```

Suffix (empty in my example):
```

```

My example would result in the x64dbg patch export format. This is flexible and at the same time doesn't bother us with language specific features.

torusrxxx:
Finished. Pull it now.
```
This template exports patch as C program source.
$TEMPLATE_PREFIX:
/********************************************
 *         x64dbg generated patch           *
 ********************************************/

#include <windows.h>
#include <stdio.h>
#include <tlhelp32.h>
#include <psapi.h>

HANDLE findProcess();
void applyPatch(HANDLE hProcess, const wchar_t* moduleName, char* baseAddress);

int main()
{
    HMODULE hMods[2048];
    DWORD cbNeeded;
    unsigned int i;
    HANDLE hProcess;
    printf("Patch for $exeName\n");
    printf("This patch is generated by x64dbg(http://x64dbg.org) at $date.\n");
    printf("Start patching...\n");
    hProcess = findProcess();
    if(hProcess == NULL)
    {
        printf("Finished, but the target process is not found. No patch is applied.\n");
        return 0;
    }
    printf("Apply $numPatches patches.\n");
    if(EnumProcessModules(hProcess, hMods, sizeof(hMods), &cbNeeded) && cbNeeded < sizeof(hMods))
    {
        for(i = 0; i < cbNeeded / sizeof(HMODULE); i++)
        {
            wchar_t modName[MAX_PATH];
            if(GetModuleBaseNameW(hProcess, hMods[i], modName, sizeof(modName) / sizeof(char)))
            {
                applyPatch(hProcess, modName, (char*)hMods[i]);
            }
        }
    }
    else
    {
[...]

PR #712, opened 2016-06-04
Export patches as C program source

torusrxxx:
#686.
Template is implemented, but is embedded in the source of x64dbg in the form of raw string now. If someone comes about a new template, then maybe we have to load them dynamically from file.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/712)


src/gui/Src/Gui/PatchDialog.cpp
```diff
@@ -682,8 +684,228 @@ void PatchDialog::saveAs1337(const QString & filename)
...
+    msg.setParent(this, Qt::Dialog);
+    msg.setWindowFlags(msg.windowFlags() & (~Qt::WindowContextHelpButtonHint));
+    msg.exec();
+}
+
+void PatchDialog::saveAsC(const QString &filename)
+{
+    QString template_C;
+    // Loading template
+    template_C = QString(R"xxx(
```
mrexodia: Is this supported by Visual Studio 2013? My preference would be to load templates from a file that is not contained in the x64dbg source code directly (we can create a separate repository for it and include it as a submodule). I also think there should be a separate button for saving with a template so the current behavior is unchanged.

torusrxxx:
I have built it successfully.VS2013 professional.

mrexodia:
Sorry, but I don't want to accept the pull request in its current form. I will check it out and see how it will fit in.

mrexodia:
This is till on my todo list, will try to find time for it.

mrexodia:
Implemented as a plugin: https://github.com/torusrxxx/x64dbgpatchexporter

Issue #722, opened 2016-06-05
Wrong coloring and spacing of instructions in CPU tab (GUI only)


wk-952:
- To reproduce this issue, set the font of the `Disassembly` to anything bigger than 8 (or less than that).
- Using latest snapshot `snapshot_2016-06-05_09-44.7z`

Notice how the 2 words `ptr` and the segment register are concatenated,
also the immediate memory address is not completely painted (or excessively painted if font < 8).
Increasing or decreasing the font produces different output.
pic 1:
![err](https://cloud.githubusercontent.com/assets/11916365/15808742/f1b8171a-2b7f-11e6-99e3-306e61591f77.jpg)

pic 2:
![err2](https://cloud.githubusercontent.com/assets/11916365/15808753/1d5636fe-2b80-11e6-88e9-03d1f8d2c5c4.jpg)

wk-952:
@mrexodia I wanted to start testing the debugger again today regarding the db issue,
so despite this GUI bug, is it ok to use this snapshot for testing or should i wait ?

mrexodia:
You can try the one of two days ago. I rewrote some of the painting code
but I'm going to sleep now so I can't fix it.

torusrxxx:
I know what's happening. I have fixed it on my local branch.

mrexodia:
should be fixed now

blaquee:
This is not fixed. Just used latest snapshot and having the highlighting issues.
Also the Registers view still suffers from the clipping of the addresses.
<img width="738" alt="clipping" src="https://cloud.githubusercontent.com/assets/1324227/15829558/412190da-2be2-11e6-91ee-8cb498c9ac84.PNG">

mrexodia:
Did you use the sourceforge or the github snapshot? They are always
slightly out of sync.

I didn't touch the registers view as far as I know...

blaquee:
I always use the github snapshot

wk-952:
I tried this issue now on `snapshot_2016-06-06_06-15` and it it's partially solved (although the first post didn't describe the following).

This is a picture in which all fonts are set to `12`: (all is good)
![1](https://cloud.githubusercontent.com/assets/11916365/15863256/890acf42-2cd2-11e6-8fe4-abc80a122962.png)

This is another picture with all fonts set to `6`: (all is good too)
![2](https://cloud.githubusercontent.com/assets/11916365/15863280/a2af504e-2cd2-11e6-9220-094ee782be49.png)

[...]

PR #723, opened 2016-06-06
Invalidate font size cache

torusrxxx:
Solving issue #722 
And add special handling of surrogates.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/723)

### DIFF
--- a/src/gui/Src/BasicView/Disassembly.cpp
+++ b/src/gui/Src/BasicView/Disassembly.cpp
@@ -116,8 +116,7 @@ void Disassembly::updateColors()
 void Disassembly::updateFonts()
 {
     setFont(ConfigFont("Disassembly"));
-    delete mFontMetrics;
-    mFontMetrics = new CachedFontMetrics(this, font());
+    invalidateCachedFont();
 }
 
 void Disassembly::tokenizerConfigUpdatedSlot()
--- a/src/gui/Src/BasicView/Disassembly.cpp
+++ b/src/gui/Src/BasicView/Disassembly.cpp
@@ -116,6 +116,8 @@ void Disassembly::updateColors()
 void Disassembly::updateFonts()
 {
     setFont(ConfigFont("Disassembly"));
+    delete mFontMetrics;
+    mFontMetrics = new CachedFontMetrics(this, font());
 }
 
 void Disassembly::tokenizerConfigUpdatedSlot()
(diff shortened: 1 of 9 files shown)


src/gui/Src/BasicView/Disassembly.cpp
```diff
@@ -116,6 +116,8 @@ void Disassembly::updateColors()
 void Disassembly::updateFonts()
 {
     setFont(ConfigFont("Disassembly"));
+    delete mFontMetrics;
```
mrexodia: This should call the base AbstractTableView::updateFonts

src/gui/Src/Utils/CachedFontMetrics.h
```diff
@@ -19,6 +19,14 @@ class CachedFontMetrics : public QObject
     int width(const QChar & ch)
     {
         auto unicode = ch.unicode();
+        if(unicode >= 0xD800)
```
mrexodia: What is this for?

torusrxxx: This is for surrogates. Do you know the details of UTF-16 encoding? Maybe you want to look at [https://en.wikipedia.org/wiki/UTF-16](https://en.wikipedia.org/wiki/UTF-16)

src/gui/Src/BasicView/Disassembly.cpp
torusrxxx: No, `updateFonts` is a virtual function which is already overridden.

mrexodia: Yes that's the whole point of virtual functions. If you call the base class
function you don't need to duplicate the delete mFontMetrics line (actually
not sure why I don't call the base class here).

torusrxxx: Probably I can move that into `AbstractTableView` as a separate function, but not in `updateFonts` because there is another `setFont` to set the font for generic table views there.

torusrxxx: I have changed AbstractTableView a lot on my local branch. A new "Edit Columns” dialog is introduced to allow users to reorder/hide columns in a table view. That have to be pushed together with this pull request now. Just wait a moment please.

mrexodia: Hm right, then it would need a separate function to invalidate the cache...

torusrxxx:
I think I have committed all the files now.
x64dbg-hide-columns

mrexodia:
Looks good, thanks a lot!

PR #726, opened 2016-06-06
GDB: fixed thread sometimes switch to other when trying pause thread

yuxuanchiadm:
This often happened when current in a hardware/software interrupt. And other thread reached the breakpoint first.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/726)

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -403,7 +403,10 @@ static bool getConditionValue(const char* expression)
 
 void cbPauseBreakpoint()
 {
-    hActiveThread = ThreadGetHandle(((DEBUG_EVENT*)GetDebugData())->dwThreadId);
+    if(hActiveThread != ThreadGetHandle(((DEBUG_EVENT*)GetDebugData())->dwThreadId))
+    {
+        return;
+    }
     auto CIP = GetContextDataEx(hActiveThread, UE_CIP);
     DeleteBPX(CIP);
     GuiSetDebugState(paused);


yuxuanchiadm:
And we should to build some logic(like lock the UI down and showing user a message) to prevent UI do other thing(like force switch to other thread) when trying pause thread or do other sync needed job.

mrexodia:
Another way would be to suspend all threads and only resume them when the pause breakpoint is it right?

yuxuanchiadm:
But if current thread is waiting for another thread's semaphore. If we suspend all threads. Whole program will deadlock.

mrexodia:
Yes that's a good point. However I'm not sure a GUI is needed also this change seems weird to me. Basically you 'lie' to the user about which thread reached the breakpoint first. The hActiveThread is only for the GUI.

yuxuanchiadm:
Or we must determine if current thread is not able to allocate CPU time slice. And we should set current thread in "Pause" status. If we resume it before thread can run. We just set it to "Running" status.

mrexodia:
Whatever works best for you, I don't really use the pause feature so I prefer a simple solution like this. The whole point is to break the debuggee so you can interact with it right?

yuxuanchiadm:
Ok. how about this:
`enum THREAD_SUSPEND_STATUS
{
    RUNNING = 0,
    PAUSED = 1;
};

struct THREAD_SUSPEND_STATUS_RECORD
{
    DWORD dwThreadId;
    THREAD_SUSPEND_STATUS status;
    ULONG_PTR ulpCIP;
    bool bReached;
};

vector<THREAD_SUSPEND_STATUS_RECORD> threadSuspendStatusRecord;`

* Suspend all thread.
* Set current thread in "PAUSE" mode.
* Remember current thread IP in ulpCIP.
* Notify UI current thread in "PAUSE" mode. (Showing "Paused")
* SetBPX at current thread CIP.
* Resume all thread.
If:
* User switching to other thread -> Notify UI thread changed(If switched thread is pause so UI showing paused. Otherwise running).
* Thread reached pause breakpoint -> If status == PAUSED && bReached == false. DeleteBPX and bReached = true
* Thread reached pause breakpoint. But not current thread -> If status == PAUSED && bReached == false. DeleteBPX, bReached = true and suspend this thread(In cbPauseBreakpoint). If UI switched this thread to current thread. Continue debug.
* CIP already have breakpoint. -> Ignore SetBPX. If a breakpoint remove. Check threadSuspendStatusRecord. If some thread paused and bReached == false && breakpoint address == ulpCIP. SetBPX at CIP.
* Thread reached normal breakpoint. Check if some thread paused and bReached == false && breakpoint address == ulpCIP. bReached = true.
* Thread dead -> Remove suspend record.

mrexodia:
I think it's far too complex. I don't see the problem with the current implementation.

yuxuanchiadm:
There is some problem with current implementation.
* If thread is calling WaitForSignalObject or do anything will cause it in BLOCKED or WAITING mode. So thread can not be suspend(still showing Running). Until it reached breakpoint.
* If CIP already has breakpoint setted. The pause breakpoint can not be set and when thread reached breakpoint. Original breakpoint will be removed.

yuxuanchiadm:
I think pause feature will not be a simple/easy feature. Because the x86 architecture

mrexodia:
Okay, the second issue is solved when the new debug engine is in place (it allows multiple callbacks on one breakpoint). But feel free to implement it. As long as it doesn't (greatly) impact other parts of the code I will merge it.

yuxuanchiadm:
Yeah. That will helps a lot. Does it allows send some userdata to debug engine? So we can send the thread ID with breakpoint. And we can determine if the current thread is the thread we want to pause later.

yuxuanchiadm:
If it does. Close this RP. If not. Maybe add a function calls SetBPXEx?

yuxuanchiadm:
Wait. If any callback called "wait(WAITID_RUN);". The callbacks follow will blocked right?

mrexodia:
It will wait until the user has pressed the run button or stepped or
something.

yuxuanchiadm:
So this is logically no problem. But feel a little wired to user: When paused thread. Calling normal breakpoint callback. Pressed run button. Still at current CIP. Pressed run button again. Continue.

yuxuanchiadm:
So I just close this RP. When new debug engine is in place. I will check it again.

Issue #368, opened 2015-08-15
Save column widths


MoonDoggie42:
A feature that would be nice is x64 remembering the widths of my columns so I can see my labels easily in the the disasm and dump.

MoonDoggie42:
Thanks for referencing this.  How do I add a bounty?

mrexodia:
https://www.bountysource.com/issues/25785690-save-column-widths i think you can do it from there

MoonDoggie42:
Great, got a team member and I bounty-ing this bug.  Thanks mrexodia.

mrexodia:
Should be fixed now, the bounty is for @torusrxxx

PR #730, opened 2016-06-07
Fix multiple issues

torusrxxx:
I add a new tab "GUI" in the settings dialog to set the new option. By default FPU registers are shown as big-endian. Edit float registers dialog is not changed by now since the byte order is explicitly stated there. The tab "GUI" will be populated with various settings such as the option to save column width or dialog positions, as well as current language.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/730)


torusrxxx:
I am going to push commits for #368 (Save column widths).

torusrxxx:
OK, it is ready to merge now.

mrexodia:
Looks good, thanks!

PR #736, opened 2016-06-09
GUI: add shortcut for "analyze single function"

torusrxxx:
The current analyze algorithm is unsuitable for use in object-oriented application. I am debugging an executable now, and only "analyze single function" is reliable.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/736)


mrexodia:
Yes there are various analysis algorithms, but I created them before learning about graph theory, also Nukem created one. The single function one does BFS on the disassembly and gets a range from that.

I planned to add support for more functionalities but there is a lot of stuff to do 😃

If you plan to create your own analysis engine, please subclass the Analysis class and add a new command.

PR #743, opened 2016-06-13
Code folding

torusrxxx:
The code folding branch is stable enough for release now.
Known drawbacks:
1. When the user folds a large section of code, selects the instruction below the folded section, and then select upwards using keyboard, the selection may end up hiding itself inside the folded range. This is due to the "short-sightedness" of `DisassembleBack` algorithm, which only allocates a small memory range ahead of selection for instruction analysis. Therefore, it can't predict whether the whole selection have been folded far ahead. This issue is hard to fix by now. Selection with mouse and selection downwards have no issue.
2. Saving code folding information into the database is not implemented, because I think the best way to do it is by exposing `jansson` interface to all the plugins and the GUI, and expose an interface from which the GUI and the plugins can register a callback to save & load config. This should be done in a coordinated fashion.
3. The user cannot fold a section that overlaps existing code folding segment, enforcing a tree-like internal data structure.

Also, the translation is updated, and more strings are translated.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/743)


torusrxxx:
Selection backwards without "shift" key pressed works fine now. It treats the whole folded section as a single instruction.

mrexodia:
I'm still looking into this. Recently there have been some performance issues with x64dbg and while I don't think this will have much impact I still want to isolate some issues before getting into the performance of code folding.

mrexodia:
Merged, thanks a lot!

PR #749, opened 2016-06-14
Trace until condition

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/749)

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -27,6 +27,7 @@
 
 static PROCESS_INFORMATION g_pi = {0, 0, 0, 0};
 static char szBaseFileName[MAX_PATH] = "";
+ExpressionParser* RtcondCondition;
 static bool bFileIsDll = false;
 static duint pDebuggedBase = 0;
 static duint pCreateProcessBase = 0;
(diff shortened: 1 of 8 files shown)


mrexodia:
Maybe we should set a max step limit here?

src/dbg/debugger.cpp
```diff
@@ -774,6 +775,48 @@ void cbRtrStep()
         StepOver((void*)cbRtrStep);
 }
 
+void cbTOCNDStep()
+{
+    duint value = 1;
+    if (!RtcondCondition || RtcondCondition->Calculate(value, valuesignedcalc()) == false)
+    {
+        delete RtcondCondition;
```
mrexodia: This will crash if the condition is null

src/dbg/debugger_commands.cpp
```diff
@@ -1255,6 +1255,52 @@ CMDRESULT cbDebugeRtr(int argc, char* argv[])
...
+    {
+        dputs("Too many arguments.");
+        return STATUS_ERROR;
+    }
+    else if (argc < 1)
+    {
+        dputs("Too few arguments.");
+        return STATUS_ERROR;
+    }
+    if (RtcondCondition != nullptr)
```
mrexodia: This might need locking, gotta check it out.

src/dbg/debugger.cpp
justanotheranonymoususer: http://stackoverflow.com/questions/4190703

mrexodia:
Thanks a lot!

PR #753, opened 2016-06-15
Fix the issue with column resizing

torusrxxx:
I can't reproduce your issue with column reordering in breakpoints view, however I spotted another issue related to column resizing. Please add that issue to the bug tracker with steps to reproduce the problem.
And I added menu items for memory allocation.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/753)


mrexodia:
Merged, thanks!

PR #756, opened 2016-06-16
Trace into/over beyond/into trace record

torusrxxx:
With trace record on, you can let the debugger trace the program until it reaches somewhere you have not been before. And this can be conditional!
Also fixes some issues with trace record so it now works again.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/756)


src/dbg/debugger.cpp
```diff
@@ -853,6 +849,94 @@ void cbTICNDStep()
...
+    // Trace record
+    duint CIP = GetContextDataEx(hActiveThread, UE_CIP);
+    if(!traceCondition)
+    {
+        _dbg_dbgtraceexecute(CIP);
+        dprintf("Bad tracing state.\n");
+        cbRtrFinalStep();
+        return;
+    }
+    if((TraceRecord.getTraceRecordType(CIP) != TraceRecordManager::TraceRecordNone && TraceRecord.getHitCount(CIP) == 0) || !traceCondition->ContinueTrace())
```
mrexodia: Isn't checking for the condition first faster?

torusrxxx: This command is trace record oriented. Condition is often set to "0".

mrexodia: Oh I thought you also modified the current tracing, my bad.

mrexodia:
Looking great, thanks!

Issue #716, opened 2016-06-05
Add pseudo instruction DB, DD, etc


torusrxxx:
DB emits a byte into the disassembly.

wynick27:
I think analysis hint and structure view should be added to disasm engine, just like ollydbg and IDA.  And the Instruction_t and QBeaEngine for drawing GUI should be an abstract class, so it can be changed dynamically to debug vm code like IL or from a virtualizing obfuscator.

mrexodia:
I don't think an abstract class for Instruction_t will do anything useful. In my opinion it's best to customize the CapstoneTokenizer in combination with analysis hints that are pulled from the DBG. I have some ideas for this but there are a lot of things going on right now so I cannot focus on it.

torusrxxx:
I just want DB/DD. These pseudo-instructions are easy to implement in XEDParse.

mrexodia:
Should be implemented.

wqj66:
怎么添加伪指令？

Issue #632, opened 2016-04-21
Unable to detect ASCII?


DigiAngel:
Not sure this is the right question...I have a very simple Win32 GUI app that just pops up a message.  x64dbg appears to disassemble the actual message of "Hi! I'm the example program!" as code.  How can I manually set this to ASCII?  Screenshot enclosed, thank you.

![2016-04-21 15_20_23-x32dbg - file_ blehx32 exe - pid_ 2b24 - module_ blehx32 exe - thread_ ac4](https://cloud.githubusercontent.com/assets/947191/14724716/127ba1fc-07d5-11e6-950b-a5a817ee40fd.jpg)

mrexodia:
You cannot at the moment. Code and data mixed is not supported.

DigiAngel:
Ah...well there it is then...thank you much.

mrexodia:
No need to close the issue. It's simply not implemented right now

DigiAngel:
Ah...awesome thank you :)  Love this app..I use it daily.

mrexodia:
Data instructions are implemented now.

DigiAngel:
Awesome...thanks so much!

fenderluvr:
I don't see how this bug is fixed.  If I am incorrect, please explain.  I am looking at a Delphi application and can see ZERO strings, such as TForm1, when doing a string search.  Olly has not issue with this.

mrexodia:
Then use Olly. The bug was fixed because it is possible to mark data ranges as ASCII.

fenderluvr:
So you can manually mark a data range as ASCII that you are unaware of, got it. 

:)

Trying very hard to get away from Olly, but it seems will never happen.

mrexodia:
You can write an analysis plugin, that was the original request. The cunty snark certainly isn’t going to help you here so I’ll close this.

PR #758, opened 2016-06-16
Data Instruction support

wynick27:
Implemented feature mentioned in #632 #716
Currently supports disassemble and assemble integer (1-16 bytes), float (single and double), string (ascii and unicode) and mmx, xmm, ymm data instructions.
80 bit extended float is not supported.
Adds another analysis that can analyze functions, xrefs and data. (analadv command)
Allows code mixed with data.
Currently information is stored in encode map file.
image

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/758)


src/dbg/_exports.cpp
```diff
@@ -1049,6 +1056,41 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
...
+    {
+        EncodeMapDelSegment((duint)param1);
+    }
+    break;
+
+    case DBG_RELEASE_ENCODE_TYPE_BUFFER:
+    {
+        EncodeMapReleaseBuffer(param1);
+    }
+
```
mrexodia: this will fall through to `DBG_GET_STRING_AT` right?

src/dbg/advancedanalysis.cpp
```diff
@@ -0,0 +1,373 @@
...
+            FunctionDelete(end);
+            FunctionAdd(start, end, false, icount);
+        }
+    }
+    GuiUpdateAllViews();
+}
+
+void AdvancedAnalysis::analyzeFunction(duint entryPoint, bool writedata)
+{
+    //BFS through the disassembly starting at entryPoint
```
mrexodia: Notice that this code does not handle overlapping blocks correctly, not sure if this will cause problems later.

src/dbg/advancedanalysis.cpp
```diff
@@ -0,0 +1,373 @@
...
+            byte type = mEncMap[xref.from - mBase];
+            if(desttype == enc_code && type != enc_unknown && type != enc_code)
+                xref.valid = false;
+            else if(desttype == enc_middle)
+                xref.valid = false;
+        }
+    }
+}
+
+bool isFloatInstruction(x86_insn opcode)
```
mrexodia: Maybe we should add this to `capstone_wrapper`? Also this list appears to be pretty incomplete, is this just for float references?

src/dbg/assemble.cpp
```diff
@@ -181,8 +191,19 @@ bool assembleat(duint addr, const char* instruction, int* size, char* error, boo
...
     int destSize;
     unsigned char dest[16];
-    if(!assemble(addr, dest, &destSize, instruction, error))
-        return false;
+    unsigned char* newbuffer = nullptr;
+    if(!assemble(addr, dest, 16, &destSize, instruction, error))
+    {
+        if(destSize > 16)
+        {
+            newbuffer = new unsigned char[destSize];  //retry for long string sequence
```
mrexodia: We have `Memory<unsigned char*>` for this to prevent possible memory leaks.

src/dbg/breakpoint.cpp
```diff
@@ -97,7 +97,8 @@ bool BpNew(duint Address, bool Enable, bool Singleshot, short OldBytes, BP_TYPE
 
 bool BpGet(duint Address, BP_TYPE Type, const char* Name, BREAKPOINT* Bp)
 {
-    ASSERT_DEBUGGING("Export call");
+    if(!DbgIsDebugging())
```
mrexodia: Why is this here? Is BpGet called when not debugging? The assert should take care of that (although I never run in debug mode).

mrexodia:
It's looking good so far, but I have to check the code better so I will merge it later.

src/dbg/advancedanalysis.cpp
wynick27: I know, maybe I'll resolve overlapping problem later. There could be gaps between basic blocks, especially in control flow obfuscated executable.

src/dbg/advancedanalysis.cpp
wynick27: Yes, it's a work in progress, I plan to add simd instruction as well, but it takes a lot of time to look up the manuals.

src/dbg/breakpoint.cpp
wynick27: BpGet  and PatchEnum always get galled when debugging is being stopped.

wynick27:
I fixed the bug in dbg message function and uses Memory template to allocate memory.
I also added temp code list support for encode map, addresses in the list will always be decoded as code and any other address decoding that conflicts with it will be decoded as data.
Currently CIP is added to the list and if instruction in CIP is branch its target is also added.
These screenshots shows the effect of code list.
I write this test program with junk code at rva 0x1a, without analysis, it's wrongly decoded as instruction.
image
After CIP points to that address, it's correctly decoded.
image

mrexodia:
Thanks a lot, I will test the code and commit fixes if necessary...

PR #762, opened 2016-06-18
Run to user code

torusrxxx:
This works by placing temporary memory breakpoints on all user pages.
I introduced "party number" property for finer control of "run to user code" feature. The user can mark a module as user, system, or third-party in the Symbol View, and use `RunToParty` command to run to the code within a party. This is more flexible than OllyDbg's equivallent feature.
Please add an icon for "Run to user code".
In addition to "run to user code", `BPGOTO` is implemented using existing conditional breakpoint capability. It's equivallent to the following script sequence:
```
SetBreakpointCondition %1, 0
SetBreakpointCommand %1, CIP=%2
SetBreakpointCommandCondition %1, 1
SetBreakpointFastResume %1, 0
```

Translation is updated.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/762)


src/dbg/debugger_commands.cpp
```diff
@@ -699,6 +700,29 @@ CMDRESULT cbDebugGetBPXMemoryHitCount(int argc, char* argv[])
...
 }
 
+CMDRESULT cbDebugSetBPGoto(int argc, char* argv[])
+{
+    if (argc != 3)
+    {
+        dputs("argument count mismatch!\n");
+        return STATUS_ERROR;
+    }
+    char cmd[64];
```
mrexodia: this should probably be `deflen` instead of `64`

mrexodia:
Don't you think its better to use single stepping for this? Memory breakpoints are quite easily modified and/or detected by a debuggee...

torusrxxx:
We are in a system module now, so there is probably no tricky code. And we can then let the program run instead of single stepping, and rely on DEP feature on the processor serving as a hardware breakpoint.

mrexodia:
I have to check the code more, but memory breakpoints on modules sounded like you put a membp on the non-system modules to break there...

src/dbg/debugger.cpp
```diff
@@ -612,6 +613,31 @@ void cbMemoryBreakpoint(void* ExceptionAddress)
     cbGenericBreakpoint(BPMEMORY, ExceptionAddress);
 }
 
+void cbRunToUserCodeBreakpoint(void* ExceptionAddress)
+{
+    EXCLUSIVE_ACQUIRE(LockRunToUserCode);
```
mrexodia: Won't this deadlock if you try to go to user code when waiting inside this callback?

src/dbg/module.cpp
```diff
@@ -105,6 +105,20 @@ bool ModLoad(duint Base, duint Size, const char* FullPath)
     info.fileMap = nullptr;
     info.fileMapVA = 0;
 
+    // Determine whether the module is located in system
+    wchar_t sysdir[MAX_PATH];
+    GetEnvironmentVariableW(L"windir", sysdir, sizeof(sysdir));
+    String Utf8Sysdir = StringUtils::Utf16ToUtf8(sysdir);
```
mrexodia: Maybe we should add to lowercase here, just to be sure? Maybe there are some rare cases where windows reports a different case?

src/dbg/module.cpp
```diff
@@ -400,4 +414,30 @@ bool ModAddImportToModule(duint Base, const MODIMPORTINFO & importInfo)
...
+    // If the module is not found, it is an user module
+    if (!module)
+        return 0;
+
+    return module->party;
+}
+
+void ModSetParty(duint Address, int Party)
+{
+    SHARED_ACQUIRE(LockModules);
```
mrexodia: This should be EXCLUSIVE_ACQUIRE because it modifies the module.

src/dbg/module.cpp
torusrxxx: I'm using memicmp. It is case-insensitive.

src/dbg/module.cpp
torusrxxx: May be it should be, but I think it is useless. Modern processor cache coherence protocol already guarantees that an aligned word read/write is atomic.

torusrxxx:
I'm going to sleep now.

src/dbg/module.cpp
Nukem9: Why not use InterlockedExchange then? Or at least write a comment explaining it, since it might raise issues later.

torusrxxx:
These issues have been fixed.

mrexodia:
Okay cool! I will try to look at it as soon as possible....

torusrxxx:
The current icon (recent-files) for "run until user code" is just a place
holder waiting to be replaced by a dedicated icon...

mrexodia:
Okay, sure.

Issue #763, opened 2016-06-19
Add an interface to allow GUI and plugins to save data in the database


torusrxxx:
`jansson` is already included in the plugin sdk.

mrfearless:
Might be handy to also expose certain default paths for plugins (and/or scripts) to use:

Could be default variables obtained via a function call or a structure passed back or something like that?

x64dbg_ini - stores a path to x32dbg.ini/x64dbg.ini for example m:\x64dbg\x32\x32dbg.ini
x64dbg_root_folder - stores path to root folder containing exe - m:\x64dbg\x64\
x64dbg_plugins_folder - stores path to plugins folder: m:\x64dbg\x32\plugins\
x64dbg_db_folder - stores path to db folder: m:\x64dbg\x64\db\

Maybe a _plugin_getpaths function that returns a structure with the path information populated for some locations that might be required.

mrexodia:
Yeah sounds like a good idea!

wynick27:
I think we need apis for both local setting storage and global setting storage.

mrexodia:
There is already an api for global settings (the ini). For local I assume
you mean per debuggee settings? That should be implemented once we export
the database api.

PR #766, opened 2016-06-21
Plugin database storage

torusrxxx:
This is the code for #763 
All plugin data is stored inside "plugins" object.
This branch has not been tested. It needs to be tested with a working plugin.

CB_LOADDB/CB_SAVEDB callback types are introduced with argument type PLUG_CB_LOADSAVEDB.

_Since all the plugin callbacks are supplied with a JSON object, should we increment its reference count before every plugin callback, or left that as a responsibility of the plugin?_

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/766)

### DIFF
--- a/src/dbg/_plugin_types.h
+++ b/src/dbg/_plugin_types.h
@@ -19,6 +19,7 @@
 
 #include "bridgemain.h"
 #include "_dbgfunctions.h"
+#include "jansson/jansson.h"
 
 #endif // BUILD_DBG
 
--- a/src/dbg/_plugins.h
+++ b/src/dbg/_plugins.h
@@ -25,6 +25,9 @@
 //defines
 #define PLUG_SDKVERSION 1
 
+#define PLUG_DB_LOADSAVE_DATA 1
+#define PLUG_DB_LOADSAVE_ALL 2
+
 //structures
 typedef struct
 {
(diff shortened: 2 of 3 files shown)


mrexodia:
Great! I think we should make sure the plugin cannot screw up references. This could be done by setting a custom allocator before the plugin handling. The reference count increase should go automatically if you allocate a new json object you don't have to do it manually as far as I know...

torusrxxx:
We're giving a single root object to every plugins. As far as I know the
plugin callback dispatcher just  call every callback with the same
parameter.

mrexodia:
Yes but yoy don't need to increase the reference count, it should be handled automatically if you set something in the root object.

torusrxxx:
So if the plugin just creates its own object and append that object to the
root without json_decref(root), everything is fine. But if it invokes
json_decref(root), then everything is wrong.

torusrxxx:
And some plugins just change the root parameter, so that all subsequent
plugins will load/save data from a different storage. These issues are
really plugin bugs. We can just document the issue in the documentation and
leave the debugging work to the plugin developers.

mrexodia:
Right... We can just backup the reference count maybe and hard crash if it gets changed?

torusrxxx:
We needn't do anything. We just need to test if the pluginsdk can be built
and used. The debugger can be built.

PR #768, opened 2016-06-22
watchdog and script controlled conditional breakpoint

torusrxxx:
Watchdog is a feature related to the unfinished watch view. It notifies the user if a watch item has changed or become true since last checkpoint. In principle it can be checked in a worker thread, but currently a command is provided to check the watchdog conditions.

Four new commands are added. "addwatch" adds a expression to the watch view, "delwatch" deletes a watch. "setwatchdog" sets the watchdog mode. "watchdog" command checks all the watch items and notify the user if any watchdog is triggered.
"watchdog" command can be placed in the breakpoint command, so that the breakpoint becomes a checkpoint. It will set "breakcondition" variable and "result" variable to true if any watchdog is triggered.

In order to pause the debugger when a watchdog is triggered, I introduced script controllable conditional breakpoint. Break condition is passed in and out as "breakcondition" variable, which can be accessed from script. The user can check something in the script, and perhaps modify the program context, then generate the break condition in script. However this feature is only useful if the user can execute a script instead of a command. We need to allow "jump" and "call" commands outside the script environment so that the user can write "jump myconditionalbreakpoint" in the edit breakpoint dialog.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/768)

### DIFF
--- a/src/dbg/database.cpp
+++ b/src/dbg/database.cpp
@@ -15,6 +15,7 @@
 #include "loop.h"
 #include "commandline.h"
 #include "database.h"
+#include "watch.h"
 #include "threading.h"
 #include "filehelper.h"
 #include "xrefs.h"
@@ -54,6 +55,7 @@ void DbSave(DbLoadSaveType saveType)
         XrefCacheSave(root);
         TraceRecord.saveToDb(root);
         BpCacheSave(root);
+        WatchCacheSave(root);
 
         //save notes
         char* text = nullptr;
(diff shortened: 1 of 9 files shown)


torusrxxx:
I have changed my mind. I just want to check the watchdog all the times. Maybe we can add it an option instead of a command. So although this patch works, it's better deferred until the watch view is implemented, and finalize the watch view design then.

mrexodia:
So should I close this pull request? As for executing a script we can make a command for that.

torusrxxx:
It can be closed. I'll push another (probably large) pull request after
instruction undo and watch view finalize.

mrexodia:
Okay, I'm currently kind of in merging hell so it might be painful to create a pull request that has no conflicts, sorry about that. By the way, you can get 10$ for a bounty on bountysource.

PR #793, opened 2016-06-23
Translate the launcher and add more shortcuts

torusrxxx:
The translated strings are stored in "Strings.rc" and loaded via `LoadString`. `LoadString` is the only viable way to get translated strings before the GUI is loaded.
The shortcuts added are :

- ActionEditBreakpoint
- ActionToggleLogging
- ActionAllocateMemory
- ActionFreeMemory
- ActionSyncWithExpression
- ActionEntropy
- ActionCopyAllRegisters
- ActionMarkAsUser
- ActionMarkAsSystem
- ActionMarkAsParty
- AnalysisRemoveAnalysisFromModule
- AnalysisRemoveAnalysisFromSelection
- AnalysisTreatSelectionAsByte
- AnalysisTreatSelectionAsASCII
- AnalysisTreatSelectionAsUNICODE
- ActionSetHwBpE
- ActionRemoveHwBp
- DebugTraceIntoConditional
- DebugTraceOverConditional
- DebugEnableTraceRecordBit
- DebugTraceRecordNone

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/793)

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -20,35 +20,35 @@ static bool bDisableGUIUpdate;
         return;
 
 #ifdef _WIN64
-#define dbg_lib "x64dbg.dll"
-#define gui_lib "x64gui.dll"
+#define dbg_lib L"x64dbg.dll"
+#define gui_lib L"x64gui.dll"
 #else
-#define dbg_lib "x32dbg.dll"
-#define gui_lib "x32gui.dll"
+#define dbg_lib L"x32dbg.dll"
+#define gui_lib L"x32gui.dll"
 #endif // _WIN64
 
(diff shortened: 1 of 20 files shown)


mrexodia:
I don't want to merge the translation because it's ugly and slows down development...

torusrxxx:
Launcher doesn't need to be updated very frequently. Last update was months
ago...

mrexodia:
Yes that's true. I still think we should find a reliable framework to translate it to other languages without having to update resource files in the repository. Also there were changes to the bridge which are not necessary as far as I'm aware.

torusrxxx:
As far as I know there is no such framework except LoadString

mrexodia:
Okay so I will merge this :+1:

PR #801, opened 2016-06-24
unicode character display

torusrxxx:
x64dbg-ucs

We should add "wide character support" in RichTextPainter. In such mode the rich text painter paints each character with a fixed but wide enough width, so that almost every unicode characters can fit within.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/801)

### DIFF
--- a/src/gui/Src/BasicView/HexDump.cpp
+++ b/src/gui/Src/BasicView/HexDump.cpp
@@ -729,8 +729,8 @@ QString HexDump::wordToString(uint16 word, WordViewMode_e mode)
 
     case UnicodeWord:
     {
-        QChar wChar = QChar::fromLatin1((char)word & 0xFF);
-        if(wChar.isPrint() == true && (word >> 8) == 0)
+        QChar wChar(word);
+        if(wChar.isPrint() == true)
             wStr = QString(wChar);
         else
             wStr = ".";


mrexodia:
As said, I don't want to merge this because of the performance degrade.

torusrxxx:
I'll push the implementation of "wide mode" here if I have enough time. Display unicode string is not performance impacting. Ollydbg does it quite well, and every other hex editor, and every software can display them well. It will impact performance only if you want to get the width of every character, but we will not do it here.

mrexodia:
Qt calculates the width of every character for drawing the character. I checked with a performance profiler and the performance is really bad. Obviously checking multiple times (which RichTextPainter does) is bad, but that was 'fixed' by the font width cache.

torusrxxx:
It seems modern browsers, word processors and even multi-line text boxes
cache the painted bitmap, and when it needs redraw, only the dirty area is
rendered(other area is fetched from the cache). Qt supports this technology
in "QStaticText". If used properly it may resolve the performance issue.

mrexodia:
Yes, I looked into it, but it has certain issues (for example in the disasm its completely useless and only drains performance). In the dump it might be viable because the dump doesn't update so often. There is also a comment in `RichTextPainter` about using `QTextLayout` instead of the current rich text implementation, however this will break token selection amongst other things.

torusrxxx:
If you view it in another aspect, things aren't so difficult. Most tokens
are just "rax", "mov" and so. If we cache them, then lots of slower
"drawText" will be replaced with faster "drawStaticText".

mrexodia:
Right that's a good point!

torusrxxx:
further work abandoned.

PR #803, opened 2016-06-25
Fix some tiny issues

torusrxxx:
and add script controllable conditional breakpoint support in advance

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/803)

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -515,6 +515,8 @@ static void cbGenericBreakpoint(BP_TYPE bptype, void* ExceptionAddress = nullptr
     bp.addr += ModBaseFromAddr(CIP);
     bp.active = true; //a breakpoint that has been hit is active
 
+    varset("$breakpointcounter", bp.hitcount, false); //save the breakpoint counter as a variable
+
     //get condition values
     bool breakCondition;
     bool logCondition;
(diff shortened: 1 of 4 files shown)


src/dbg/debugger.cpp
```diff
@@ -581,7 +583,42 @@ static void cbGenericBreakpoint(BP_TYPE bptype, void* ExceptionAddress = nullptr
     if(*bp.commandText && commandCondition)  //command
     {
         //TODO: commands like run/step etc will fuck up your shit
-        DbgCmdExec(bp.commandText);
+        varset("$breakpointcondition", breakCondition ? 1 : 0, false);
+        varset("$breakpointlogcondition", logCondition, false);
+        _dbg_dbgcmddirectexec(bp.commandText);
```
mrexodia: Won't this deadlock if your command is waiting for the user to resume debugging?

torusrxxx: But it is not possible to get the script-controlled condition if async execution mode is used.

mrexodia: Yes that's a good point. Well I can just merge it and wait for a report.
Probably most users won't be using these advanced features anyway.

torusrxxx: In fact we shouldn't allow these commands to execute here. We have introduced script-controlled condition, so there is no need for such commands.

mrexodia: But if you allow execution of an arbitrary script file anything can happen,
scripts can call commands registered by plugins that change run locks and
so on...

torusrxxx: There isn't any ready-made script file that make use of script command to pause now. If we introduce script-controlled break condition earlier, then less people will make these script files. We should tell them to use "breakpointcondition" instead of such commands.

PR #807, opened 2016-06-28
trace record during trace

torusrxxx:
commit selected feature on my local branch once it is stable

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/807)

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -77,6 +77,7 @@ HANDLE hActiveThread;
 HANDLE hProcessToken;
 bool bUndecorateSymbolNames = true;
 bool bEnableSourceDebugging = true;
+bool bTraceRecordEnabledDuringTrace = false;
 duint DbgEvents = 0;
 
 static duint dbgcleartracecondition()
@@ -882,6 +883,8 @@ void cbRtrStep()
     unsigned char ch = 0x90;
     duint cip = GetContextDataEx(hActiveThread, UE_CIP);
     MemRead(cip, &ch, 1);
+    if(bTraceRecordEnabledDuringTrace)
+        _dbg_dbgtraceexecute(cip);
     if(ch == 0xC3 || ch == 0xC2)
         cbRtrFinalStep();
     else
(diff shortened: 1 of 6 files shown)


mrexodia:
Just a quick train of thought, we can enable only saving branches and inferring the execution path from that. Obviously this will not work in non-modules or self-modifying code but it will greatly decrease the size of the data produced.

torusrxxx:
No. It will get lost quickly if you step over several obfuscated loops with
indirect calls or jumps. We must record it per byte.

mrexodia:
I don't agree with 'must'. Branch tracing does also only record branches.
Obviously obfuscation and such won't work but for normal code it will work
fine in all cases just think.

torusrxxx:
I will implement software-based branch record with trace record framework
after I finish debugging my executables. This is a separate trace mode and
the first variable-length trace mode. It will consider all branch
instructions, calls/rets and exceptions as branch. It will record branch
destinations and percentage.

torusrxxx:
Could you please take a look at this pull request and tell me if it is okay to merge? I think this pull request is more complete than 2c3eb1ec47ab90d6de302e4. It introduces an option controlling whether to record data during a trace so that tracing can speed up if trace record is not needed. I like to make separate pull requests for each feature I want to introduce so probably no more commits will come here.

mrexodia:
Yeah, it's looking better indeed. 2c3eb1e was just a quick fix I did for a potential demo during the conference. So can I merge this code? I was under the impression that it was not yet finished.

src/dbg/debugger.cpp
```diff
@@ -1079,6 +1098,8 @@ static void cbCreateProcess(CREATE_PROCESS_DEBUG_INFO* CreateProcessInfo)
             sprintf_s(command, "bp " fhex ",\"entry breakpoint\",ss", (duint)CreateProcessInfo->lpStartAddress);
             cmddirectexec(command);
         }
+
+        bTraceRecordEnabledDuringTrace = settingboolget("Engine", "TraceRecordEnabledDuringTrace");
```
mrexodia: Shouldn't this be relocated at the appropriate location in `_exports.cpp:825` (`DBG_SETTINGS_UPDATED`)?

mrexodia: Hm it already appears to be there, so why is it here?

torusrxxx: coz DBG_SETTINGS_UPDATED is not triggered when the application initializes so the `bTraceRecordEnabledDuringTrace` is not initialized. This line of code initializes it.

mrexodia: Oh that's actually quite a bug in that case...

PR #821, opened 2016-07-05
Fixed #817; added async rate limiting calls

### LINKED ISSUE #817: Rapidly executing breakpoints causes access violation
To reproduce, set a breakpoint somewhere in a loop and add 'run' as its command. 

Namely coming from [here](https://github.com/x64dbg/x64dbg/blob/development/src/dbg/dbghelp_safe.cpp#L13). 

The comment there is correct; with the lock it deadlocks almost immediately. However without it, the access violation is possible, and I expect that the reproduction instructions just stress test it, and that it can happen without a somewhat useless BP command set. 

I'm not sure what would be the preference for a fix, or I'd submit a PR. In my development environment I fixed it by adding recursive lock logic in threading.h. It is decently lightweight, but I'd understand if there was an end goal of not needing any recursive locks.

jdavidberger:
Major changes:

- SectionLockerGlobal now supports recursive SRW exclusive locks. Semantics should now match what happens with CriticalSection support. 
- With that change, the lock referenced in #817 is added back in with no risk of deadlock from recursive locking. 

- From profiling, rapid running and breaking would also cause the QT message pump to be flooded, so UI updates don't happen, among other issues with performance. Many of these function calls can be safely rate limited since they don't have an accumulative effect. 

    For this, a utility class TaskThread is added. This class takes a function and spawns a dedicated thread for that function. WakeUp kicks the thread to execute the given function. The intended semantics for this class is that it will drop / delay consecutive function calls so they are invoked at an interval which can be specified in milliseconds. Also, the last argument passed into WakeUp will always be the last one executed upon. 

    This change made breakpoint setups which would effectively freeze the application churn through it at an acceptable pace while the UI remains responsive. I think, in general, a good goal would be to keep the number of messages bridge puts into the events queue less than 100 per second.

    Besides speed and correctness, the implementation hides out the details in an opaque pointer. Normally I dislike this approach but since most TaskThreads would be created at init and not dynamically allocated after that, it seemed worth it to make the inclusion costs for the header marginally lower. 

Minor issues fixed:
- Fixed tab order on the edit breakpoint dialog. 
- x86 had a build error for me; unclear if this was just me or not. `_EXCEPTION_REGISTRATION_RECORD` was coming up as also defined in winnt.h. For x64, no such issue. If this isn't happening anywhere else I'll relook at my dev setup; which otherwise works.

Apologies for the wall of text / long PR. Let me know which parts to keep and I'll redo it. Hopefully the tab order fix makes it in though; it was the original reason I built from source in the first place!

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/821)


src/dbg/threading.h
```diff
@@ -96,9 +110,18 @@ class SectionLockerGlobal
...
             if(Shared)
+            {
                 m_ReleaseSRWLockShared(&m_srwLocks[LockIndex]);
-            else
+                return;
+            }
+
+            assert(m_owner[LockIndex].count && m_owner[LockIndex].thread);
+            m_owner[LockIndex].count--;
+            if(m_owner[LockIndex].count == 0)
```
mrexodia: Doesn't this require locking because no atomic operations are used to manipulate or compare these count values? This means you have a race condition on this count variable.

src/dbg/_global.h
```diff
@@ -2,7 +2,9 @@
 #define _GLOBAL_H
 
 #define _WIN32_WINNT 0x0501
+#ifndef WINVER
```
mrexodia: Why was this addded? The WINVER is defined as `0x0501` because x64dbg should be compiled with Windows XP headers.

src/dbg/debugger.cpp
```diff
@@ -127,7 +128,7 @@ static DWORD WINAPI memMapThread(void* ptr)
             memMapThreadCounter = GetTickCount();
         }
 
-        Sleep(50);
+        Sleep(1000);
```
mrexodia: This timer might be too long. It might be worth to change this to something similar as the `TaskThread` you implemented. The semantics I would prefer are like `2000ms` when idle and an immediate wake up as soon as an update is required.

src/dbg/_global.h
jdavidberger: Seems like a toolchain difference. Mine was setting WINVER to 0x601. Since we want to target 501 explicitly, this code is likely the wrong approach. I verified this by turning on 'Show includes' and there were no headers before _global.h in compilation units which displayed the warning.

Not sure why you aren't seeing the redef warning; probably it is just a minor difference in toolchain versions or something. How do you feel about something like this:

```
#ifdef WINVER // Overwrite WINVER if toolchain gives us a default
#undef WINVER
#endif
#define WINVER 0x0501
```

which squashes the warning and applies the right WINVER.

src/dbg/debugger.cpp
jdavidberger: I agree something like TaskThread might be nice here. I didn't realize memMapThreadCounter was externally visible and reset, so I mistakenly thought that sleep(1000) wouldn't change any behavior. 

Now that I have a better understanding of the code, I'll replace it with a TaskThread which has the desired semantics.

src/dbg/taskthread.cpp
```diff
@@ -0,0 +1,95 @@
+#include <Windows.h>
```
mrexodia: This is the include that gives the warnings. If you remove it everything works as desired (without modifying `_global.h`).

src/dbg/debugger.cpp
mrexodia: Just a quick thing I wanted to change. We should give the `TaskThread` an identifier and then store/load their timeout in the `settings.ini` this way you can easily customize how heavily you care about GUI updates. Setting the default to `100` milliseconds gave me instant GUI updates when spamming buttons, but nice skips (as desired) while holding the button.

There is also a bug where the first GUI update would not be registered correctly (it didn't change the CIP). I did not figure out why this happened yet.

src/dbg/taskthread.cpp
jdavidberger: Even moving that around I get this kind of thing: https://gist.github.com/jdavidberger/d6b93ffed5195a7ad6139f5d907fc278 in other compilation units. cl is giving a version as Version 18.00.40629 for x86 and VS2013 version 12.0.40629.00 update 5 for what its worth. 

This does bring up though that apparently InitializeConditionVariable isn't available on XP. I'll reimplement TaskThread with a semaphore and fix the header order.

mrexodia: There is no need to move it, just delete it. Global includes all windows
headers. I will update visual studio 2013 to the latest version.

mrexodia: Okay so it seems like you define WINVER somewhere in the command line or something. Possibly a global setting in Visual Studio, maybe an environment variable? But yeah in that case the WINVER define should probably be added in the project files.

Issue #820, opened 2016-07-05
Watch view


torusrxxx:
Implement a watch view similar to the ones in Visual Studio.
It's almost finished, however there is no issue about it yet.

Issue #819, opened 2016-07-05
Undo last instruction


torusrxxx:
Allow the user to restore the context to a previous state to undo the instruction last stepped.
It's almost finished, however there is no issue about it yet.

Issue #500, opened 2016-01-11
Add more menus with favourite tools


aimtsou:
It would be nice if somewhere in the options we could specify a path or many paths with our selected tools and then to get a new entry into the menu or in the toolbar as a button.

Nukem9:
The idea is alright, but what is the purpose? Why not just use a launcher program or the Windows taskbar shortcuts?

mrexodia:
Nukem too easy thinking :D

mrexodia:
might be a good idea to start a tool with the current debuggee as argument.

torusrxxx:
Progress have been made on this.

Issue #499, opened 2016-01-11
Ability to force-terminate the debugger and generate dump file


wk-952:
In some situations the debugger hangs, and when you press the close button
it won't terminate, eventually you'll be looking at the message `"Closing Debugger..."`
and the log output area at the very bottom will have the message `"Cleaning wait objects..."`,
and of course it can only be closed from the task manager.
Add to that in most of these situations the database file is completely deleted.

Could you please implement an ability to force-terminate the debugger, but that option would
generate the minidump files as it would do it the normal crash scenarios ?

The problem is, in that specific scenario no dump files are generated and i don't get the option
to debug x64dbg itself to copy the callstack table or to provide you with any useful information.

*(off-topic): could you also please make the debugger generate a backup of the db file
that would get written on a successful termination ?

Nukem9:
mrexodia:
Should x64dbg create a `dbname.dd32.bak` as a copy of the database before loading it in `DbLoad()`?
As far as the hanging part.......some kind of timeout could throw an exception after X seconds. Won't be necessary if the root cause is found.

wk-952:
Task manager should allow you to create a dump file, though I don't know if it's present in all windows versions. (The DMP from here should be kept/uploaded privately).
![image](https://cloud.githubusercontent.com/assets/4228359/12227506/a34c721c-b7fb-11e5-8a44-92f86d231cea.png)

mrexodia:
yea, it should make a bak* thing with `DbLoad`

wk-952:
So when the debugger hangs and can't be closed normally, all i have to do is generate a dump file from the taskmanager right ?

@mrexodia as Nukem said, they should be uploaded privately, so could i just send you those dumps directly to your e-mail, or is there anywhere else to upload them privately ?

wk-952:
@mrexodia i would appreciate any response from you to the above 2 questions please.

mrexodia:
Eh yeah sure I wasn't paying attention when this originally happened. Will
check it out.

mrexodia:
Just a quick note, you can save the database if the command thread is not hanging with the `dbsave` command.

wk-952:
Yeah sure of course, from time to time while debugging i invoke the `dbsave` command.
What i meant in the very first post is that the debugger will sometimes becomes completely unresponsive (you can't do anything except closing it from TaskManager).

[...]

PR #825, opened 2016-07-07
Multiple features

torusrxxx:
- Watch (#820)
- Instruction undo (#819)
- Favourite tools (#500)
- minidump (#499)

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/825)


src/dbg/debugger_commands.cpp
```diff
@@ -715,20 +728,76 @@ CMDRESULT cbDebugSetBPGoto(int argc, char* argv[])
...
     _snprintf(cmd, sizeof(cmd), "SetBreakpointFastResume %s, 0", argv[1]);
-    if(!DbgCmdExecDirect(cmd))
+    if(!_dbg_dbgcmddirectexec(cmd))
         return STATUS_ERROR;
     return STATUS_CONTINUE;
 }
 
+CMDRESULT cbDebugSetBPHwKeep(int argc, char* argv[])
+{
+    // will this have an effect?
```
mrexodia: Yep, you will get an access violation if the user called with no arguments and you try to access `argv[1]`.

src/dbg/debugger.cpp
```diff
@@ -1254,7 +1297,7 @@ static void cbExitThread(EXIT_THREAD_DEBUG_INFO* ExitThread)
     }
 }
 
-static void cbSystemBreakpoint(void* ExceptionData)
+static void cbSystemBreakpoint(void* ExceptionData) // TODO: System breakpoint event shouldn't be dropped
```
mrexodia: What exactly do you mean with 'dropped'?

src/dbg/debugger.cpp
```diff
@@ -35,13 +37,21 @@ struct TraceCondition
...
+    inline bool ContinueTrace()
     {
         steps++;
         if(steps >= maxSteps)
             return false;
         duint value = 1;
-        return condition.Calculate(value, valuesignedcalc()) && !value;
+        if(!(condition.Calculate(value, valuesignedcalc()) && !value))
+            return false;
+        cbWatchdog(0, nullptr);
```
mrexodia: Won't this horribly affect tracing performance when the watchdog is not of interest?

src/dbg/watch.cpp
```diff
@@ -0,0 +1,551 @@
...
+        WatchSetWatchdogModeUnlocked(id, watchdog_mode);
+    }
+    WatchInitIdCounter(); // Initialize id counter so that it begin with a unused value
+}
+
+CMDRESULT cbWatchdog(int argc, char* argv[])
+{
+    EXCLUSIVE_ACQUIRE(LockWatch);
+    bool watchdogTriggered = false;
+    auto lbl = SymGetSymbolicName(GetContextDataEx(hActiveThread, UE_CIP));
```
mrexodia: This function is very bad performance-wise.

src/dbg/historycontext.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include "_exports.h"
+
+duint HistoryMaxCount = 4096;
+std::deque<HistoryContext> history;
+
+//This will capture the current instruction
+HistoryContext::HistoryContext()
+{
+    DISASM_INSTR instruction;
+    if(GetFullContextDataEx(hActiveThread, &registers) && MemIsValidReadPtr(registers.cip))
```
mrexodia: This will kill performance (both `GetFullContextDataEx` and `MemIsValidReadPtr` are hotspots in performance analysis). I think it should be disabled per default until a new debug engine is in place that will cache this data correctly.

src/dbg/watch.cpp
torusrxxx: then just print the CIP is enough

src/dbg/debugger.cpp
torusrxxx: Well I can introduce an option

src/dbg/historycontext.cpp
torusrxxx: I think we can disable history recording during an automated trace.

torusrxxx:
It compiles now.

src/dbg/debugger.cpp
torusrxxx: if there are too many debug events then it is dropped

mrexodia:
Merged, thanks a lot!

PR #833, opened 2016-07-09
Two feature additions, two minor bug fixs, one build setting fix

jdavidberger:
Bug fixs:
- 4e43788 - When selecting an item in autocomplete, and hitting space, QT always would throw an assert in debug mode. Occasionally it would cause an access violation (seemingly random, likely timing based). The guard just doesn't reset the index if it'd be setting the index to be greater than the number of available items
- b1edb1a - The command to scroll to the end was given before inserting text into log view. This is mostly unnoticeable normally, but if there are multiple lines in a single incoming log message it scrolls offscreen

Features:
- The edit breakpoint dialog now has a 'silent' option, which squelches the default log message when a breakpoint is hit
-  _plugin_registerexprfunctionuserdata added, which is like _plugin_registerexprfunction except the callback passes back a given pointer

Build fix:
- Switched pointed at libs to be the correct one in debug mode. 
- Possibly not the best solution; but made a change so that in debug mode it doesn't try to invoke snowman at all. This is a practical fix since the precompiled version of snowman is release mode and incompatible with debug mode.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/833)


src/dbg/expressionfunctions.h
```diff
@@ -5,20 +5,26 @@
...
     static bool Register(const String & name, int argc, CBEXPRESSIONFUNCTION cbFunction);
+    static bool Register(const String & name, int argc, CBEXPRESSIONFUNCTIONWITHUSERDATA cbFunction, void* user);
     static bool Unregister(const String & name);
     static bool Call(const String & name, const std::vector<duint> & argv, duint & result);
     static bool GetArgc(const String & name, int & argc);
 
 private:
     struct Function
     {
+        CBEXPRESSIONFUNCTION cbFunction;
```
jdavidberger: I'm not terribly happy with this implementation, but I figured the options were: 

- Make everything require the userdata invocation. This would break existing uses of Register, and in doing so making call sites that didn't need a user data pointer to be uglier
- Make Function an interface and have two subclasses. Downside here is we'd have to change the map to point to a pointer of that interface; and manage some dynamic allocations. 
- Tagged unions for the CB function; however VS2013 doesn't let you do C++11 union stuff and C++03 doesn't let you do unions on complex types like std::function. 
- Wrap functions without a user data arg in a function that takes in the CBEXPRESSIONFUNCTION as it's user data function. Seemed unnecessarily maze like though; and I'm not actually sure if you could find a good way to cast CBEXPRESSIONFUNCTION to void*.  

In the end I just tried to keep it simple and easy to follow through the code.

mrexodia: `std::function` also has captures and since the userdata has to be provided during registration there don't have to be changes to this particular definition.

mrexodia: nvm that doesn't make sense.

PR #834, opened 2016-07-10
Added ability to roughly measure UI latency; code to try and maintain that latency at <50ms

jdavidberger:
The added logic is in deferable taskthreads; they now check the current latency and defer if that latency is too high. This should prevent any additional events from being sent to QT; which should put a firm limit on how backed up the queue can get during heavy load. 

Also modified the event count logic so it took elapsed time into account. Without this logic, the number reported was based on message processing timing, which fluctuates during load; even when the messages are added at one second intervals.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/834)

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -84,7 +84,7 @@ HANDLE hProcessToken;
 bool bUndecorateSymbolNames = true;
 bool bEnableSourceDebugging = true;
 bool bTraceRecordEnabledDuringTrace = true;
-duint DbgEvents = 0;
+uint64 DbgEvents = 0;
 
 static duint dbgcleartracecondition()
 {
@@ -280,26 +280,9 @@ bool dbgcmddel(const char* name)
     return true;
 }
 
-duint dbggetdbgevents()
+uint64 dbggetdbgevents()
 {
-    static duint DbgEventsLastSample = 0;
-    static uint32 DbgEventsLastReset = 0;
-
-    duint eventsSinceLastQuery = InterlockedExchange(&DbgEvents, 0);
-    uint32 now = GetTickCount();
-    uint32 timeDiffMs = now - DbgEventsLastReset;
-
-    // With too small of a time difference, it's hard to get a consistent sample.
-    if(timeDiffMs < 50)
-        return DbgEventsLastSample;
-
-    if(DbgEventsLastReset == 0 || (now < DbgEventsLastReset))
-    {
-        // Either firt time this was called, or rollover. Either way, just call it a second
-        timeDiffMs = 1000;
-    }
-    DbgEventsLastReset = now;
-    return DbgEventsLastSample = (eventsSinceLastQuery * 1000 / timeDiffMs);
+    return InterlockedCompareExchange64((int64*)&DbgEvents, -1, -1);
 }
 
 static DWORD WINAPI updateCallStackThread(duint ptr)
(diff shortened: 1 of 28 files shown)


torusrxxx:
I dislike some modifications of it. I want to know how many events are passed precisely. So I cannot accept any modifications of the result of dbggetdbgevents. I would rather like "%1 events in %2 ms".

jdavidberger:
If that is what is wanted UI-wise, I'd consider removing the clear behavior from dbggetdbgevents altogether, and let the widget itself handle the logic by tracking time between updates and subtracting off the last call from dbggetdbgevents. This would also open the door to maybe being able to configure that display to be in hertz, or the %1 events in %2 ms status.

torusrxxx:
# Usages of event counters:
1. As a general purpose counter: When there are many events bursting out in a second, you can count it with events counter.
When the program starts you may notice there are many debug events. Most of these events are DLL load events. If the program loads within one or two seconds, you can directly watch the events counter and roughly know how many DLLs the program uses. Similarly, if the program outputs some debug strings, you can count the number of strings from the event counter. __This usage requires an exact counter.__
2. As a tool to watch the state of the program: Well-designed "animate into" operation should maintain a very steady stepping rate. If there are any extra debug events, you know something unexpected is happening, and probably stop stepping immediately. __This usage does not require an exact counter.__

jdavidberger:
Sure, but at no point do you have an exact counter right now -- it is clearing itself, and at some points is doing so within 50ms of the last time it cleared based on what the internal timings of the message queue was. 

Here is what I see as the possible ways to go with this, let me know which you want and I'll change it to match:

1. Modify getdbgevents to remove the clearing behavior, and let the UI display both 'total events' which is clearable by the user, and possibly events in hertz. This would have benefit of being able to remove the interlock on that variable; and would allow other UI elements to use dbggetdbgevents -- currently they can't because two clients calling dbggetdbgevents would clear the event counter and each would get roughly half the information. 
2. Revert out all the dbgevent changes. I don't think this is the best approach long term, but I think there is other useful stuff in this PR and I don't want to hold it up for what boils down to a minor UI change.

mrexodia:
I think another alternative is to simply keep two counters. One with the
original behavior and one for these changes.

torusrxxx:
Don't have to output only one variable in dbggetdbgevents! You can change the signature of dbggetdbgevents to `void dbggetdbgevents(uint* events, uint* milliseconds)` and output both the raw eventsSinceLastQuery and the timeDiffMs. Then you can write any logic in the GUI. If you feel you don't have time to add an option for switching between different counter modes it is still acceptable, because it's very easy to add one.

jdavidberger:
Fixed the dbggetdbgevents functionality. You can now click the time wasted widget and it will swap between view modes; although these still only work when events are coming in. Default is to show delta from last time displayed, plus the # of ms since then.

src/dbg/debugger.cpp
```diff
@@ -280,9 +280,9 @@ bool dbgcmddel(const char* name)
     return true;
 }
 
-duint dbggetdbgevents()
+uint64 dbggetdbgevents()
 {
-    return InterlockedExchange(&DbgEvents, 0);
+    return DbgEvents;
```
torusrxxx: race condition when reading out the value. Use an interlocked function to read instead.

mrexodia:
So did you guys sync up @jdavidberger and @torusrxxx can this be merged or is there still something blocking it?

jdavidberger:
I believe everything is resolved now; and it should be ready to merge in.

mrexodia:
Alright I will check it out in detail and merge.

mrexodia:
Great, thanks a lot!

torusrxxx:
Build error : undefined symbol `InterlockedIncrement64` and `InterlockedCompareExchange64` in release package and official VS13 development environment. x32dbg.dll cannot be built.

Issue #852, opened 2016-07-16
Bswap


mrexodia:
Add this as a command

bloodwrath:
added in c32eefa

PR #856, opened 2016-07-16
Issue #852 add bswap command

bloodwrath:
I decided to use the msdn byteswap functions because it looks cleaner imo.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/856)


mrexodia:
Thanks a lot!

PR #865, opened 2016-07-18
use MenuBuilder in CPUDump

torusrxxx:
and clear trace condition whenever the debuggee pauses.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/865)


torusrxxx:
and fix error #866

PR #869, opened 2016-07-18
fix wrong behaviour with prefixed retn

torusrxxx:
In "rtr", the debugger cannot stop tracing when the executable uses some prefixed retns such as "40 C3"

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/869)


mrexodia:
Thanks!

Issue #842, opened 2016-07-13
Add CPU time into threads view


torusrxxx:
The timing information is available via [GetThreadTimes](https://msdn.microsoft.com/en-us/library/windows/desktop/ms683237(v=vs.85).aspx).

mrexodia:
Might also be interesting to show [QueryThreadCycleTime](https://msdn.microsoft.com/en-us/library/windows/desktop/ms684943(v=vs.85).aspx) in that case.

PR #885, opened 2016-07-25
CPU Time in threads view

torusrxxx:
1. PEB and TEB symbols integrated
2. use MenuBuilder in CallStack view
3. new appearance of CPUSideBar ( #814 )
4. locale dependent compile date to string function
5. CPU Time in threads view ( #842 **might break plugins** )
6. temporarily disable update checker because jenkins is down.
7. other minor fixes

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/885)

### DIFF
--- a/src/dbg/value.cpp
+++ b/src/dbg/value.cpp
@@ -1403,7 +1403,7 @@ bool valapifromstring(const char* name, duint* value, int* value_size, bool prin
                     if(szBaseName)
                     {
                         szBaseName++;
-                        HMODULE hModule = LoadLibraryExW(szModuleName, 0, DONT_RESOLVE_DLL_REFERENCES | LOAD_LIBRARY_AS_DATAFILE);
+                        HMODULE hModule = LoadLibraryExW(szModuleName, 0, DONT_RESOLVE_DLL_REFERENCES);
                         if(hModule)
                         {
                             ULONG_PTR funcAddress = (ULONG_PTR)GetProcAddress(hModule, name);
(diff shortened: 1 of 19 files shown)


src/dbg/value.cpp
```diff
@@ -1702,6 +1723,8 @@ bool valfromstring_noexpr(const char* string, duint* value, bool silent, bool ba
             *isvar = true;
         return true;
     }
+    else if(valenvfromstring(string, value)) //environment block
```
mrexodia: Hm, probably I'll move this to `exprfuncs.cpp` so you can use `teb()` and `peb()`... Doesn't hurt to have though.

src/gui/Src/Utils/StringUtil.cpp
```diff
@@ -224,4 +226,23 @@ QString ToDateString(const QDate & date)
         "Dec"
     };
     return QString().sprintf("%s %d %d", months[date.month() - 1], date.day(), date.year());
+    */
+    return QLocale(QString(currentLocale)).toString(date);
```
mrexodia: I actually did this for the reason of bug reporting. I want my date format in every screenshot so I can see what version people are using.

src/gui/Src/Utils/StringUtil.h
```diff
@@ -30,7 +31,7 @@ static QString ToPtrString(duint Address)
 
 static QString ToLongLongHexString(unsigned long long Value)
 {
-    char temp[32];
+    char temp[33];
```
mrexodia: This already was double the maximum needed size (16 characters).

src/gui/Src/Utils/StringUtil.cpp
torusrxxx: In fact it is also used in the thread creation time, that's why I want locale-dependent date formatting.

mrexodia: Yeah then use locale dependent dates in the thread creation 👍

PR #892, opened 2016-07-27
stack frame lines similar to the ones in ollydbg

torusrxxx:
x64dbg-stack-frame-line

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/892)

### DIFF
--- a/src/gui/Src/Gui/CPUStack.cpp
+++ b/src/gui/Src/Gui/CPUStack.cpp
@@ -751,7 +751,7 @@ void CPUStack::gotoExpressionSlot()
     mGoto->validRangeEnd = base + size;
     mGoto->setWindowTitle(tr("Enter expression to follow in Stack..."));
     if(mGoto->exec() == QDialog::Accepted)
-        DbgCmdExec(QString("sdump \"%s\"").arg(mGoto->expressionText).toUtf8().constData());
+        DbgCmdExec(QString("sdump \"%1\"").arg(mGoto->expressionText).toUtf8().constData());
 }
 
 void CPUStack::gotoPreviousSlot()
--- a/src/gui/Src/Gui/CPUStack.cpp
+++ b/src/gui/Src/Gui/CPUStack.cpp
@@ -59,6 +59,7 @@ void CPUStack::updateColors()
     backgroundColor = ConfigColor("StackBackgroundColor");
     textColor = ConfigColor("StackTextColor");
     selectionColor = ConfigColor("StackSelectionColor");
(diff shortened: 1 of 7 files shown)


src/gui/Src/Gui/CPUStack.cpp
```diff
@@ -406,7 +424,80 @@ QString CPUStack::paintContent(QPainter* painter, dsint rowBase, int rowOffset,
...
         return QString();
     }
-    return HexDump::paintContent(painter, rowBase, rowOffset, col, x, y, w, h);;
+    else if(col == 1) // paint stack data
+    {
+        // get call stack
+        DBGCALLSTACK callstack;
+        memset(&callstack, 0, sizeof(DBGCALLSTACK));
+        // TODO: need to cache the call stack data.
+        DbgFunctions()->GetCallStack(&callstack);
```
mrexodia: I think the performance loss of getting the callstack on every repaint will be very noticeable, especially with the unpredictable performance of dbghelp. Possibly this needs to be integrated in the dbg where the callstack is retrieved from a thread. If I have time tonight I will check if this is easy to do and do it myself.

torusrxxx: In fact I have not observed a significant performance loss. After all there is only a few lines in the CPUStack that need to repaint.

mrexodia: Try loading some symbols for all dlls and then stepping deep in a call
stack.

mrexodia: Hm, possibly I cached this function some time ago so that means it won't be
a problem👍

src/gui/Src/Gui/CPUStack.cpp
```diff
@@ -555,10 +751,7 @@ void CPUStack::gotoExpressionSlot()
     mGoto->validRangeEnd = base + size;
     mGoto->setWindowTitle(tr("Enter expression to follow in Stack..."));
     if(mGoto->exec() == QDialog::Accepted)
-    {
-        QString cmd;
-        DbgCmdExec(cmd.sprintf("sdump \"%s\"", mGoto->expressionText.toUtf8().constData()).toUtf8().constData());
-    }
+        DbgCmdExec(QString("sdump \"%s\"").arg(mGoto->expressionText).toUtf8().constData());
```
mrexodia: I think Qt uses %1 instead of %s right?

mrexodia:
Thanks, everything should be merged and working!

PR #896, opened 2016-07-28
fix app crash

torusrxxx:
caught when documenting this command.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/896)


mrexodia:
Thanks!

Issue #915, opened 2016-08-05
Executable with mismatching architecture appear in most recent files menu


torusrxxx:

torusrxxx:
Assumed to be fixed.

PR #916, opened 2016-08-05
fix issue #915

torusrxxx:
fix issue #915

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/916)

### DIFF
--- a/src/gui/Src/Gui/MainWindow.cpp
+++ b/src/gui/Src/Gui/MainWindow.cpp
@@ -851,7 +851,7 @@ void MainWindow::openFile()
     {
         filename = fileToOpen->text();
     }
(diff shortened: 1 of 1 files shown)


mrexodia:
This is not the right way to fix it. It should call GuiAddRecentFile somewhere in the initialization routine, that should just be moved a little bit.

mrexodia:
https://github.com/x64dbg/x64dbg/blob/development/src/dbg/debugger.cpp#L2279

torusrxxx:
I checked the source. It calls and only calls `GuiAddRecentFile` when it successfully bootstraps. So, we just remove the unnecessary MRU code and it just work better and resolves #915 !

PR #978, opened 2016-08-14
Trace Record in Graph View

torusrxxx:
Make traced instructions green in the graph.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/978)


mrexodia:
Oh that's quite a nice idea! Will definitely help with the understanding.

PR #991, opened 2016-08-18
Fix when double-clicking on the dump when not debugging modify dialog…

torusrxxx:
… would appear

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/991)


mrexodia:
Thanks! Soon I will also run the translation update.

Issue #971, opened 2016-08-13
no IsNop checking on db instructions


mrexodia:
if you mark `90` as `db 90` it will show grey, it shouldnt

PR #993, opened 2016-08-19
fix issue 971 and add thread id function

torusrxxx:
#971

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/993)


mrexodia:
Thanks!

Issue #742, opened 2016-06-11
Translate strings in the DBG


torusrxxx:
There are 2 ways to do it as far as I know. One is windows's `LoadString`(have to update translation templates manually). The another one is to create a dummy Qt project file, mark every string in the DBG with QT_TR_NOOP, and then load the translations dynamically from the GUI where Qt stays.(more complicated, but we can use `lupdate` tools.)

mrexodia:
Yeah I think the dummy qt project is the best option, especially since we
will crowd source the translations.

Issue #313, opened 2015-06-09
Replace fhex and similar macros


Nukem9:
I stole this idea from IDA:

It is to replace
`dprintf("Delete hardware breakpoint failed: "fhex"\n", bp->addr);`
with
`dprintf("Delete hardware breakpoint failed: %fhex\n", bp->addr);`
by using a find & replace type function.

Just a suggestion though.

Nukem9:
https://github.com/cppformat/cppformat

Considered if this issue is ever actually referenced or implemented.

`dprintf("Delete hardware breakpoint failed: {0:fhex}\n", bp->addr);`

mrexodia:
Yea, that's actually a good idea. I already did a lightweight formatting thing (for the log function) https://github.com/x64dbg/x64dbg/blob/master/x64_dbg_dbg/stringformat.cpp

PR #998, opened 2016-08-21
x64dbg is translated to other language

torusrxxx:
Translate the dbg with a dummy qt project. Removed any fhex and fext macro.
I have a language pack for Simplfied Chinese. Contact me if you need it for testing.
Problem: Potential race condition. I use a static buffer to store the translated text not to require the application to free it. If many threads ask for a translated string, they may get the same string. There is big overhead to get Qt to add the strings to the translation template (QT_TRANSLATE_NOOP) that cannot be elimated. Only one context "DBG" is used to simplify the interface. I don't include any Qt file in the project.

Related issues: #742 #313

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/998)


src/dbg/_exports.cpp
```diff
@@ -169,7 +169,7 @@ static bool getLabel(duint addr, char* label)
             duint start;
             if(FunctionGet(addr, &start, nullptr) && addr == start)
             {
-                sprintf_s(label, MAX_LABEL_SIZE, "sub_%" fext "X", start);
```
mrexodia: Actually this changes the behavior of how the string is printed. The idea was unpadded and now it will be padded.

torusrxxx: Yes, it needs more fine tuning. Also the race condition is a major concern
that could lead to crash. By the way, could you check the previous PR so we
can get rid of those irrelavant diffs?

mrexodia: Sorry I'm currently very ill but I will try to check your previous pull
request once I feel a bit better.

torusrxxx:
``fext`` behavior is (mostly) restored now with conditionals. Since this pull request is already very large, should I stop committing every day?

mrexodia:
I'm all good with you updating the pull request, but it is hard for me to
merge if you focus on multiple things in the pull request (I really have to
check this translation, but currently I feel too Ill to check that for
smaller fixes I can easily check it from the Web interface and merge for
there).

PR #1003, opened 2016-08-25
Lock searchbox

Forsari0:
Add **Lock** checkbox right after **mSearchBox** (and before **Regex** checkbox) in **SearchListView** class. 

At interface point of view, it's just a simple checkbox, which locks search textbox if checked. So, the main apply of this option is lay down at implementation POV. 

At that time it's used only in **SymbolView** UI, for purpose - now you can search for, for example, **_Nt_** functions in one module, check **Lock** checkbox, and then go through another modules without tedious retyping of the same search query again and again. Simple, but useful.

> P.S. sorry for empty commit. I'm synch with upstream with PR on Github and that method add this _Update from original_ commits.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1003)


mrexodia:
Looks good, thanks!

Issue #1000, opened 2016-08-22
Vertical scroll singe step size


kkthx:
It should be possible to change/modify number of lines at time of one mouse scroll, like on windows (image attached). Currently it is fixed to 4 lines/notch.

![mousescroll](https://cloud.githubusercontent.com/assets/4583553/17846234/a94e7da2-6847-11e6-8f14-f22d34a212ae.png)

torusrxxx:
HKEY_CURRENT_USER\Control Panel\Desktop : REG_SZ WheelScrollLines
"-1":One screen at a time.

Issue #939, opened 2016-08-13
disable/cancel log redirect


mrexodia:
once the log is redirected to a file I couldnt find a way to disable that

torusrxxx:
You open the log redirection dialog again and click "Cancel" button, then log redirection is cancelled...

mrexodia:
Alright so maybe create a menu option since that was non obvious.

PR #1004, opened 2016-08-25
GUI improvements

torusrxxx:
*  Forward and Backward mouse buttons found in some models of mice.
*  No confirmation when adding functions.
*  Allocate memory from the dump menu.
*  Placeholder tips for learners.
*  disable/cancel log redirect (issue #939)
*  use windows config singe step size (issue #1000)

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1004)

### DIFF
--- a/src/gui/Src/BasicView/AbstractTableView.cpp
+++ b/src/gui/Src/BasicView/AbstractTableView.cpp
@@ -3,6 +3,9 @@
 #include "Configuration.h"
 #include "ColumnReorderDialog.h"
 #include "CachedFontMetrics.h"
+#include <windows.h>
+
+int AbstractTableView::mMouseWheelScrollDelta = 0;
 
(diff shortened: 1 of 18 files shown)


mrexodia:
Hm are you sure it's a good plan to not make the user confirm on function or argument deletion? For adding it makes sense but the user might accidentally delete a tool-analyzed function...

torusrxxx:
I often delete the auto functions because that's inaccurate. Maybe we can
log the old function start and end when deleting the function, so the user
can restore it from the log.

mrexodia:
Alright I will just merge it.

mrexodia:
Should I wait until later today with merging?

torusrxxx:
I'm debugging now. No work is pending.

mrexodia:
Alright, thanks a lot!

PR #1007, opened 2016-08-25
don't halt

torusrxxx:
Don't know if this is the right solution, but at least I can stop the debugger without task manager.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1007)


mrexodia:
Yeah, I think this is a good solution. Sometimes (especially with huge programs) there will be some hangs and nothing happens.

PR #1012, opened 2016-08-27
Signed and Unsigned byte in dump

torusrxxx:
* In highlight mode, the user can click on the registers view to highlight registers.
* The user can also highlight registers with context menu.
* Allow the user to modify EIP like other registers, and EIP can be pushed and poped.
* Signed and Unsigned byte in dump.
* Fixed font size inconsistency when changing font in stack view.
* When changing uppercase setting, token is still highlighted.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1012)


mrexodia:
Thanks a lot!

PR #1020, opened 2016-08-30
Animate into

torusrxxx:
due to "TaskThread", animate appearance is not very smooth compared to ollydbg.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1020)


mrexodia:
Looks great, thanks a lot!

PR #1023, opened 2016-08-30
Use last code page conveniently

torusrxxx:
* Add last code page to the context menu of the dump
* View the source string in selected code page before modifying it in hex editor.
* Try to use less `toLatin1` since it is not compatible with Windows ASCII convention. We should use `toLocal8Bit` instead.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1023)


mrexodia:
Thanks, I hope it will work nicely!

PR #1028, opened 2016-09-01
Various enhancements to thread management

torusrxxx:
*  Don't display any message if the user switches to the same thread.
*  Allow the user to create a thread.
*  Don't run if all the threads are suspended.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1028)


src/dbg/debugger_commands.cpp
```diff
@@ -142,6 +142,11 @@ CMDRESULT cbDebugRun(int argc, char* argv[])
     // Don't "run" twice if the program is already running
     if(dbgisrunning())
         return STATUS_ERROR;
+    if(!ThreaderIsAnyThreadActive())
```
mrexodia: Does this actually make sense to do? Run does not do anything about threads, it just releases the run locks and makes sure `ContinueDebugEvent` is called...

src/dbg/debugger_commands.cpp
```diff
@@ -1997,6 +2002,41 @@ CMDRESULT cbDebugKillthread(int argc, char* argv[])
...
+        return STATUS_ERROR;
+    if(!MemIsCodePage(Entry, false))
+        return STATUS_ERROR;
+    if(argc > 2)
+    {
+        if(!valfromstring(argv[2], &Argument))
+            return STATUS_ERROR;
+    }
+    DWORD ThreadId = 0;
+    if(ThreaderCreateRemoteThread(Entry, true, reinterpret_cast<LPVOID>(Argument), &ThreadId) != 0)
```
mrexodia: Isn't it easier to use `CreateRemoteThread` here directly?

src/dbg/debugger_commands.cpp
```diff
@@ -1997,6 +2002,41 @@ CMDRESULT cbDebugKillthread(int argc, char* argv[])
...
+        dputs(QT_TRANSLATE_NOOP("DBG", "Create thread failed!"));
+        return STATUS_ERROR;
+    }
+    else
+    {
+        char label[MAX_LABEL_SIZE];
+        if(!LabelGet(Entry, label))
+            label[0] = 0;
+#ifdef _WIN64
+        dprintf(QT_TRANSLATE_NOOP("DBG", "Thread %X created at %s %p(Argument=%llX)\n"), ThreadId, label, Entry, Argument);
```
mrexodia: I believe this can be replaced with `Argument=%p`

src/dbg/debugger_commands.cpp
torusrxxx: Because I believe TitanEngine needs to be informed of the new thread.

src/dbg/debugger_commands.cpp
torusrxxx: Sometimes the argument is a small integer.

src/dbg/debugger_commands.cpp
torusrxxx: If you suspend all the threads and click "run", the debugger would enter a "running" state but no thread actually runs.

torusrxxx: If you suspend all the threads and click "run", the debugger would enter a "running" state but no thread actually runs.

src/dbg/debugger_commands.cpp
mrexodia: It doesn't have to be because it will be informed with the `CREATE_THREAD_DEBUG_EVENT`, the code for `ThreaderCreateRemoteThread` is:

```
__declspec(dllexport) ULONG_PTR TITCALL ThreaderCreateRemoteThreadEx(HANDLE hProcess, ULONG_PTR ThreadStartAddress, bool AutoCloseTheHandle, LPVOID ThreadPassParameter, LPDWORD ThreadId)
{
    if(hProcess != NULL)
    {
        if(!AutoCloseTheHandle)
        {
            return (ULONG_PTR)CreateRemoteThread(hProcess, NULL, NULL, (LPTHREAD_START_ROUTINE)ThreadStartAddress, ThreadPassParameter, NULL, ThreadId);
        }
        else
        {
            HANDLE myThread = CreateRemoteThread(hProcess, NULL, NULL, (LPTHREAD_START_ROUTINE)ThreadStartAddress, ThreadPassParameter, NULL, ThreadId);
            EngineCloseHandle(myThread);
            return NULL;
        }
    }
    return NULL;
}
```

src/dbg/debugger_commands.cpp
mrexodia: Right :+1:

src/dbg/debugger_commands.cpp
torusrxxx: Oh, if I create a thread and then suspend the main thread and run, TitanEngine's `ThreaderIsAnyThreadActive` still returns false. This code has to be removed or TitanEngine has to be fixed.

mrexodia: I think this code has to be removed, the 'running' state is true its just that `WaitForDebugEvent` will be stuck until a thread resumes. Maybe we can just warn the user, I think the suspend count is available from the thread module in x64dbg...

src/dbg/debugger_commands.cpp
torusrxxx: The debugging engine can use this information to improve `ThreaderIsAnyThreadActive` implementation so it won't return false when a new thread is about to be created.

mrexodia: In theory it can, but I don't see the benefit in improving titanengine...

torusrxxx: We are already improving titanengine by writing an alternative of it. And since this code is shorter than `CreateRemoteThread` and provides a chance for the engine to recieve the information, why shouldn't we use it? `ThreaderCreateRemoteThread` should be very easy to add in gleebug.

mrexodia: Alright, fair enough :)

PR #1040, opened 2016-09-03
Add Watch Items to Labels in Sidebar

torusrxxx:
Now the user can define custom labels in sidebar by watching it.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1040)

### DIFF
--- a/src/dbg/instruction.cpp
+++ b/src/dbg/instruction.cpp
@@ -207,7 +207,7 @@ CMDRESULT cbInstrMov(int argc, char* argv[])
         int len = (int)dataText.length();
         if(len % 2)
         {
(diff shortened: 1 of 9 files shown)


src/dbg/instruction.cpp
```diff
@@ -2128,19 +2128,19 @@ CMDRESULT cbInstrLog(int argc, char* argv[])
 {
     if(argc == 1)  //just log newline
     {
-        dprintf("\n");
+        dputs("\n");
```
mrexodia: This is incorrect, it will log two newlines.

mrexodia: But doesn't matter I will change it back after merging.

mrexodia:
Thanks!

PR #1043, opened 2016-09-04
Synchronize branch

torusrxxx:
*   New splitter design
*   Conditional DLL breakpoint
*   Split breakpoint-related code into a separate file
*   Don't animate when it stops

Adding more types of breakpoint is not a trival task. Synchronize the branch now so the users can use it and feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1043)


torusrxxx:
While I writing the GUI for DLL breakpoints, I found it very buggy. It may be fixed in a few days.

PR #1045, opened 2016-09-04
restored watchdog submenu

torusrxxx:
Previously, watchdog submenu does not work because it is cleared.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1045)


torusrxxx:
This change may have side effects on other menus. Test is needed.

mrexodia:
Has unintended side effects with the breakpoint menus, but I fixed the issue from the Watchdog view directly.

PR #1048, opened 2016-09-05
DLL breakpoint GUI

torusrxxx:
*  Conditional DLL breakpoint, GUI for conditional DLL breakpoint, DLL breakpoint can be saved to the database.
*  Fixed hit count setting.
*  When not debugging, the user can use "run" button to restart.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1048)


src/gui/Src/Gui/BreakpointsView.cpp
```diff
@@ -146,14 +163,14 @@ void BreakpointsView::reloadData()
     {
         QString addr_text = QString("%1").arg(wBPList.bp[wI].addr, sizeof(dsint) * 2, 16, QChar('0')).toUpper();
         mSoftBPTable->setCellContent(wI, 0, addr_text);
-        mSoftBPTable->setCellContent(wI, 1, QString(wBPList.bp[wI].name));
+        mSoftBPTable->setCellContent(wI, 1, QString::fromUtf8(wBPList.bp[wI].name));
```
mrexodia: Actually you don't have to use `fromUtf8` since that's the default option in Qt5 and in Qt4 we use `QTextCodec::setCodecForCStrings()`

PR #1069, opened 2016-09-09
special drawing if both traced and selected

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1069)


mrexodia:
This is extremely slow, it will deteriorate drawing performance significantly.

torusrxxx:
It can only print gradients up to screen size. That's acceptable.

mrexodia:
I implemented this with color mixing instead, last time there were serious issues with QLinearGradient...

PR #1098, opened 2016-09-18
Registers view enhancements

torusrxxx:
-   Edit MMX registers with the same dialog used to edit XMM registers, with short, long and float data types.
-   Show a tooltip for the flags register to help the user to understand the register.
-   Respect configurated endianess in the edit floating-point register dialog.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1098)


mrexodia:
Great, thanks!

PR #1106, opened 2016-09-20
Fix enable of buttons in favourite tools dialog

torusrxxx:
*   Fix enable of buttons in favourite tools dialog
*   shortcut for blog

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1106)


mrexodia:
Nice!

PR #1124, opened 2016-09-29
Allow debugging dotNET applications

torusrxxx:
Allow the user to debug dotNET applications. Note it only changes the file validation function to let dotNET applications pass the test, it does not introduce a pleasant user experience when debugging dotNET applications. The user is required to start dotNET application with the right architecture of x64dbg, and the user must set x64dbg to pause at system breakpoint. x64dbg cannot pause at the entry point of the dotNET application.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1124)

### DIFF
--- a/src/dbg/_global.cpp
+++ b/src/dbg/_global.cpp
@@ -289,7 +289,9 @@ arch GetFileArchitecture(const char* szFileName)
                 IMAGE_NT_HEADERS* pnth = (IMAGE_NT_HEADERS*)(data + pdh->e_lfanew);
                 if(pnth->Signature == IMAGE_NT_SIGNATURE)
                 {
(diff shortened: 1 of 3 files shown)


mrexodia:
So a .NET executable will work on both architectures or you need to use x64dbg for x64 and x32dbg for x86 .NET files?

torusrxxx:
If the dotNET executable is set to "AnyCPU", you must use x64dbg on 64-bit Windows, or use x32dbg on 32-bit Windows. If the dotNET executable is set to "x86" or "x64", you must use the right version of x64dbg.

mrexodia:
Ah I see, thanks!

PR #1125, opened 2016-09-29
Save Splitter State

torusrxxx:
The state of custom splitters can be saved now. It only works for breakpoints view and handles view. Also adds a "close" event instead of relying on the destructor to save GUI state.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1125)

### DIFF
--- a/src/gui/Src/Gui/SettingsDialog.ui
+++ b/src/gui/Src/Gui/SettingsDialog.ui
@@ -33,7 +33,7 @@
       <bool>true</bool>
      </property>
      <property name="currentIndex">
-      <number>4</number>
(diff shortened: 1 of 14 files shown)


src/gui/Src/Gui/SettingsDialog.ui
```diff
@@ -33,7 +33,7 @@
       <bool>true</bool>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>4</number>
```
mrexodia: You forgot to change this back

PR #1133, opened 2016-10-01
small performance improvements

torusrxxx:
use a faster converter

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1133)

### DIFF
--- a/src/gui/Src/BasicView/Disassembly.cpp
+++ b/src/gui/Src/BasicView/Disassembly.cpp
@@ -440,7 +440,7 @@ QString Disassembly::paintContent(QPainter* painter, dsint rowBase, int rowOffse
             if(i)
                 richBytes.push_back(space);
             auto byte = (unsigned char)dump.at(i);
-            curByte.text = QString("%1").arg(byte, 2, 16, QChar('0')).toUpper();
+            curByte.text = ToByteString(byte);
             DBGPATCHINFO patchInfo;
             if(DbgFunctions()->PatchGetEx(cur_addr + i, &patchInfo))
                 curByte.textColor = byte == patchInfo.newbyte ? mModifiedBytesColor : mRestoredBytesColor;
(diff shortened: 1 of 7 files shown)


src/gui/Src/BasicView/Disassembly.cpp
```diff
@@ -1928,7 +1928,7 @@ QString Disassembly::getAddrText(dsint cur_addr, char label[MAX_LABEL_SIZE])
         *label_ = 0;
     if(label)
         strcpy_s(label, MAX_LABEL_SIZE, label_);
-    return addrText;
+    return std::move(addrText);
```
shamanas: Shouldn't the compiler already be able to perform return value optimization here, since the function exclusively returns 'addrText'?  

In fact, I believe it can actually hurt performance (see [here](https://www.ibm.com/developerworks/community/blogs/5894415f-be62-4bc0-81c5-3956e82276f3/entry/RVO_V_S_std_move?lang=en) and [here](http://stackoverflow.com/questions/17473753/c11-return-value-optimization-or-move)).

torusrxxx:
I remember that any named variable must not be optimized by return value
optimization according to the standard.

shamanas:
This is true when you are returning different variables, for example:  

```c++

QString foo_func() {
    QString foo = get_some_foo();
    if (some_condition) {
        process(foo);
        return foo;
    }

    QString bar = get_bar();
    // or 'return get_bar();'
    return bar;
}

```

This technique is commonly referred to as NRVO (named return value optimization).  

See [here](http://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization/12953145#12953145) for relevant parts of the standard and an example given by it.  

(It also make sense if you consider that, at least with the Win64 calling convention, returning structs really means passing a struct pointer in and modifying it, which means you can ellide creating the local function object directly and write into the passed struct)

src/gui/Src/BasicView/Disassembly.cpp
mrexodia: I removed this optimization I don't think it will hurt performance though.

shamanas: The difference is probably negligible for a small QString, you are right.
It just caught my eye and I was just wondering if there was a legitimate reason for the move.

mrexodia: If you have time, we are in need of code reviews or the whole code 😄

shamanas: I'm actually setting up all the compilation tools you require (since I usually work with gcc and I haven't touched Qt in a while) to tackle a couple of bugs :P

If I see anything along the way that looks like it could be better, I will be writing it down (and I will open a PR).
I don't think there will be much though, the code looks nice from the limited amount I've seen up to now :)

PR #1134, opened 2016-10-01
Keystone: for short jumps, reject commands larger than 2 bytes

RaMMicHaeL:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1134)

### DIFF
--- a/src/dbg/assemble.cpp
+++ b/src/dbg/assemble.cpp
@@ -42,7 +42,7 @@ namespace Keystone
         return nullptr;
     }
 
-    static bool StrDel(char* Source, char* Needle, char StopAt = '\0')
+    static bool StrDel(char* Source, const char* Needle)
     {
         // Find the location in the string first
         char* loc = stristr(Source, Needle);
@@ -53,7 +53,7 @@ namespace Keystone
         // "Delete" the word by shifting it over
         auto needleLen = strlen(Needle);
 
-        memcpy(loc, loc + needleLen, strlen(loc) - needleLen + 1);
+        memmove(loc, loc + needleLen, strlen(loc) - needleLen + 1);
 
         return true;
     }
@@ -68,7 +68,7 @@ namespace Keystone
             sep = strstr(XEDParse->instr, "\n");
         if(sep)
             *sep = '\0';
-        StrDel(XEDParse->instr, "short ");
+        bool short_command = StrDel(XEDParse->instr, "short ");
 
         ks_engine* ks;
         ks_err err = ks_open(KS_ARCH_X86, XEDParse->x64 ? KS_MODE_64 : KS_MODE_32, &ks);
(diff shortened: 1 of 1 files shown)


RaMMicHaeL:
For asmjit, I've created an issue: https://github.com/asmjit/asmtk/issues/4

mrexodia:
Thanks a lot!

mrexodia:
I can also just copy your patch to
https://github.com/x64dbg/asmjit_xedparse/blob/master/asmjit/XEDParseEmulator.cpp

RaMMicHaeL:
You can do this as a temporary solution, until https://github.com/asmjit/asmjit/issues/138 is properly solved.
Frankly, I don't know what are the advantages of asmjit over Keystone, or vice versa. They both seem like high quality projects, and Keystone seems to work pretty well with Multiline Ultimate Assembler.
For XEDParse, there are still commands which it doesn't support. For example, a command which I found in a short testing session: `mov edx, dword ptr [0x7ffe0330]`

mrexodia:
@RaMMicHaeL what exactly is the problem with `mov edx, dword ptr [0x7ffe0330]` it outputs to `8B 15 30 03 FE 7F` to me which appears to be fine... I will add the workaround for short jumps in asmjit. Keystone and asmjit are just both supported. Currently the advantage of asmjit is that it supports symbols but keystone is adding support too. Another advantage of asmjit is that it's smaller but that's hardly relevant for x64dbg 😄

RaMMicHaeL:
Here's what I see...

mrexodia:
Oh yeah I tried on 32 bits.

RaMMicHaeL:
It's fairly easy to find these kinds of issues: disassemble a large block with Multiline Ultimate Assembler, and then assemble it back. Here's another one: `and rax, 0xfffffffffffff000`

mrexodia:
Yeah they are easy to find maybe @Nukem9 should check them.

mrexodia:
@RaMMicHaeL I have updated asmjit to error on short jumps it should be deployed some time soon.

RaMMicHaeL:
It works perfectly, thanks!

RaMMicHaeL:
It works perfectly, thanks!

One additional thing that I'm not sure whether it should be fixed in the debugger or in the plugin:
Some commands, especially which use addressing, disassemble differently in the plugin. For example:
pic

The plugin uses the `DisasmFast` function. Is that what it should be using?

mrexodia:
Yeah the rip-relative thing is just a hack, see
https://github.com/x64dbg/x64dbg/blob/development/src/gui/Src/Gui/CPUDisassembly.cpp#L939

RaMMicHaeL:
I guess that it would be best to make the hacks and the customization (such as uppercase disasm) available to plugins, too.

mrexodia:
They are you can use GuiDisasmGet if you want the exact representation the
GUI has with labels and uppercase 👍🏻

RaMMicHaeL:
I didn't find that one.
https://github.com/x64dbg/x64dbg/search?utf8=%E2%9C%93&q=GuiDisasmGet&type=Code

mrexodia:
I meant GuiGetDisassembly but I wouldn't recommend using it because it
includes module names and labels and such.

RaMMicHaeL:
OK, so my suggestion still applies. I think that plugins should be able to get the same disasm that the user gets when pressing space (which is with the hack, and with case customization, but without labels and such). If that's not going to happen, I might copy the hack to the plugin one day, but this solution is not very elegant.

mrexodia:
Yeah I agree that it would be the best thing. Please open a new issue
though I will see if I can implement it.

RaMMicHaeL:
> You can do this as a temporary solution, until asmjit/asmjit#138 is properly solved.

The issue was fixed in asmjit, so you can revert the hack once you update the asmjit version.

mrexodia:
Alright cool!

PR #1135, opened 2016-10-01
Added a plugin callback that allows filtering dbghelp emitted symbols.

### LINKED ISSUE #445: symbol filter callback
See issue #431 for more information (basically allow plugins to filter the symbols emitted from dbghelp to improve readability in a certain domain like Qt)

shamanas:
Closes #445.
Tested with a simple plugin that filters out a specific symbol.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1135)

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -111,7 +111,10 @@ static bool shouldFilterSymbol(const char* name)
         return true;
     if(strstr(name, "__imp_") == name || strstr(name, "_imp_") == name)
         return true;
-    return false;
+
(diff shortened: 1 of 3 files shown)


shamanas:
Could be modified to allow the plugin to provide an alternative instead of just filtering it (although I don't know if this is already achievable and it would probably be cleaner in another callback processed after the filtering happens).

mrexodia:
@Nukem9 could you verify that this indeed solves the issue we talked about? I don't quite remember what was supposed to be implemented.

mrexodia:
Merged, thanks!

PR #1139, opened 2016-10-02
Add length specifier for string formatter

Forsari0:
The syntax for that addition is `{s|expression:}`
For example, now we can create conditional breakpoints on **memcpy** function and get string even without null terminator on end:
`log "Text without \0 from memcpy - {s|r8:rdx}"`

One change, that _can_ _possibly_ hit the performance, is exchange of `isasciistring` and `isunicodestring` function calls at `disasmgetstringat` function. If it's true, I can try to figure it out.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1139)

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -899,10 +899,13 @@ BRIDGE_IMPEXP void DbgClearAutoFunctionRange(duint start, duint end)
     _dbg_sendmessage(DBG_DELETE_AUTO_FUNCTION_RANGE, (void*)start, (void*)end);
 }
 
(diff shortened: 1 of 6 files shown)


Forsari0:
This changes can badly hit performance of `DbgGetStringAt` and related functions, so, for now, it's not mergeable in the current state.

mrexodia:
I have implemented [_plugin_registerformatfunction](https://github.com/x64dbg/x64dbg/blob/development/src/dbg/_plugins.h#L302) @Forsari0 this will allow you to create custom log functions such as this one. See #1336 for more details.

PR #1142, opened 2016-10-03
GUI: Add undo register context menu action

### LINKED ISSUE #966: undo register changes
changes to registers are marked in red, but there is no option to restore the old value

Atvaark:
This new action will restore the changes made to a register to the previous value.

There's a small shortcoming when a register is changed multiple times (e. g. first by an instruction then by the user). That can be fixed later on by saving the changes to a stack and implementing proper undo/redo.

Closes #966

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1142)


mrexodia:
Thanks a lot!

PR #1153, opened 2016-10-08
GUI: Add hide tab action

Atvaark:
* This action will close the current tab and open the tab to the right of the closed one.
* Default hotkey is Ctrl+W.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1153)


mrexodia:
Is it also possible to get the tab back?

Atvaark:
Yeah it's possible.
Restoring a tab will add it back as the rightmost position on the bar. (Just like when you close  the window of a detached tab).

Maybe we could insert the detached/closed tabs in same position they previously were in (if possible). e. g. Visual Studio tries to restore hidden windows that way.

mrexodia:
Thanks a lot, very useful additions!

PR #1164, opened 2016-10-16
DBG: add autocomment for call $0 (closes #940)

### LINKED ISSUE #940: add autocomment for call $0
the sequence `E8 00 00 00 00` should get a comment that it's calling the next instruction.

joesavage:
I'm a first time contributor and am not overly familiar with x86 assembly, but I think this should do the trick. (Though perhaps `instr.arg[0].segment` also needs checking, I'm not totally sure.)

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1164)


joesavage:
Good to hear! You may well be right about that. I was looking through the `instr.arg[0].x` values of various `call` instructions and saw some with differing `memvalue` and `value` members -- not being sure about what this really meant, I decided to check that the `memvalue` was zero just to be on the safe side.

mrexodia:
Yeah, perhaps memvalue should be checked anyway I will do some more tests.

PR #1168, opened 2016-10-18
Don't use the word "toggle".

torusrxxx:
As a guideline for GUI design, don't use "toggle","switch","switch view" or any word like that. I'm really confused about which view is being switched to(in memory map) and which value does it toggle.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1168)


justanotheranonymoususer:
These are checkboxes, so just removing the "Toggle" word is enough.

On the other hand, there's a problematic menu in the memory map:
http://i.imgur.com/kSv8Sat.png

mrexodia:
I agree with @justanotheranonymoususer here. I added checkable actions to avoid this confusion, but the word 'toggle' can be removed.

PR #1173, opened 2016-10-18
issue 858

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1173)

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -32,6 +32,7 @@
 #include "argument.h"
 #include "watch.h"
 #include "animate.h"
+#include "TraceRecord.h"
 
 static bool bOnlyCipAutoComments = false;
 static duint cacheCflags = 0;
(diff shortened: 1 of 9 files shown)


src/dbg/_exports.cpp
```diff
@@ -1272,6 +1273,15 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
     }
     break;
 
+    case DBG_SELCHANGED:
+    {
+        PLUG_CB_SELCHANGED plugSelChanged;
+        plugSelChanged.hWindow = (int)param1;
+        GuiSelectionGet(GUI_DISASSEMBLY, &plugSelChanged.selection);
```
mrexodia: This will take at least 1ms on every selection change. I think it's better to have `DbgSelChanged(int hWindow, duint addr)` and if plugins need the size (almost never) they can query it manually...

torusrxxx:
Selection change is infrequent. It can only be changed in a mouse release
event or mouse move event, which has a minimum delay of 1/30 seconds.

PR #1177, opened 2016-10-20
Added copy selection to file options for CPUDisassembly.

### LINKED ISSUE #1096: Copying CPU window content to file.
Hello.

Can somebody implement this future? I just want to copy all selected data in CPU window to file. I can't do this using clipboard, because sometimes CPU window contains too much data.

I checked the sources. It have to be an easy task. I can't do this myself. Still using XP...

shamanas:
Closes #1096.

The gif is kinda shitty this time but have it anyways :P 

out

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1177)

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -1,4 +1,6 @@
 #include <QMessageBox>
+#include <QFileDialog>
+#include <QFile>
 #include <QDesktopServices>
 #include <QClipboard>
 #include "CPUDisassembly.h"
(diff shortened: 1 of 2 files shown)


mrexodia:
Hi, while this solves the original issue partially it does not resolve it completely. The problem is that the amount of disassembly does not fit in memory. It should basically be put in a buffer and flushed to the file every 10 mb or so.

shamanas:
@mrexodia  
Ah, gotcha, I didn't get that from the issue.  
I will implement it in a bit!

mrexodia:
Yeah sorry about not making that clear....

shamanas:
Flushing after every instruction is probably a bit too wasteful but I couldn't find any convenient way of querying the write buffer's current size on the Qt docs and it seems to work (tested with 303K instructions, it took a bit but worked after all :P)

0x900x90:
In my function i flushed the buffer every 4 Mb. I think it's OK. But this value could be increased.

shamanas:
@0x900x90  
I looked at your code and wanted to get rid of duplication since copySelectionToFileSlot was basically copy pasted, so I initially just went with returning a QString.  

Perhaps building a QString in the function and committing it to the stream in 4Mb chunks is a good way to avoid duplication as well as too many flushes.
I may modify the code tomorrow and do it that way :)

mrexodia:
Okay looks good, thanks!

PR #1194, opened 2016-10-24
VS2015 fixes

Mattiwatti:
I know x64dbg is meant to be compiled with VS2013, but it's only a few 'const's away from compiling in VS2015 too ;)
Also added a new annoying file type to gitignore

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1194)


mrexodia:
Thanks!

PR #1232, opened 2016-11-08
GUI: Add "Copy RVA" shortcut

ecx86:
Adds a shortcut for ActionCopyRva, set to nothing by default.

Also makes Shift-Middle Click copy the RVA instead of the address.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1232)


mrexodia:
Looks good, thanks a lot!

PR #1243, opened 2016-11-12
decode function offset in stack

torusrxxx:
x64dbg-decode-stack-function-offset

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1243)

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -160,7 +160,7 @@ static bool getLabel(duint addr, char* label, bool noFuncOffset)
         {
             if(addr != 0 && ModEntryFromAddr(addr) == addr)
             {
-                strcpy_s(label, MAX_LABEL_SIZE, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "EntryPoint")));
+                strcpy_s(label, MAX_LABEL_SIZE, "EntryPoint");
                 return true;
             }
             duint start;
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -117,11 +117,12 @@ static bool shouldFilterSymbol(const char* name)
     return filterInfo.retval;
 }
 
-static bool getLabel(duint addr, char* label)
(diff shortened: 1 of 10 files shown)


src/dbg/_exports.cpp
```diff
@@ -157,24 +158,35 @@ static bool getLabel(duint addr, char* label)
         }
         if(!retval)  //search for module entry
         {
-            duint entry = ModEntryFromAddr(addr);
-            if(entry && entry == addr)
+            if(addr != 0 && ModEntryFromAddr(addr) == addr)
             {
-                strcpy_s(label, MAX_LABEL_SIZE, "EntryPoint");
-                retval = true;
+                strcpy_s(label, MAX_LABEL_SIZE, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "EntryPoint")));
```
mrexodia: This isn't actually a translatable thing. `module:EntryPoint` is a pseudo label you can use to get the entry point.

src/dbg/_exports.cpp
```diff
@@ -157,24 +158,35 @@ static bool getLabel(duint addr, char* label)
...
         if(!retval)  //search for module entry
         {
-            duint entry = ModEntryFromAddr(addr);
-            if(entry && entry == addr)
+            if(addr != 0 && ModEntryFromAddr(addr) == addr)
             {
-                strcpy_s(label, MAX_LABEL_SIZE, "EntryPoint");
-                retval = true;
+                strcpy_s(label, MAX_LABEL_SIZE, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "EntryPoint")));
+                return true;
```
mrexodia: This is also incorrect. The if statement should fall through so other types can be queried at the same time and also allow plugins to use `CB_ADDRINFO`.

mrexodia: Oh yeah you're right I misread the context, sorry.

mrexodia:
Oh sorry, I forgot to merge it yesterday...

Issue #1270, opened 2016-11-18
Strange behaviour of search commands


torusrxxx:
When I use "findall a.exe:EntryPoint, 50, 100" command, it can find 5000 matches!(The selected range only has 256 bytes)
Most of the search commands don't use the start and size arguments properly. And the supposed behaviour is not documented correctly for every search command either.

mrexodia:
The command appears to work fine here.

mrexodia:
Nevermind that, for `findall` it's definitely broken.

Issue #1255, opened 2016-11-17
Shortcuts for "tiit", "tibt" etc


torusrxxx:
I usually execute them alternatively.

mrexodia:
What about using a favorite command?

torusrxxx:
Yes I've bind them to "5" and "6".

PR #1272, opened 2016-11-18
gui improvements

### LINKED ISSUE #1260: Shortcuts for go to previous/next/current stack frame in stack view

### LINKED ISSUE #1264: "Follow in memory map" in modules view

torusrxxx:
* 9 more shortcut configurations (fix #1260 #1255)
* Follow in memory map in stack and modules view(fix #1264). Fixed "Follow in memory map" in stack view to follow the content of the stack in the memory map, instead of the memory page of the stack before.
*  Search for various things in the current function. Due to #1270, this submenu is disabled currently.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1272)

### DIFF
--- a/src/gui/Src/Gui/CPUStack.cpp
+++ b/src/gui/Src/Gui/CPUStack.cpp
@@ -218,13 +218,13 @@ void CPUStack::setupContextMenu()
     gotoMenu->addAction(makeShortcutAction(DIcon("geolocation-goto.png"), tr("Go to &Expression"), SLOT(gotoExpressionSlot()), "ActionGotoExpression"));
 
     //Go to Base of Stack Frame
-    gotoMenu->addAction(makeAction(DIcon("neworigin.png"), tr("Go to Base of Stack Frame"), SLOT(gotoFrameBaseSlot())));
+    gotoMenu->addAction(makeShortcutAction(DIcon("neworigin.png"), tr("Go to Base of Stack Frame"), SLOT(gotoFrameBaseSlot()), "ActionGotoBaseOfStackFrame"));
 
     //Go to Previous Frame
-    gotoMenu->addAction(makeAction(DIcon("previous.png"), tr("Go to Previous Stack Frame"), SLOT(gotoPreviousFrameSlot())));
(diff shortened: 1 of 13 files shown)


src/gui/Src/Gui/CPUStack.cpp
```diff
@@ -1005,7 +1005,7 @@ void CPUStack::dbgStateChangedSlot(DBGSTATE state)
 
 void CPUStack::followInMemoryMapSlot()
 {
-    DbgCmdExec(QString("memmapdump %1").arg(ToHexString(rvaToVa(getInitialSelection()))).toUtf8().constData());
+    DbgCmdExec(QString("memmapdump [%1]").arg(ToHexString(rvaToVa(getInitialSelection()))).toUtf8().constData());
```
mrexodia: I think this behavior is better, but it should then replace `Follow in Memory Map` with `Follow DWORD in Memory Map` and also check if that DWORD is a valid memory address before adding it to the menu.

torusrxxx:
* Added Variables View (varlist)
* Use the same shortcut for graph refresh as in other views
* Changed to "Follow DWORD in Memory Map"
* Show size of stack call frame in call stack view

src/gui/Src/Gui/MainWindow.cpp
```diff
@@ -402,7 +403,7 @@ void MainWindow::setupLanguagesMenu()
     languageMenu->setIcon(DIcon("codepage.png"));
     QLocale enUS(QLocale::English, QLocale::UnitedStates);
     QString wCurrentLocale(currentLocale);
-    QAction* action_enUS = new QAction(QString("[%1] %2 - %3").arg(enUS.name()).arg(enUS.nativeLanguageName()).arg(enUS.nativeCountryName()), languageMenu);
+    QAction* action_enUS = new QAction(QString("[%1]%2 - %3").arg(enUS.name()).arg(enUS.nativeLanguageName()).arg(enUS.nativeCountryName()), languageMenu);
```
mrexodia: I changed this previously it's more natural to have a space there.

torusrxxx: It is inconsistent with the next string

torusrxxx: So I changed it.

PR #1282, opened 2016-11-22
Show suspected call stack frame, and fix a bug

torusrxxx:
Bug: log hyperlink never works on 32-bit platform

Feature: Show suspected/active call stack frame. Active stack frames are the same as the current ones. When showing suspected call stack frames, then instead of using the structure of call stack frame to walk the stack, it will search for the entire stack to find an address that is possible as a return address. This will typically result in more call stack frames available.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1282)

### DIFF
--- a/src/dbg/stackinfo.cpp
+++ b/src/dbg/stackinfo.cpp
@@ -17,6 +17,7 @@
 
 using SehMap = std::unordered_map<duint, STACK_COMMENT>;
 static SehMap SehCache;
+bool ShowSuspectedCallStack;
 
 void stackupdateseh()
 {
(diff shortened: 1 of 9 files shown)


src/dbg/stackinfo.cpp
```diff
@@ -195,7 +239,7 @@ void stackgetcallstack(duint csp, std::vector<CALLSTACKENTRY> & callstackVector,
             return;
         }
         callstackVector.clear();
-        return;
+        //return;
```
torusrxxx: It is because updating the setting in the GUI (and then clearing the cache via `stackupdatesettings()`) will empty the call stack view. Any way not to update but not empty that view?

mrexodia:
❤️

PR #1351, opened 2016-12-07
Call DebugRemoveBreakpoints before DbClear, resolves #1310

### LINKED ISSUE #1310: Wrong behavior of `dbclear` command (regarding breakpoints)
The `dbclear` command clears the entire database as expected, but that's not reflected on the breakpoints of the current debugging session, you have to restart the dubugee to see the effect,
i.e. they're not deleted until you restart the degugee.

Steps to reproduce:

1. Debug anything and place some random software/hardware BPs.
2. Type `dbclear` in the command bar.
3. Press run and the debugger will still pause at those BPs (despite them being deleted).
Also the log says `Breakpoint reached not in list!` when you reach any of them.

It's not a big deal or anything, but rather confusing, because the debugger will pause/break at non-existent BPs.

ner0x652:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1351)


ner0x652:
Closing it because there are other (past)commits non related to this issue.

PR #1355, opened 2016-12-08
config command

torusrxxx:
# config
Get or set the configuration of x64dbg. It can also be used to load and store configuration specific to the script in the configuration file of x64dbg.

## arguments

``arg1``: Section name of the INI file.

``arg2``: Key name of the INI file.

\[``arg3``\]: Optional new value of the configuration. If this argument is set to a number, it will be stored in the configuration file and ``$result`` is not updated. If this argument is not set, the current configuration will be read into ``$result``.

## results
This command sets ``$result`` to the current configuration number if ``arg3`` is not set.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1355)

### DIFF
--- a/src/dbg/commands/cmd-misc.cpp
+++ b/src/dbg/commands/cmd-misc.cpp
@@ -582,3 +582,44 @@ bool cbInstrMnemonicbrief(int argc, char* argv[])
     dputs(MnemonicHelp::getBriefDescription(argv[1]).c_str());
     return true;
 }
+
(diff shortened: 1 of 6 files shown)


torusrxxx:
Add a new button in attach dialog to find the process of a window.

mrexodia:
Thanks!

torusrxxx:
A fix is pushed immediately after this branch is merged to get "attach to hidden process" really working.

PR #1356, opened 2016-12-08
Extend Follow in Disassembler button in Calculator window

ner0x652:
In default theme there is no problem:

dev

In other themes(like dark theme) the problem reappears:

dark_theme

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1356)

### DIFF
--- a/src/gui/Src/Gui/CalculatorDialog.ui
+++ b/src/gui/Src/Gui/CalculatorDialog.ui
@@ -28,7 +28,7 @@
     <rect>
      <x>10</x>
      <y>240</y>
-     <width>141</width>
+     <width>181</width>
      <height>23</height>
     </rect>
    </property>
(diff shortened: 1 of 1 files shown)


mrexodia:
Hm? Why was it closed?

ner0x652:
@mrexodia Remembered that yesterday we talked about this on IRC, and you pointed out that the issue is the font. Should it be reopened?

mrexodia:
I would have to check on a PC. I only mentioned that putting the buttons in
a layout might solve the issues because then they will automatically adjust
on font changes.

Button - button - horizontal spacer

Those inside a horizontal layout should solve the issue...

ner0x652:
@mrexodia Didn't know about that. It would be great to do so, I'll change it.

mrexodia:
The layout editor can be a pain sometimes so good luck 😀

PR #1357, opened 2016-12-08
look at the symbol log font #1329

### LINKED ISSUE #1329: look at the symbol log font
It appears that the font of the symbol view is incorrect, it should be the same as the log font.

changeofpace:
fixes #1329 

This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1357)


mrexodia:
Thanks a lot!

PR #1358, opened 2016-12-08
#1312 disable exinfo on exceptions

### LINKED ISSUE #1312: disable exinfo on exceptions
There should be an option (or heuristic) to disable calling the `exinfo` command on exceptions every time (it can slow down applications with lots of C++ exceptions).

Perhaps disable it if Shift+F9 is used...

changeofpace:
fixes #1312

Added an engine setting, "Enable Verbose Exception Logging", which determines if the exinfo command is executed in the exception callback.  Enabled by default.

preview

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1358)

### DIFF
--- a/src/gui/Src/Gui/SettingsDialog.ui
+++ b/src/gui/Src/Gui/SettingsDialog.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>381</width>
-    <height>484</height>
+    <height>508</height>
    </rect>
   </property>
   <property name="sizePolicy">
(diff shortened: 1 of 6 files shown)


src/gui/Src/Gui/SettingsDialog.ui
```diff
@@ -33,7 +33,7 @@
       <bool>true</bool>
      </property>
      <property name="currentIndex">
-      <number>0</number>
```
mrexodia: The default tab changed with this. Apart from that it looks great!

mrexodia:
Looks great, thanks!

PR #1361, opened 2016-12-10
added extra info prints for access violation exceptions for the exinfo command

### LINKED ISSUE #1331: show more information about access violations
See [this](https://msdn.microsoft.com/en-us/library/windows/desktop/aa363082(v=vs.85).aspx) MSDN page, specifically:

> The first element of the array contains a read-write flag that indicates the type of operation that caused the access violation. If this value is zero, the thread attempted to read the inaccessible data. If this value is 1, the thread attempted to write to an inaccessible address. If this value is 8, the thread causes a user-mode data execution prevention (DEP) violation.
> The second array element specifies the virtual address of the inaccessible

changeofpace:
resolves #1331 

I removed the current ExceptionInformation logging code, but it may be a good idea to have some handling for undocumented use of this var.  What do you think?

preview for each access violation case:

access_violation_preview

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1361)


mrexodia:
Thanks!

Issue #1339, opened 2016-12-02
The scroll button of memory tabs


hulucc:
The scroll button of memory tabs keep showing even if there is enough space for all the tabs.

![image](https://cloud.githubusercontent.com/assets/4343900/20826207/40584b9a-b8a4-11e6-8007-3f724fc8bb49.png)

mrexodia:
Yeah I have seen this a few times. Perhaps @Nukem9 knows what's going on here?

hulucc:
I find out that the bug won't happen when you select english as language.
Please select chinese to reproduce it.

PR #1367, opened 2016-12-13
fixed additional buttons on tab bar

torusrxxx:
#1339
x64dbg-no-additional-buttons-on-dump-widget

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1367)


mrexodia:
So what happens if there is not enough room for the tabs?

torusrxxx:
See the window on the right in the picture.

mrexodia:
Okay I wasn't sure if that was the same commit. Thanks!

PR #1369, opened 2016-12-13
Fix failing to read IMAGE_NT_HEADERS when it is at an address greater…

jossgray:
… than 0x1000.

Found that x64dbg was failing to open this executable and reporting  "Invalid PE file!", http://downloads.dell.com/FOLDER04030973M/1/XPS_9550_1.2.16.exe

The problem was that `(size_t)pdh->e_lfanew < readSize` was returning false and so x64dbg was unable to read the file architecture correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1369)


mrexodia:
Thanks! I made slight changes to support a wider variety of executables and also added support for this in the launcher.

PR #1417, opened 2017-01-12
handles and windows view

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1417)

### DIFF
--- a/src/dbg/_dbgfunctions.cpp
+++ b/src/dbg/_dbgfunctions.cpp
@@ -26,6 +26,7 @@
 #include "tcpconnections.h"
 #include "watch.h"
 #include "animate.h"
+#include "thread.h"
 
 static DBGFUNCTIONS _dbgfunctions;
 
@@ -370,4 +371,5 @@ void dbgfunctionsinit()
     _dbgfunctions.DbgGetDebuggeeInitScript = dbggetdebuggeeinitscript;
     _dbgfunctions.EnumWindows = _enumwindows;
     _dbgfunctions.EnumHeaps = _enumheaps;
+    _dbgfunctions.ThreadGetName = ThreadGetName;
 }
--- a/src/dbg/_dbgfunctions.cpp
+++ b/src/dbg/_dbgfunctions.cpp
@@ -295,6 +295,22 @@ static bool _enumtcpconnections(ListOf(TCPCONNECTIONINFO) connections)
     return BridgeList<TCPCONNECTIONINFO>::CopyData(connections, connectionsV);
 }
 
+static bool _enumwindows(ListOf(WINDOW_INFO) windows)
+{
+    std::vector<WINDOW_INFO> windowInfoV;
+    if(!HandlesEnumWindows(windowInfoV))
+        return false;
+    return BridgeList<WINDOW_INFO>::CopyData(windows, windowInfoV);
+}
+
+static bool _enumheaps(ListOf(HEAPINFO) heaps)
+{
+    std::vector<HEAPINFO> heapInfoV;
+    if(!HandlesEnumHeaps(heapInfoV))
+        return false;
+    return BridgeList<HEAPINFO>::CopyData(heaps, heapInfoV);
+}
+
 void dbgfunctionsinit()
 {
     _dbgfunctions.AssembleAtEx = _assembleatex;
@@ -352,4 +368,6 @@ void dbgfunctionsinit()
     _dbgfunctions.AnimateCommand = _dbg_animatecommand;
     _dbgfunctions.DbgSetDebuggeeInitScript = dbgsetdebuggeeinitscript;
     _dbgfunctions.DbgGetDebuggeeInitScript = dbggetdebuggeeinitscript;
+    _dbgfunctions.EnumWindows = _enumwindows;
+    _dbgfunctions.EnumHeaps = _enumheaps;
 }
--- a/src/dbg/_dbgfunctions.h
+++ b/src/dbg/_dbgfunctions.h
@@ -172,6 +172,7 @@ typedef void(*DBGSETDEBUGGEEINITSCRIPT)(const char* fileName);
 typedef const char* (*DBGGETDEBUGGEEINITSCRIPT)();
 typedef bool(*HANDLESENUMWINDOWS)(ListOf(WINDOW_INFO) windows);
 typedef bool(*HANDLESENUMHEAPS)(ListOf(HEAPINFO) heaps);
+typedef bool(*THREADGETNAME)(DWORD tid, char* name);
 
 //The list of all the DbgFunctions() return value.
 //WARNING: This list is append only. Do not insert things in the middle or plugins would break.
@@ -234,6 +235,7 @@ typedef struct DBGFUNCTIONS_
     DBGGETDEBUGGEEINITSCRIPT DbgGetDebuggeeInitScript;
     HANDLESENUMWINDOWS EnumWindows;
     HANDLESENUMHEAPS EnumHeaps;
+    THREADGETNAME ThreadGetName;
 } DBGFUNCTIONS;
 
 #ifdef BUILD_DBG
--- a/src/dbg/_dbgfunctions.h
+++ b/src/dbg/_dbgfunctions.h
@@ -94,6 +94,27 @@ typedef struct
     unsigned int State;
 } TCPCONNECTIONINFO;
 
+typedef struct
+{
+    duint handle;
+    duint parent;
+    DWORD threadId;
+    DWORD style;
+    DWORD styleEx;
+    duint wndProc;//not used yet
+    bool enabled;
+    RECT position;
+    char windowTitle[MAX_COMMENT_SIZE];
+    char windowClass[MAX_COMMENT_SIZE];
+} WINDOW_INFO;
+
+typedef struct
+{
+    duint addr;
+    duint size;
+    duint flags;
+} HEAPINFO;
+
 typedef bool (*ASSEMBLEATEX)(duint addr, const char* instruction, char* error, bool fillnop);
 typedef bool (*SECTIONFROMADDR)(duint addr, char* section);
 typedef bool (*MODNAMEFROMADDR)(duint addr, char* modname, bool extension);
@@ -149,7 +170,11 @@ typedef bool(*MEMISCODEPAGE)(duint addr, bool refresh);
 typedef bool(*ANIMATECOMMAND)(const char* command);
 typedef void(*DBGSETDEBUGGEEINITSCRIPT)(const char* fileName);
 typedef const char* (*DBGGETDEBUGGEEINITSCRIPT)();
+typedef bool(*HANDLESENUMWINDOWS)(ListOf(WINDOW_INFO) windows);
+typedef bool(*HANDLESENUMHEAPS)(ListOf(HEAPINFO) heaps);
 
+//The list of all the DbgFunctions() return value.
+//WARNING: This list is append only. Do not insert things in the middle or plugins would break.
 typedef struct DBGFUNCTIONS_
 {
     ASSEMBLEATEX AssembleAtEx;
@@ -207,6 +232,8 @@ typedef struct DBGFUNCTIONS_
     ANIMATECOMMAND AnimateCommand;
     DBGSETDEBUGGEEINITSCRIPT DbgSetDebuggeeInitScript;
     DBGGETDEBUGGEEINITSCRIPT DbgGetDebuggeeInitScript;
+    HANDLESENUMWINDOWS EnumWindows;
+    HANDLESENUMHEAPS EnumHeaps;
 } DBGFUNCTIONS;
 
 #ifdef BUILD_DBG
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -527,7 +527,19 @@ void DebugUpdateGui(duint disasm_addr, bool stack)
     char threadName[MAX_THREAD_NAME_SIZE + 1] = "";
     if(ThreadGetName(currentThreadId, threadName) && *threadName)
         strcat_s(threadName, " ");
-    sprintf_s(title, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "File: %s - PID: %X - %sThread: %s%X%s")), szBaseFileName, fdProcessInfo->dwProcessId, modtext, threadName, currentThreadId, threadswitch);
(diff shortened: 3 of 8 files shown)


mrexodia:
I just tested this, but the function for `enumHeaps` is very, very, very slow on x64 (it takes minutes) and there seems to be no way to refresh a single view manually.

Perhaps http://securityxploded.com/enumheaps.php will help.

mrexodia:
I checked some more and it looks like the function that `CreateToolhelp32Snapshot` uses to take a snapshot of the heap uses some kind of RPC. This function is `RtlQueryProcessDebugInformation` it uses `NtCreateThreadEx` in the debuggee to call `RtlpQueryProcessDebugInformationRemote` so heap enumeration will never complete if the debuggee is paused.

mrexodia:
I did some more testing and the official API appears to be completely broken. The `Address` field after enumeration is totally broken. It points vaguely inside heap blocks, but inside strings and it doesn't seem to be coherent at all...

Mattiwatti:
That blog entry has the right idea; unfortunately he has some incomplete type info which (for me at least) caused the heap enumeration to crash, and since the code uses hardcoded pointer offsets instead of types, I don't really feel like figuring out the cause.

The Process Hacker headers should help if you're interested in using the method from the blog with proper type info:

[RTL_HEAP_ENTRY](https://github.com/processhacker2/processhacker2/blob/b443aa77ffd5072689f31d1a9e9d1799f8aad523/phnt/include/ntrtl.h#L3224)
[RTL_HEAP_TAG](https://github.com/processhacker2/processhacker2/blob/b443aa77ffd5072689f31d1a9e9d1799f8aad523/phnt/include/ntrtl.h#L3254)
[RTL_HEAP_INFORMATION](https://github.com/processhacker2/processhacker2/blob/b443aa77ffd5072689f31d1a9e9d1799f8aad523/phnt/include/ntrtl.h#L3264) (`DEBUG_HEAP_INFORMATION` in the blog)
[RTL_PROCESS_HEAPS](https://github.com/processhacker2/processhacker2/blob/b443aa77ffd5072689f31d1a9e9d1799f8aad523/phnt/include/ntrtl.h#L3281) (blog: `HeapInfo`)
[RTL_DEBUG_INFORMATION](https://github.com/processhacker2/processhacker2/blob/b443aa77ffd5072689f31d1a9e9d1799f8aad523/phnt/include/ntrtl.h#L3905) (blog: `DEBUG_BUFFER`)

The following ntdll functions are needed:
```C++
NTSYSAPI
PRTL_DEBUG_INFORMATION
NTAPI
RtlCreateQueryDebugBuffer(
	_In_opt_ ULONG MaximumCommit,
	_In_ BOOLEAN UseEventPair
);

NTSYSAPI
NTSTATUS
NTAPI
RtlDestroyQueryDebugBuffer(
	_In_ PRTL_DEBUG_INFORMATION Buffer
);

NTSYSAPI
NTSTATUS
NTAPI
RtlQueryProcessDebugInformation(
	_In_ HANDLE UniqueProcessId,
	_In_ ULONG Flags,
	_Inout_ PRTL_DEBUG_INFORMATION Buffer
);
```

I would say using ntdll functions for heap enumeration is preferable to `CreateToolhelp32Snapshot` and co by far. They may be undocumented but at least they work. (and are orders of magnitude faster.)

Re: `RtlpQueryProcessDebugInformationRemote` and pausing: I know ntdll expressly creates this thread using `NtCreateThreadEx` with `THREAD_CREATE_FLAGS_HIDE_FROM_DEBUGGER` set in order to prevent these kinds of issues. Whether that is enough to prevent all problems I don't know. I recall at least one instance when I found out a program was deadlocked because of an `RtlQueryProcessDebugInformation` call that was issued by another process.

mrexodia:
That remote thread won't execute if the debuggee is in a paused state,
that's the main problem...

Mattiwatti:
Yes, that's what I thought as well. `THREAD_CREATE_FLAGS_HIDE_FROM_DEBUGGER` is very powerful, to the point where a thread with the flag set may even be ignored when a process suspend is issued (though don't quote me on this), but it's not enough to start execution in a process that's already suspended, which would be the case here. I know because I wrote an implementation of `NtCreateThreadEx` for the XP kernel once, ironically with the purpose of banning this flag altogether as a ScyllaHide-type hook, and I ended up having to special-case the ntdll debug heap query thread to essentially give it godmode powers to prevent issues with suspended threads. (and processes, which from the kernel's POV are nothing more than processes containing only suspended threads, though it may lie and say each thread's individual suspend count is 0).

There's no reason why `RtlpQueryProcessDebugInformationRemote` couldn't be implemented out-of-process using `ReadProcessMemory` however. You'd simply have to calculate the offsets, basically like the guy in the blog already did minus the crash, and then issue a `ReadProcessMemory` for each individual node and block. It would be slower, but not by that much; `NtReadVirtualMemory` is very fast (it doesn't really care if the virtual address you're reading is in your own process or another one after it has performed the access checks). This would allow doing the enumeration in the debugger while the debuggee is suspended. It might be preferable to do it this way regardless since it removes the dependency on the ntdll imports. You'd still be reliant on the undocumented heap node structure layout, but that struct hasn't been changed in a looong time. I'm not saying I have a copy of the leaked NT4 source from 1995 because that would be illegal, but if I *did*, I might find the exact same struct as the one linked above in there.

mrexodia:
I looked at the heap structure and enumerating the heap blocks (the bigger
chunks that contain nodes) is really easy. ProcessHacker uses this to show
if a memory region is a heap. It's just the PEB member called ProcessHeaps
and the number of addresses there is also in the PEB. However I cannot find
any kind of logical structure there, done you have a link to more
information on those?

Mattiwatti:
Sure. A `ProcessHeap` in the PEB is what's returned from `RtlCreateHeap`:
```C++
typedef struct _RTL_HEAP_PARAMETERS
{
	ULONG Length;
	SIZE_T SegmentReserve;
	SIZE_T SegmentCommit;
	SIZE_T DeCommitFreeBlockThreshold;
	SIZE_T DeCommitTotalFreeThreshold;
	SIZE_T MaximumAllocationSize;
	SIZE_T VirtualMemoryThreshold;
	SIZE_T InitialCommit;
	SIZE_T InitialReserve;
	PRTL_HEAP_COMMIT_ROUTINE CommitRoutine;
	SIZE_T Reserved[2];
} RTL_HEAP_PARAMETERS, *PRTL_HEAP_PARAMETERS;

NTSYSAPI
PVOID
NTAPI
RtlCreateHeap(
	_In_ ULONG Flags,
	_In_opt_ PVOID HeapBase,
	_In_opt_ SIZE_T ReserveSize,
	_In_opt_ SIZE_T CommitSize,
	_In_opt_ PVOID Lock,
	_In_opt_ PRTL_HEAP_PARAMETERS Parameters
	);
```
You can ignore `CommitRoutine` AFAIK since it's normally set to a function that does `xor eax,eax, retn`.

The fact that `RtlCreateHeap` returns PVOID is telling though; I wouldn't be so sure the internal layout of the struct that is returned is unchanged between OS versions. Your best would probably be starting with this:
[`_HEAP` as defined in ReactOS](https://github.com/reactos/reactos/blob/474b5bea1ee4d8e5638e692bb4541194604f9866/reactos/sdk/lib/rtl/heap.h#L206)
and going from there to see if the definition is still correct. Fortunately it looks like you'll only need the first couple of fields which are least likely to have changed.

Relevant sources:
[RtlCreateHeap](https://github.com/reactos/reactos/blob/474b5bea1ee4d8e5638e692bb4541194604f9866/reactos/sdk/lib/rtl/heap.c#L1221)
...which ends up calling...
[RtlpInitializeHeap](https://github.com/reactos/reactos/blob/474b5bea1ee4d8e5638e692bb4541194604f9866/reactos/sdk/lib/rtl/heap.c#L86)

Mattiwatti:
I just realised the code in the blog entry was based on 32 bit disassembly, so it's no wonder my attempt to run it failed, since I compiled as 64 bit which wouldn't mesh well with the hardcoded offsets. (NB: ReactOS headers are prone to this too: there are often ULONG fields that should be ULONG_PTR or SIZE_T, but this isn't so easy to find out when you're disassembling 32 bit code.)

I recompiled as 32 bit and it worked. Some notes:
- `RtlQueryProcessDebugInformation` works even when the debuggee is paused.
- The results from the two methods are not 100% identical: compare [CreateToolhelp32Snapshot output](http://dumptext.com/gCSRr4jG) and [RtlQueryProcessDebugInformation output](http://dumptext.com/nx3whUh2) ([diff](http://dumptext.com/tBeY2kpL)). The extra blocks `CreateToolhelp32Snapshot` found all looked like valid entries to me. Don't know what's up with that.
- As expected, the `RtlQueryProcessDebugInformation` method is many times faster (basically instant). `CreateToolhelp32Snapshot` stalled x64dbg for a good second even though I was only running calc.exe.

torusrxxx:
I will disable the heaps view for now and properly implement it in future commits.

Mattiwatti:
If you are interested in bringing back the heaps view later, here is a version of the heap block enumeration code from the blog that uses the PH types I posted above:
```C++
bool GetFirstHeapBlock(PRTL_HEAP_INFORMATION curHeapNode, HEAPINFO *hb)
{
  PRTL_HEAP_ENTRY entry = curHeapNode->Entries;

  while ((entry->Flags & 2) == 2)
  {
    hb->index++;
    hb->addr = (duint)((unsigned char*)entry->FirstBlock + curHeapNode->EntryOverhead);
    entry = entry + 1;
    hb->size = entry->Size;
  }

  // Update the flags
  if ((entry->Flags & 0xF1) != 0 || (entry->Flags & 0x200) != 0)
    hb->flags = 1;
  else if ((entry->Flags & 0x20) != 0)
    hb->flags = 4;
  else if ((entry->Flags & 0x100) != 0)
    hb->flags = 2;

  return true;
}

bool GetNextHeapBlock(PRTL_HEAP_INFORMATION curHeapNode, HEAPINFO *hb)
{
  // Make it point to next block address entry
  hb->index++;
  PRTL_HEAP_ENTRY entry = (PRTL_HEAP_ENTRY)(curHeapNode->Entries + hb->index);
  if (entry == NULL)
    return false;

  if ((entry->Flags & 2) == 2)
  {
    do
    {
      // New address = current block address + overhead
      hb->addr = (duint)((unsigned char*)entry->FirstBlock + curHeapNode->EntryOverhead);

      // If all the blocks have been enumerated, exit
      if (hb->index > curHeapNode->NumberOfEntries)
        return false;

      hb->index++;
      hb->size = entry->Size;
    } while ((entry->Flags & 2) == 2);
  }
  else
  {
    // New address = previous address + previous block size
    hb->addr = (duint)(hb->addr + hb->size);
    hb->size = entry->Size;
  }

  // Update the flags
  if ((entry->Flags & 0xF1) != 0 || (entry->Flags & 0x200) != 0)
    hb->flags = 1;
  else if ((entry->Flags & 0x20) != 0)
    hb->flags = 4;
  else if ((entry->Flags & 0x100) != 0)
    hb->flags = 2;

  return true;
}

bool HandlesEnumHeaps(std::vector<HEAPINFO> & heapList)
{
  typedef PRTL_DEBUG_INFORMATION(NTAPI *RTLCREATEDEBUGBUFFER)(ULONG MaximumCommit, BOOLEAN UseEventPair);
  typedef NTSTATUS(NTAPI *RTLQUERYPROCESSDEBUGINFORMATION)(HANDLE UniqueProcessId, ULONG Flags, PRTL_DEBUG_INFORMATION Buffer);
  typedef NTSTATUS(NTAPI *RTLDESTROYQUERYDEBUGBUFFER)(PRTL_DEBUG_INFORMATION Buffer);
  static auto RtlCreateQueryDebugBuffer =
    (RTLCREATEDEBUGBUFFER)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlCreateQueryDebugBuffer");
  static auto RtlQueryProcessDebugInformation =
    (RTLQUERYPROCESSDEBUGINFORMATION)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlQueryProcessDebugInformation");
  static auto RtlDestroyQueryDebugBuffer =
    (RTLDESTROYQUERYDEBUGBUFFER)GetProcAddress(GetModuleHandleW(L"ntdll.dll"), "RtlDestroyQueryDebugBuffer");
  if (RtlCreateQueryDebugBuffer == NULL || RtlQueryProcessDebugInformation == NULL || RtlDestroyQueryDebugBuffer == NULL)
    return false;

  // Create debug buffer
  PRTL_DEBUG_INFORMATION debugInfo = RtlCreateQueryDebugBuffer(0, FALSE);
  if (debugInfo == NULL)
    return false;

  // Get process heap data
  NTSTATUS status = RtlQueryProcessDebugInformation((HANDLE)fdProcessInfo->dwProcessId, /*PDI_HEAPS | PDI_HEAP_BLOCKS*/ 0x14, debugInfo);
  if (status < 0 || debugInfo->Heaps == NULL)
    return false;

  // Go through each of the heap nodes
  PRTL_HEAP_INFORMATION heapNodes = debugInfo->Heaps->Heaps;
  HEAPINFO hb;
  for (unsigned int i = 0; i < debugInfo->Heaps->NumberOfHeaps; i++)
  {
    // Enumerate all blocks within this heap node
    memset(&hb, 0, sizeof(hb));
    if (GetFirstHeapBlock(&heapNodes[i], &hb))
    {
      do
      {
        heapList.push_back(hb);
      } while (GetNextHeapBlock(&heapNodes[i], &hb));
    }
  }

  RtlDestroyQueryDebugBuffer(debugInfo);
  return true;
}
```
Since the hardcoded offsets are gone, this now works on 64 bit. Unfortunately the output is still close to but not exactly the same as that of `CreateToolhelp32Snapshot`. I've narrowed this down to blocks following LF32_FREE blocks with size 0 (they cause the rest of the blocks in the node to be skipped), but haven't been able to fix it.

torusrxxx:
Do you know whether RtlQueryProcessDebugInformation would work if the user
wants to browse the heap when the debuggee is paused? If it doesn't we must
re-implement this function in x64dbg or GleeBug with ReadProcessMemory,
which is very hard to do.

Mattiwatti:
Yes, contrary to what I expected this works when the debuggee is paused. It's the power of `NtCreateThreadEx` with `THREAD_CREATE_FLAGS_HIDE_FROM_DEBUGGER`.

torusrxxx:
Thanks for your generous contribution. I will enable the heaps view as soon as possible.

PR #1426, opened 2017-01-18
fix MemRead to only return true if NumberOfBytesRead is equal to the requested read size

changeofpace:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1426)


changeofpace:
this fix is causing "find string references in this module" to fail.

mrexodia:
Perhaps because that tries to read the whole module memory range at once...

PR #1440, opened 2017-01-28
treat fs / gs usage as unusual

### LINKED ISSUE #1332: hint users about fs/gs usages
(outside of system modules)?

Usually the presence of fs/gs signals exception handlers (x32 only) or some PEB/TEB stuff.

changeofpace:
closes #1332

preview:

fsgs1
fsgs2

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1440)


mrexodia:
Looks good. GS is not unusual on 32 bit though, so probably it's better to use `ArchValue(FS, GS)` to only include the relevant unusual instruction...

mrexodia:
Thanks, merged and manually updated.

PR #1460, opened 2017-02-10
Do not add breakpoint if SetBPX fails.

changeofpace:
Normal breakpoints are not removed if SetBPX fails in cbDebugSetBPX.  The invalid breakpoint appears in the breakpoint tab and is highlighted in the disassembly view:

attempting to set the bp:
invalid_bp_1

removing the bp:
invalid_bp_2

I removed the MemIsValidReadPtr check because BpAdd will fail if MemIsValidReadPtr(addr) is false.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1460)


mrexodia:
Thanks!

PR #1470, opened 2017-02-20
Add wait reason to threads list

Mattiwatti:
I noticed threads were always being shown as being in the "executive" wait state, which I found a bit silly because WrExecutive is normally reserved for the kernel. Taking a look at the source I found ThreadGetWaitReason was actually unimplemented :P So that would explain it.

This is an implementation that uses NtQuerySystemInformation(SystemProcessInformation), the only possible way to get the thread wait reason.

It is monstrously inefficient because it gets info on all running processes AND all of their threads (even though we only need one piece of info on one thread). It could be made more efficient by at least caching the results for the whole process so that ThreadGetWaitReason would only have to be called once per process instead of for every thread. But that would require some restructuring of how the thread info is queried right now. There's also no reason to loop over the threads of processes we already know aren't the debuggee, but I found no easy way to get the PID from the THREADALLINFO that ThreadGetList makes. There is GetProcessIdOfThread, but it isn't available on XP. Anyway, even with all this I tested the performance and it came in at under 1 ms per call, so it's actually not too bad. Since the thread list only gets refreshed once per second it should be acceptable.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1470)

### DIFF
--- a/src/dbg/thread.cpp
+++ b/src/dbg/thread.cpp
@@ -125,12 +125,14 @@ void ThreadGetList(THREADLIST* List)
         List->list[index].ThreadCip = GetContextDataEx(threadHandle, UE_CIP);
         List->list[index].SuspendCount = ThreadGetSuspendCount(threadHandle);
         List->list[index].Priority = ThreadGetPriority(threadHandle);
-        List->list[index].WaitReason = ThreadGetWaitReason(threadHandle);
         List->list[index].LastError = ThreadGetLastErrorTEB(itr.second.ThreadLocalBase);
         GetThreadTimes(threadHandle, &List->list[index].CreationTime, &threadExitTime, &List->list[index].KernelTime, &List->list[index].UserTime);
         List->list[index].Cycles = ThreadQueryCycleTime(threadHandle);
         index++;
     }
+
+    // Get the wait reason for every thread in the list
+    ThreadGetWaitReasons(List);
 }
 
 void ThreadGetList(std::vector<THREADINFO> & list)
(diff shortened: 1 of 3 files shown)


mrexodia:
Isn't that pid from thread implemented as a queryinformationthread call?
I'll check the code out later...

Mattiwatti:
Yeah, `GetProcessIdOfThread` == `NtQueryInformationThread(ThreadInformation)`. I can add it if you want, I just figured there should be some way to get the PID that I'm not seeing, since it's pretty relevant information for a debugger.

mrexodia:
Hm, I'm rather concerned for performance. If this takes 1ms per call it would mean 1ms overhead per thread on each call to `DebugUpdateGui` (eg on every step/breakpoint), which will definitely add up when spamming stepping. The information about the WaitReason is definitely not critical so perhaps its indeed better to cache it and call it periodically...

Mattiwatti:
I said under 1 ms, not 1 ms :)

I just profiled it with a high accuracy timer and it came in at 0.159 ms per call (this is on a Core i7 6700K). I expected the constant memory allocations to be a contributing factor, but 99.46% (!) of the time was used by NtQuerySystemInformation itself. It's a really expensive syscall.

One thing that could be done is change the signature of `ThreadGetWaitReason` to `ThreadGetWaitReason(std::unordered_map<DWORD, THREADINFO>&)` and call it once in `ThreadGetList` with the thread list so it could do all the thread ID comparisons in one go. If it could also have the PID from somewhere that would be nice, but I wouldn't do an `NtQueryInformationThread` for it if it can be helped. I wouldn't bet money on another added syscall being faster than just doing a few hundred integer comparisons.

mrexodia:
Yeah that idea with the map sounds good, makes it pretty much constant and
small too.

mrexodia:
Thanks!

mrexodia:
Hm I didn't notice your commit was so recent, sorry if you still wanted to push something...

Mattiwatti:
No that's fine, it's (hopefully) working as intended now. I did change the `unordered_map&` to a `THREADLIST*`, since that contains both the thread count and the actual struct to put the wait reasons in. This one isn't gonna end up on my C++ hall of fame, what with the triple nested loops (one of which is a `while (true)`) but I blame MS for that for the terrible design of `NtQuerySystemInformation`.

Btw, there is a new `ThreadSystemThreadInformation` class for `NtQueryInformationThread` now which should return 1 thread struct instead of `N` process structs * `M` thread structs for the entire system. But it requires Windows 10, so I'll pass on that one.

PR #1478, opened 2017-02-25
Use symbolic names in the stack view

Mattiwatti:
The comments in the stack view aren't that useful at the moment IMO. Basically they all say "return to x from y", where x and y are raw pointer values that are already shown in the To and From columns. This is an attempt to make the comments more useful/readable by using symbolic names where possible, in the order:
- `module.ext!symbol+offset` (if symbols are available)
- `module.ext+offset` (if only the module name is available)
- raw pointer value (final fallback for virtual modules with no info and the like)

Screenshot of what it looks like with PDBs available for all modules:
Callstack

Note that this only shows the symbolic name for Rip/Eip for a given frame. But in the case that To doesn't match the previous frame (is this common?), the value can still be read from the column on the left as before. I also left the suspected call stack unchanged for now, mostly because it tends to have missing data for some frames.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1478)

### DIFF
--- a/src/dbg/stackinfo.cpp
+++ b/src/dbg/stackinfo.cpp
@@ -167,11 +167,48 @@ void StackEntryFromFrame(CALLSTACKENTRY* Entry, duint Address, duint From, duint
     Entry->from = From;
     Entry->to = To;
 
-    char returnToAddr[MAX_COMMENT_SIZE] = "";
-    getSymAddrName(To, returnToAddr);
-    char returnFromAddr[MAX_COMMENT_SIZE] = "";
-    getSymAddrName(From, returnFromAddr);
-    sprintf_s(Entry->comment, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "return to %s from %s")), returnToAddr, returnFromAddr);
+    if(From == 0)
+    {
+        memcpy(Entry->comment, "???", 4);
+        return;
+    }
+
+    auto module = ModInfoFromAddr(From);
+    if(!module || !module->name || module->name[0] == '\0')
+    {
+        // No module name available. Just print the address
+        sprintf_s(Entry->comment, MAX_COMMENT_SIZE, "%p", (void*)From);
+        return;
+    }
+
+    // Try to get a symbol name
+    DWORD64 displacement = 0;
+    Memory<PSYMBOL_INFO> pSymbol = Memory<PSYMBOL_INFO>(sizeof(SYMBOL_INFO) + MAX_SYM_NAME * sizeof(char));
+    pSymbol()->SizeOfStruct = sizeof(SYMBOL_INFO);
+    pSymbol()->MaxNameLen = MAX_COMMENT_SIZE - (ULONG)strlen(module->name) - 24;
+    if(SafeSymFromAddr(fdProcessInfo->hProcess, (DWORD64)From, &displacement, pSymbol()))
+    {
+        SafeUnDecorateSymbolName(pSymbol()->Name, pSymbol()->Name, pSymbol()->MaxNameLen, UNDNAME_NAME_ONLY);
+    }
+
+    if(pSymbol()->NameLen == 0)
+    {
+        // No symbol name available. Use module.ext+offset
+        sprintf_s(Entry->comment, MAX_COMMENT_SIZE, "%s%s+0x%llu", module->name, module->extension, From - module->base);
+    }
+    else
+    {
+        if(displacement == 0)
+        {
+            // module.ext!symbol
+            sprintf_s(Entry->comment, MAX_COMMENT_SIZE, "%s%s!%s", module->name, module->extension, pSymbol()->Name);
+        }
+        else
+        {
+            // module.ext!symbol+offset
+            sprintf_s(Entry->comment, MAX_COMMENT_SIZE, "%s%s!%s+0x%llu", module->name, module->extension, pSymbol()->Name, displacement);
+        }
+    }
 }
 
 #define MAX_CALLSTACK_CACHE 20


mrexodia:
I think it's definitely a nice change, however it breaks the style of every other symbolic name in x64dbg `ntdll.NtQueryInformationProcess+4C` for example. It also doesn't respect the symbol undecoration setting or the symbol filter. I would suggest changing the `stackinfo.cpp:getSymAddrName` function instead (with some of the code from _exports.cpp:getLabel to handle the setting and the filter). That way this is also supported in the normal stack...

It appears to be already supported, but only for user-defined or analyzed. I will push support for symbol displacement (apparently forgot when implementing it originally). As for the call stack, dbghelp is making up a lot of garbage (both for the callstack and for the symbol displacement) but there isn't much to be done about that I guess...

Mattiwatti:
I did consider changing `getSymAddrName`, but I saw there was an open issue re: performance problems with dbghelp's `SymFromAddr` which would probably be called a lot this way, no? Though I have no issue with it otherwise of course. Same for the symbolic name styles.

Re: symbol undecoration: I had it set to the default used throughout the rest of x64dbg, but it was resulting in some ridiculously long comments that were giving me a horizontal scrollbar and/or getting truncated because of the maximum buffer length. Hence why I changed it to the shortest possible.

The stack in the screenshot above was generated with `RtlVirtualUnwind` by the way, not dbghelp as I'm not a fan of it either. Unfortunately `RtlVirtualUnwind` is only available on x64 due to how SEH information is stored...

mrexodia:
The performance wouldn't change, because SymFromAddr is already called in getSymAddrName. As for the undecoration, perhaps it's a good idea to change the default?

Mattiwatti:
Oh yes, that's right. I was getting it mixed up with `ModNameFromAddr` which is called all the time in `DebugUpdateGui`, but that function doesn't involve dbghelp unless source debugging is enabled.

I'd definitely be in favour of changing the default for C++ unmangling to something less verbose. But it's hard to say what a good default is, because what's useful information depends on the context. I'd say a simple **ClassName::FuncName** would actually be sufficient in most cases though.
- Calling convention: this can occasionally be useful depending on what you're debugging. But normally it's simple enough to tell from the call site what the calling convention is. I also wouldn't care much for seeing it in the call stack view. It's more useful in the disassembly.
- `private`,`protected`,`public`,`virtual`,`static`, ... all of these can be done away with if you ask me. They are high level language constructs that are relevant when programming in C++ or whatever language, but not at the instruction level.
- Parameter types: can be useful, but if there are a lot of them it becomes unreadable quickly. (That's why I like xAnalyzer, it neatly separates them by line). Not sure about this one.

FWIW, WinDbg agrees with my naming scheme ;) Sample `undname` output:
`NTSTATUS __stdcall LRPC_CASSOCIATION::AlpcSendWaitReceivePort(ulong,_PORT_MESSAGE *,_ALPC_MESSAGE_ATTRIBUTES *,_PORT_MESSAGE *,unsigned __int64 *,_ALPC_MESSAGE_ATTRIBUTES *,_LARGE_INTEGER *)`

Becomes:
```
kd> u fffff880`0110d0b0
    msrpc!LRPC_CASSOCIATION::AlpcSendWaitReceivePort
```
Re: stack walks and dbghelp's crappiness: I was hoping Clang/LLVM would be able to do this. I know their MSVC compatibility has gotten very good recently, and that includes support for exception handling which must mean they can unwind stack frames properly. But I haven't been able to find anything on using the LLVM APIs to do this programmatically. So either it's well hidden or it's only meant for runtime exception handling and not debugging/call stack analysis. Shame, because I used LLVM for some simple analysis and name demangling once and it was really easy to install and use given how much it can do (like static analysis - IMO it would be pretty awesome to be able to use Clang/LLVM's static analysis from the debugger).

mrexodia:
Hm, so how would we go about parsing the names like windbg does? I definitely think the extended C++ definitions are too much. Perhaps https://bitbucket.org/Nukem9/demangle could help?

Mattiwatti:
Yeah, that looks like it should work. Another alternative is what LLVM-Symbolize does in [LLVMSymbolizer::DemangleName()](https://github.com/llvm-mirror/llvm/blob/2ae7de27f7d9276e7bada445ea7576bbc4c83ae6/lib/DebugInfo/Symbolize/Symbolize.cpp#L478), which looks like a pretty elegant solution to me. Basically it really comes down to
```C++
if (name[0] == '?') // Mangled C++ name
  return UnDecorateSymbolName(name, VARIOUS_FLAGS); // <-- dbghelp
if (isx86Module())
  return demanglePE32ExternCFunc(name); // <-- defined in the same .cpp
return name;
```
The only real difference I see at a first glance is `__imp` stuff, I'm not sure if LLVM deals with that.

It probably wouldn't be a bad idea to compile a list of garbage names that can come out of various linkers (not only MSVC's, though that one is probably most important), see what the `Demangle()` output is for each and adjust accordingly.

As for the pull request, I believe (sans the demangling) df0c75ac03d1dcb9721e2e40116ce30f6d923bb7 already did most of the work for me. My only suggestion would be to change the "return to x from y" to simply "y". That would make it a lot more readable IMO since atm each name is listed twice. (This is only for the call stack view tab, the CPU call stack should probably keep it to better distinguish stack frames from values that happen to be a function pointer.)

mrexodia:
Hm those flags are rather interesting. It should reduce the footprint of demangled C++ functions in the disassembly/call stack. For the symbol enumeration we can still fully demangle them for completeness.

As for `__imp`, there is an option that filters out certain names: https://github.com/x64dbg/x64dbg/blob/development/src/dbg/_exports.cpp#L117

And do you know if the isx86module() function makes any sense to use in x64dbg? I don't quite get what that check is supposed to do...

Mattiwatti:
`isx86module()` means 'PE is `IMAGE_FILE_MACHINE_I386`'. `extern C` functions don't get the extra decorations (underscores etc) on x64 AFAIK.

Mattiwatti:
Oh yeah, I took a look at what LLVM does to print stacktraces, but that didn't give me much hope. They have two methods (which one is used by default I don't know):
- `StackWalk64` using dbghelp, which is exactly what x64dbg does
- A wrapper function for `RtlCaptureStackBackTrace` (which is a wrapper for `RtlWalkFrameChain` which in turn is a wrapper for `RtlVirtualUnwind` at least on x64, I don't know how it is handled on x86.)

The problem with the `RtlCaptureStackBackTrace` method is that it needs to be done in the context of the thread for which you are requesting the stack. This makes it impossible to do on another thread from user mode. For the stack in the screenshot I used a kernel driver that injects an APC in the target thread, walks its frame chain and then copies back the results (hence why there is a kernel stack portion...)

Do you have an example for me where dbghelp produces a bogus stacktrace? I can use `RtlWalkFrameChain`/`RtlVirtualUnwind` to see if it makes a difference. It might be that dbghelp also uses these calls under the hood. In that case it wouldn't really matter anyway.

mrexodia:
I recently saw an example of x32dbg itself, where the Main thread would have SetCurrentDirectory in the callstack after x32dbg is fully loaded. This function is called in the plugin initialization but obviously it returned way before that...

Mattiwatti:
Only on 32 bit, not 64? Best would be a reproducable bogus stacktrace on x64, since my hack driver uses `RtlVirtualUnwind` which is x64 only. I can call `RtlCaptureStackBackTrace` but that only gives you a Rip/Eip chain, no stack pointers.

If the bogus stacks only manifest on 32 bit, then that would most likely be explained by the differences in SEH, for which x64 has static unwind tables baked into the PE. If it also happens on x64 it's worth checking out if `RtlVirtualUnwind` does a better job there.

mrexodia:
I just tried on x64dbg, but there everything appears to be working fine at the moment... It makes sense that it would be much better if dbghelp actually uses those unwind tables..

mrexodia:
I'm currently looking into some stuff related to `SymFromAddr` because the documentation is super weird. https://msdn.microsoft.com/en-us/library/windows/desktop/ms680686(v=vs.85).aspx gives all possible flags, but I get `0x400000` as a flag, which is not listed there...

However I manually did some stuff and it should be as you intended. The only 'problem' is that `SymAddrFromName` will not work with the undecorated names, but they didn't really before either so that isn't too much of a problem (especially `demanglePE32ExternCFunc` on C symbols could be seen as a problem but I'm not sure if anything can/should be done about it).

Mattiwatti:
`0x00400000` = `SYMFLAG_PUBLIC_CODE`.

These ones seem to be missing from MSDN:
```
#define SYMFLAG_NULL                0x00080000
#define SYMFLAG_FUNC_NO_RETURN      0x00100000
#define SYMFLAG_SYNTHETIC_ZEROBASE  0x00200000
#define SYMFLAG_PUBLIC_CODE         0x00400000
```

I'm getting this from the Windows SDK dbghelp.h.

PR #1487, opened 2017-03-10
fix 

torusrxxx:
When starting debuggee, it says "thread switched from 0"(fixed). When restarting the debuggee, it says "thread switched from ???(previous thread id)"(not fixed).

This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1487)


mrexodia:
Thanks!

PR #1490, opened 2017-03-11
Fix OS compatibility in .exe manifest

Mattiwatti:
As we know VS has these lovely manifest files to declare stuff like needing comctl32.dll v6.0 or being compatible with Windows (7, 8, 10). x64dbg declares the latter in app.manifest, but this file is actually completely ignored during compilation. The other file, manifest.xml, does go into the resource section (reason: it is declared in the .rc file which takes precedence). The result is that x64dbg actually runs in compatibility mode on any OS later than Windows 7 (you can check this in resmon).

This commit simply deletes app.manifest and moves the OS compatibility section to manifest.xml instead to fix the above.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1490)


mrexodia:
Cool, thanks!

PR #1498, opened 2017-03-15
Add support for Clang compilation

Mattiwatti:
This PR makes x64dbg compile and run using Clang/C2 from VS2017. The main goal of this is to add Clang's impressive diagnostics and warnings for additional compile time error checking. (There are a **lot** of warnings currently, but I've kept the changes to the minimum required to get the project to compile.)

Some notes:
- The launcher (x96dbg.exe) still requires MSVC to compile due to the ATL headers being... well, special. I haven't managed to get the project to compile without modifying the ATL headers, which is probably going a bit too far.
- The GUI DLL also requires MSVC, but this is more due to Qt's build system's lack of support for Clang/C2. It could probably be made to work with enough violence (given that Qt does support Clang/LLVM on Windows), but I'm not really interested in this.
- Clang/C2 from VS2015 doesn't work, due to some MS header deficiencies that seem to have been fixed in VS2017.
- Clang with LLVM also works (as in: compiles). I tested this with the Clang 5.0 alpha from SVN. I haven't even tried running the result though as the MSVC and LLVM ABIs are binary incompatible. Getting this to run would require compiling all dependencies including Qt with Clang/LLVM.

Most of the changes are non-functional, such as added type casts and uses of 'this' in virtual template methods. There are some exceptions though:
- `getLastExceptionInfo()` now returns `EXCEPTION_DEBUG_INFO` by reference rather than by value. This is to get the following to work:
```C++
getLastExceptionInfo().ExceptionRecord.ExceptionCode = 0;
```
To be honest I'm not sure how this even compiles in MSVC. It's possible that this only 'works' currently by modifying a local copy, i.e. effectively a no-op.
- `for(auto & i = windowsList.cbegin(); i != windowsList.cend(); i++)`: I've changed this to simply `auto` (i.e. non-reference). I don't feel bad for this as casting an rvalue to a reference this way is a nonstandard MS construct.
- The `LOADEXPORT` macro has been changed to print (in case of error) export names as ANSI strings (with uppercase `%S` specifier) rather than unicode. This is because I couldn't manage to get the Clang preprocessor to create an `L""` type string no matter what. This shouldn't be a problem though as the export names are all already required to be ANSI as they are passed to `GetProcAddress`.

There are also two changes to headers which are technically nonfunctional, but not especially pretty:
- `memory.h` is now laid out in the order function declarations -> `#include "addrinfo.h"` -> `extern`/`SimplePage` declarations. This is because including `addrinfo.h` causes `serializablemap.h` to be included, which in turn depends on some function declarations from `memory.h`. This could be fixed by moving `extern std::map<Range, MEMPAGE, RangeCompare> memoryPages` to `addrinfo.h` and including that separately where this variable is needed.
- `bridgemain.h` similarly has an \#include moved downwards (`bridgelist.h`). This is because `bridgelist.h` depends on the preceding function declarations.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1498)

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -38,7 +38,7 @@ static bool bDisableGUIUpdate;
     *((FARPROC*)&name)=GetProcAddress(hInst, #name); \
     if(!name) \
     { \
-        wsprintfW(szError, L"Export %s:%s could not be found!", szLib, L#name); \
+        wsprintfW(szError, L"Export %s:%S could not be found!", szLib, #name); \
         return szError; \
     }
 
@@ -1449,7 +1449,7 @@ BRIDGE_IMPEXP void GuiCloseQWidgetTab(void* qWidget)
 
 BRIDGE_IMPEXP void GuiExecuteOnGuiThread(GUICALLBACK cbGuiThread)
 {
-    _gui_sendmessage(GUI_EXECUTE_ON_GUI_THREAD, cbGuiThread, nullptr);
+    _gui_sendmessage(GUI_EXECUTE_ON_GUI_THREAD, (void*)cbGuiThread, nullptr);
 }
 
 BRIDGE_IMPEXP void GuiUpdateTimeWastedCounter()
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -7,9 +7,6 @@
 #include <stdbool.h>
 #endif
 
-//list structure (and C++ wrapper)
-#include "bridgelist.h"
-
 //default structure alignments forced
 #ifdef _WIN64
 #pragma pack(push, 16)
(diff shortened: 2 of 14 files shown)


mrexodia:
Definitely very interesting! Why is the this pointer needed in those
template classes though?

Mattiwatti:
Clang is more picky about wanting fully qualified types (though I'm not sure why this affects templates in particular - maybe someone with knowledge of the C++ standard could chime in). Initially I was working around this as follows:
```C++
template class<U, V>
struct Derived : Base<V>
{
  typedef Base<V> Base;
}
```
and then calling each method in the inheritance chain with `Base::Func()`. (The typedef is optional, but usually makes it more readable.) This got especially bad in `serializablemap.h` though where there are many small structs in a deep inheritance chain, so you end up with something like `Base::Base::Base::Func()`. Looking at that I thought I must be doing it wrong, so I tried simply calling it as a member function which worked. I assume MSVC applies the `this->` under the hood. I don't think any pointers are actually dereferenced for either compiler though (at least not here), since the inheritance chain can be resolved at compile time.

mrexodia:
Alright, thanks a lot! I hope the clang warning can improve the overall quality (you did already catch a few errors with these changes).

PR #1503, opened 2017-03-18
Updated telegram & freenode badge

ioncodes:
updated badges for telegram and freenode to use shields.io. this will ensure that the resolution is correct on all devices.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1503)


mrexodia:
Thanks!

Issue #1502, opened 2017-03-18
search for .pdb locally first, use symbol server after


sthalik:
> Then, make sure the following points are included in your issue:

The operating system is Windows Server 2012 R2. I'm using nightly versions.

> Brief

x64dbg makes lots of needless symbol server queries.

> Issue text

Whether using 'download this .pdb file' or 'download all .pdb files' in "Modules" right-click menu, x64dbg makes lots of needless queries to the symbol server. The queries fail as expected and only then the software searches for the .pdb file, which exists and is valid.

I'd rather have the order opposite -- validate the .pdb file with regard to the module, fall back to the symbol server if necessary. Either that, or cache negative results.

In particular, I'm using the software to verify optimized code quality, not forensics or similar purpose. The symbol server queries are generated in the UI thread, adding to their needlessness.

mrexodia:
If you know how to do it with dbghelp I'll be glad to include it. If you don't want queries to the remote server you can try to remove the symbol servers.

sthalik:
Unfortunately I don't. Would it possibly help to clear the symbol server list in the environment variable to check .pdb only?

sthalik:
Barring that, how about grabbing .pdb in a subprocess with no symbol server list, then launching it again with an actual server list?

Sorry for two responses.

I'm willing to implement that, but please point me to the area of your code that deals with retrieving .pdb's via dbghelp.dll api.

mrexodia:
Currently the code (https://github.com/x64dbg/x64dbg/blob/development/src/dbg/symbolinfo.cpp#L173) calls [SymSetSearchPath](https://msdn.microsoft.com/en-us/library/windows/desktop/ms681368(v=vs.85).aspx) with the server you specify in the settings dialog.

`SRV*C:\x64dbg\bin\x32\symbols*http://whateverserver` but I guess you could replace the server with another local path to search first to prioritize that. (like `C:\mysymbols*http://fallbackserver`)

You can find more information on symbol search paths here: https://msdn.microsoft.com/en-us/library/windows/desktop/ee416588(v=vs.85).aspx

This code has nothing related to downloading symbols it only uses a trick to be able to get symbols from servers only when requested (because you don't want to be downloading stuff during normal operation). As for a subprocess I think that's overkill. As far as I know downloading symbols is not done on the UI thread and it shouldn't block UI operations as of very recently...

sthalik:
@mrexodia see #1504

sthalik:
[...]

PR #1504, opened 2017-03-19
DBG: load symbols from local filesystem first (#1502)

sthalik:
Attempt to load symbols from .pdb path embedded in the PE executable prior to executing any network IO. Currently the order is the opposite. See #1502.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1504)


mrexodia:
Thanks!

PR #1518, opened 2017-04-03
Respect encoding type for tooltips in dump

torusrxxx:
x64dbg-cpudump-tooltip-string

This pull request enables the user to explicitly specify the data type for a piece of data and the tooltip in the dump will respect it. Currently, the user can only follow the data in disassembly in order to specify the data type even if it is not a piece of code. We can enable the user to specify encoding type in the dump in future.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1518)


torusrxxx:
This code may recurse indefinitely if the pointers have a cyclic reference
pattern.

PR #1526, opened 2017-04-08
Fixed

torusrxxx:
1. (CRITICAL BUG) x64dbg crash when mouse hovers over a pointer pointing to itself in the dump.
2. Fixed EBP based calling convention
3. Fixed some small issues with mouse cursor.
4. Added some tooltips about lock modes in arguments widget.

EBP calling convention details:
Memory addresses with arguments will not change whether current instruction is CALL or not. Only change argument address when it executes "mov ebp,esp". It may not work when the debuggee use EBP as a variable, and cannot use "Calls" update strategy.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1526)


mrexodia:
Thanks!

PR #1529, opened 2017-04-10
Improved user experience on single-core CPU

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1529)


mrexodia:
I just tried this on xp 64 and the user experience is (still) awful. It almost never registers the new EIP correctly when stepping slowly.

PR #1531, opened 2017-04-11
Add a comment for the newly allocated memory. This comment will show …

torusrxxx:
…in the memory map.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1531)


mrexodia:
Good idea!

torusrxxx:
Oh, I made a mistake here. GuiTranslateText() must be used here, otherwise
it will not be localized.

PR #1555, opened 2017-04-20
Local variables view

torusrxxx:
x64dbg-local-variables-view
A more appropriate icon is needed.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1555)


mrfearless:
Maybe some icon like:  or

PR #1556, opened 2017-04-20
added string prefix

### LINKED ISSUE #1027: Support "dword ptr ds:[eax]" syntax in expressions
Almost nobody is familiar with the "4:[eax]" syntax.

kkthx:
fixes #1027

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1556)

### DIFF
--- a/src/dbg/value.cpp
+++ b/src/dbg/value.cpp
@@ -1602,7 +1602,10 @@ bool valfromstring_noexpr(const char* string, duint* value, bool silent, bool ba
 
     if(string[0] == '['
             || (isdigitduint(string[0]) && string[1] == ':' && string[2] == '[')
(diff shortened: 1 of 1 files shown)


src/dbg/value.cpp
```diff
@@ -1627,6 +1630,58 @@ bool valfromstring_noexpr(const char* string, duint* value, bool silent, bool ba
             if(new_size < read_size)
                 read_size = new_size;
         }
+        else if (string[0] == 'b'
```
mrexodia: Shouldn't this be called after checking for `fs` and such? Because this could read from unknown memory right?

mrexodia:
Thanks!

PR #1570, opened 2017-04-29
Warn the user about mismatch of database and executable

torusrxxx:
Prepare for run trace subsystem. This approach reads the hash from the database and compare it to the debuggee when the debuggee is loaded. The reverse is not possible because when debugging a DLL, the database loads before the debuggee loads.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1570)

### DIFF
--- a/src/dbg/database.cpp
+++ b/src/dbg/database.cpp
@@ -32,6 +32,11 @@
 char dbbasepath[deflen];
 
 /**
+\brief The hash of the debuggee stored in the database
+*/
+duint dbhash = 0;
+
+/**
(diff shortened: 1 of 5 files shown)


mrexodia:
I don't think this is a good solution. The hash should be over the file contents not over the mapped module. With file contents we can compute the hash also before the module is loaded...

mrexodia:
Oh wait, is this file directly mapped or mapped like it would be after loading?

torusrxxx:
You cannot map memory of another process except via a shared section object, which is not possible in the context of x64dbg. So this memory can only be from the disk.

PR #1580, opened 2017-05-06
highlight modified bytes

torusrxxx:
Highlights modified bytes with shadow buffer.

Known issue: The shadow buffer needs to be updated just before the debuggee resumes. Current architecture does not include a semaphore to achieve reliable locking so sometimes the shadow buffer might be updated **after** the debuggee resumes, and contains bytes that are modified. This circumstance might not be very visible if you are looking only at dump1.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1580)

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -1633,6 +1633,11 @@ BRIDGE_IMPEXP void GuiFlushLog()
     _gui_sendmessage(GUI_FLUSH_LOG, nullptr, nullptr);
 }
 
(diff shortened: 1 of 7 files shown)


mrexodia:
I'm currently quite busy but I will try to check it today.

mrexodia:
I did review the code and it looks good, however I have to do more extensive testing to see how this works performance-wise. There is also the reservation of memory usage, so perhaps we should disable this per default until more testing has been done?

mrexodia:
Tracking in https://github.com/x64dbg/x64dbg/issues/3032

PR #1584, opened 2017-05-09
Don't lose data when clicking side bar

torusrxxx:
I don't wanna lose data for my conditional breakpoints when clicking on the side bar again. The user now can only use left mouse button to toggle breakpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1584)


mrexodia:
Thanks!

PR #1586, opened 2017-05-11
remembers dialog position

torusrxxx:
It can remember the window position of dialogs with a simple to use utility function. This feature should be promoted to every dialog. But it is very late now. So I will push it now.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1586)


ThunderCls:
It could be added another couple fields to save window height/width as well

torusrxxx:
Note that some of these dialogs (example: assemble) are fixed-size.

PR #1602, opened 2017-05-18
Fixes handles search tables not updating properly

### LINKED ISSUE #1597: Enable/disable window (Handle tab)
There is a bug which causes x64dbg (applies to both x86 and x64 versions) UI not to update the context menu aand the window table to 'enabled' or 'disabled'.

Steps to reproduce:
1. load up x64dbg and notepad. (order doesn't matter)
2. attach x64dbg to notepad.
3. run the target. (notepad)
4. go to handle tab and refresh it.
5. disable target's (notepad) window.

This will call EnableWindow for the target and disable the window while still showing 'enabled' in the window table as well as the context menu showing 'disable window'.
In order to be able to enable the window you have to disable it again and then UI goes back to normal.

This bug has been reproduced on windows 7 x64 sp1 with today's snapshot (05/17/2017).

ThunderCls:
Closes #1597

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1602)

### DIFF
--- a/src/gui/Src/Gui/HandlesView.cpp
+++ b/src/gui/Src/Gui/HandlesView.cpp
@@ -166,6 +166,7 @@ void HandlesView::reloadData()
         mWindowsTable->mList->reloadData();
         mTcpConnectionsTable->mList->setRowCount(0);
         mTcpConnectionsTable->mList->reloadData();
+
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #1605, opened 2017-05-26
Defense against malicious "explorer.exe"

torusrxxx:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1605)


mrexodia:
Good catch!

PR #1537, opened 2017-04-13
pull

brock7:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1537)

PR #1619, opened 2017-06-14
pull

brock7:
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1619)


mrexodia:
@blaquee 

> <torusrxxx> in pull request #1537 he wants to use a custom and more powerful OpenProcess
> <mrexodia> he can just hook TitanOpenProcess

See https://botbot.me/freenode/x64dbg/2017-04-13/?msg=83944743&page=1

src/dbg/debugger_commands.cpp
```diff
@@ -1199,7 +1199,7 @@ CMDRESULT cbDebugAttach(int argc, char* argv[])
     if(!GetModuleFileNameExW(hProcess, 0, wszFileName, MAX_PATH))
     {
         dprintf("Could not get module filename %X!\n", pid);
-        return STATUS_ERROR;
+        // return STATUS_ERROR;
```
mrexodia: Could you give more details about this bug?

src/dbg/module.cpp
```diff
@@ -82,6 +82,9 @@ bool ModLoad(duint Base, duint Size, const char* FullPath)
             strcpy_s(file, FullPath);
     }
 
+	if (file[0] == '\0')
```
mrexodia: If this to not load modules that don't have a name?

src/dbg/stringutils.cpp
```diff
@@ -24,6 +24,9 @@ StringList StringUtils::Split(const String & s, char delim)
 
 String StringUtils::Escape(const String & s)
 {
+	// FIXME: bug
```
mrexodia: Could you explain the bug?

src/dbg/thread.cpp
```diff
@@ -241,8 +242,10 @@ int ThreadSuspendAll()
     int count = 0;
     for(auto & entry : threadList)
     {
-        if(SuspendThread(entry.second.Handle) != -1)
-            count++;
+		if (SuspendThread(entry.second.Handle) != -1)
+			count++;
+		else
+			dprintf("thread(%d) cannot suspend. handle: %x", entry.second.ThreadId, entry.second.Handle);
```
mrexodia: This has been added into x64dbg:

```c++
int ThreadSuspendAll()
{
    // SuspendThread does not modify any internal variables
    SHARED_ACQUIRE(LockThreads);

    int count = 0;
    for(auto & entry : threadList)
    {
        if(SuspendThread(entry.second.Handle) != -1)
            count++;
        else
            dprintf(QT_TRANSLATE_NOOP("DBG", "Failed to suspend thread 0x%X...\n"), entry.second.ThreadId);
    }

    return count;
}
```

x64dbg_note.txt
```diff
@@ -0,0 +1,59 @@
...
+		tokenPrivileges.Privileges[0].Attributes = 0;
+	AdjustTokenPrivileges(hToken, FALSE, &tokenPrivileges, sizeof(TOKEN_PRIVILEGES), NULL, NULL);
+	dwLastError = GetLastError();
+	CloseHandle(hToken);
+	return dwLastError;
+}
+
+HANDLE _OpenProcess(DWORD dwDesiredAccess, bool bInheritHandle, DWORD dwProcessId)
+{
+	HANDLE hCsrss = OpenProcess(PROCESS_DUP_HANDLE, FALSE, 428);
```
mrexodia: Could you explain this hardcoded 428 process id?

x64dbg_note.txt
```diff
@@ -0,0 +1,59 @@
...
+		dprintf("DuplicateHandle() failed\n");
+		return NULL;
+	}
+
+	dprintf("OpenProcess() ok. handle = %x\n", h2);
+	return h2;
+	
+}
+
+HANDLE MyOpenProcess(DWORD dwDesiredAccess, bool bInheritHandle, DWORD dwProcessId)
```
mrexodia: I think it should be possible to write a plugin that hooks `TitanOpenProcess` and calls this instead...

Mattiwatti:
Re: the idea of using CSRSS as a trampoline to duplicate kernel mode protected DRM process handles mentioned on IRC: that won't work. The same mechanism that is responsible for notifying kernel mode drivers of `NtOpenProcess` calls ([ObRegisterCallbacks](https://msdn.microsoft.com/en-us/library/windows/hardware/ff558692(v=vs.85).aspx)) can be used in the same way for `NtDuplicateObject` (= `DuplicateHandle`), or in fact any system call involving a handle that the driver wants to monitor. That is apart from the fact that opening a useable handle to CSRSS itself is no longer possible since Windows 8.1, even with debug privileges enabled like in this PR. I wrote [PPLKiller](https://github.com/Mattiwatti/PPLKiller) because of this. It should already be linked in the wiki, but it's not exactly a solution for everyone since it runs in kernel mode. Generally speaking, trying to circumvent any sort of kernel mode restriction from user mode is going to be a very big waste of time.

brock7:
pull

mrexodia:
There are no changes relevant to the current build in this pull request and there are also conflicts. I cannot merge this in any way.

PR #1620, opened 2017-06-15
Replace undocumented.h with ntdll header and libs

Mattiwatti:
This PR is similar to the one I made for [ScyllaHide](https://github.com/x64dbg/ScyllaHide) a few days ago. It replaces `undocumented.h` (which was just a bunch of random declarations, some of which were outdated or incorrect) with a more complete `ntdll.h`. I have been working on this header for a while to make it usable for fool-proof importing in those cases where ntdll is needed. Most of this revolves around ensuring backwards and forwards Windows compatibility. Therefore:
- The header does not contain any declarations that are not present in Windows 10 RS2 (the current).
- The header has `NTDDI_VERSION` guards for the required OS version where applicable, with XP x86 as the minimum required/supported version. This makes it possible to use these declarations if XP support is dropped at some point in the future, but prevents them from being used at this point in time.
- The import libraries contain only exports present in all OSes from XP through Windows 10 RS2, making it impossible to accidentally bring in a static import (through another header) that breaks compatibility with any supported OS version.

Altogether this should correct the biggest problem with importing from ntdll, namely that doing so is not guaranteed to work across Windows versions. Also, I should note that x64dbg already depends on static ntdll imports (not directly but through TitanEngine).

The second commit in the PR removes a few cases in DBG where ntdll functions were being imported through `GetProcAddress` and replaces them with static imports.

---
This change is [Reviewable](https://reviewable.io/reviews/x64dbg/x64dbg/1620)


src/dbg/commands/cmd-analysis.cpp
```diff
@@ -1,4 +1,5 @@
 #include "cmd-analysis.h"
+#include "ntdll/ntdll.h"
```
mrexodia: Is this required?

src/dbg/commands/cmd-debug-control.cpp
```diff
@@ -1,4 +1,5 @@
 #include "cmd-debug-control.h"
+#include "ntdll/ntdll.h"
```
mrexodia: Same here?

src/dbg/x64dbg_dbg.vcxproj
```diff
@@ -416,7 +416,7 @@
...
+      <AdditionalDependencies>ntdll\ntdll_x86.lib;keystone\keystone_x86.lib;$(ProjectDir)..\capstone_wrapper\bin\x32\capstone_wrapper.lib;$(ProjectDir)..\capstone_wrapper\capstone\capstone_x86.lib;yara\yara_x86.lib;lz4\lz4_x86.lib;jansson\jansson_x86.lib;DeviceNameResolver\DeviceNameResolver_x86.lib;XEDParse\XEDParse_x86.lib;$(SolutionDir)bin\x32\x32bridge.lib;dbghelp\dbghelp_x86.lib;TitanEngine\TitanEngine_x86.lib;ws2_32.lib;psapi.lib;kernel32.lib;user32.lib;gdi32.lib;winspool.lib;comdlg32.lib;advapi32.lib;shell32.lib;ole32.lib;oleaut32.lib;uuid.lib;odbc32.lib;odbccp32.lib;shlwapi.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
mrexodia: I think this should be replaced with `#pragma comment(lib, "ntdll_x86.lib")`

src/dbg/commands/cmd-analysis.cpp
Mattiwatti: No, there are two alternatives:
- `#include <ntstatus.h>`
  `#include "cmd-analysis.h`
  `// all other includes`
- `// all includes`
  `#undef WIN32_NO_STATUS`
  `#include <ntstatus.h>`

Same for the other .cpp. I think the second option (`#undef` followed by `#include`) is pretty ugly. I chose this over the first option because that way the file's own header would still be first. But any of the three works.

src/dbg/x64dbg_dbg.vcxproj
Mattiwatti: Ugh, really? :P I can do it of course, but personally I hate the pragma because it conflates compilation and linking and makes any file a potential source of linker dependencies (which you then have to find). Whereas with the .vcxproj version you only have to look in the project properties to see all required .lib files in one place.

Mattiwatti:
How about something like this?

```C++
#ifndef _NTDLL_H
#define _NTDLL_H

#pragma once

#ifndef DBG_WANT_UNDOCUMENTED_APIS
#pragma message("Careful! ntdll.h is included but DBG_WANT_UNDOCUMENTED_APIS is not defined.")
#endif
```

The only part I don't like is using `#pragma message` as I believe it's nonstandard, but MSVC doesn't support `#warning` which better states the intent. Or even `#error` could be used, I think it's standard C and it's not like the error would be hard to fix if you do need ntdll.h.

mrexodia:
Thanks a lot! I think that for now no `DBG_WANT_UNDOCUMENTED_APIS` is needed

PR #1623, opened 2017-06-16
Fixed issue with detached dumps

torusrxxx:
Follow in dump N now behaves correctly. Registers view can use the tab name of dump tabs, instead of "Dump n"


mrexodia:
Great 👍

PR #1632, opened 2017-06-29
don't restore window pos invisible; clean up msgbox

torusrxxx:
Changing screen resolution from high resolution to low resolution will not hide windows; Clean up message boxes a bit.


mrexodia:
Great!

PR #1644, opened 2017-07-09
fixed x96dbg

torusrxxx:


src/exe/resource.rc
```diff
@@ -13,11 +13,11 @@
 #undef APSTUDIO_READONLY_SYMBOLS
 
 /////////////////////////////////////////////////////////////////////////////
-// Neutral resources
+// ���ض����� resources
```
MaxXor: Why rename this?

mrexodia: No, idea... Could just be Visual Studio trolling

PR #1645, opened 2017-07-11
Don't Quit x64dbg

torusrxxx:


Georgeto:
Nice feature, maybe make it configurable.

mrexodia:
Okay should be implemented now. Thanks @torusrxxx

PR #1647, opened 2017-07-13
Fixed some unsafe code

blaquee:
Fixed some small unsafe codes with the aid of static code analyzers.

### DIFF
--- a/src/dbg/commandline.cpp
+++ b/src/dbg/commandline.cpp
@@ -103,7 +103,7 @@ void CmdLineCacheLoad(JSON Root)
     EXCLUSIVE_ACQUIRE(LockCmdLine);
 
     // Clear command line
-    memset(commandLine, 0, MAX_COMMAND_LINE_SIZE);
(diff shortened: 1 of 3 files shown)


src/dbg/commandline.cpp
```diff
@@ -103,7 +103,7 @@ void CmdLineCacheLoad(JSON Root)
     EXCLUSIVE_ACQUIRE(LockCmdLine);
 
     // Clear command line
-    memset(commandLine, 0, MAX_COMMAND_LINE_SIZE);
```
nooperation: Should this really be using MAX_SETTING_SIZE instead? It might be that the bug here is the allocation of commandLine with a size of MAX_SETTING_SIZE instead of MAX_COMMAND_LINE_SIZE

mrexodia: the `commandLine` is declared as `char commandLine[MAX_SETTING_SIZE];` so this is correct.

mrexodia:
Cheers 🍺

PR #1662, opened 2017-07-27
DBG: The log now indicates the name of the breakpoint set (closes #1613)

### LINKED ISSUE #1613: The log should indicate for what event a breakpoint was set for
Version: 31. May 2017  

In settings you can select events on which events the debugger should break. Unfortunately when a break does occur, for example a thread entry breakpoint, it may sometimes be unclear why the debugger stopped, as the debugger only displays "INT3 breakpoint "Thread xxxx" at [...]" and logs "Breakpoint at XXXXXXXX set!". It would be much clearer if the message at the bottom displayed something like "Breakpoint on the entry point for thread XXXX" and the log message could be modified to be "Set thread entry breakpoint for thread XXXX at XXXXXXXX".

pointerrrr:


mrexodia:
Thanks bro!

PR #1680, opened 2017-08-05
GUI: download symbols in CPU tab

### LINKED ISSUE #1636: Add shortcut or menu entry for symbol download in CPU
Add the ability to download symbols for the currently debugged module when in the CPU tab. Either through a shortcut or Context Menu entry

Georgeto:
Closes #1636

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -528,6 +528,15 @@ void CPUDisassembly::setupRightClickContextMenu()
     analysisMenu->addMenu(encodeTypeMenu);
 
     mMenuBuilder->addMenu(makeMenu(DIcon("analysis.png"), tr("Analysis")), analysisMenu);
+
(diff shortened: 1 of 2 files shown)


mrexodia:
I don't think downloading all symbols is relevant for a context menu in the CPU view. The option should only be to download symbols for the current module (if there is any)...

Georgeto:
That is how I implemented it at first, but the name of the option _Download Symbols for Current Module_ was too long for my taste. Shortening it to something like _Download Symbols_ makes its meaning ambiguous. Therefore I went for the submenu.

blaquee:
Hi, thanks for taking this issue on! I do believe Symbols for all Modules isn't necessary for the CPU Tab. 
If you want to remove the ambiguity, then might I offer the solution of building the menu item dynamically to include the name of the currently debugged module. So it would be: 
"Download Symbol for [ X ]" where X is filled in during Context Menu painting. @mrexodia this is possible under the new menuBuilder right?

mrexodia:
Thanks for your contribution @Georgeto! I merged it, but I will change the feature to not have a submenu.

PR #1703, opened 2017-08-30
graph for switch statements

torusrxxx:
x64dbg-switch-statements


torusrxxx:
Add history in graph.
Follow parent or children of a block.

mrexodia:
Thanks!

PR #1705, opened 2017-09-01
Switch reference tabs by forward/back mouse buttons

torusrxxx:


src/gui/Src/BasicView/ReferenceView.cpp
```diff
@@ -432,3 +433,11 @@ void ReferenceView::referenceExecCommand()
         }
     }
 }
+
+void ReferenceView::mouseReleaseEvent(QMouseEvent* event)
+{
+    if(event->button() == Qt::ForwardButton)
+        mParent->setCurrentIndex(std::min(mParent->currentIndex() + 1, mParent->count()));
```
mrexodia: This should be implemented as a signal instead of letting the `ReferenceView` know about the `QTabWidget` parent...

PR #1715, opened 2017-09-09
Tripleslash for scriptapi_misc functions

rotateLeft:


mrexodia:
The thing is indented with tabs instead of spaces, you can run `format.bat` (or `install.bat` before you commit) to automatically format the sources...

rotateLeft:
Hopefully formatted correctly now. Should I be including a prefix in the git commits? (DBG/GUI etc) ?

mrexodia:
Nah, it's fine perhaps next time you can include the prefix...

mrexodia:
Oh and thanks 👍

PR #1738, opened 2017-10-01
Browse dialog and goto dialog support auto-complete

torusrxxx:
x64dbg-goto-auto-complete


torusrxxx:
Due to performance issues, there is a maximum count of 20 auto-complete entries.

torusrxxx:
Very slow when loaded x64gui.pdb (36MB) and typed a symbol in it. The user has to type a space first to stop auto-complete.

justanotheranonymoususer:
Have you considered using some optimized data structure, such as a balanced search tree? (i.e. `std::map`)
Also, why not use the component from the "Symbols" tab? It has a nice red underline, supports regex and all the goodness.

torusrxxx:
There is no such data structure exist in x64dbg core. Also, we cannot store
gigabytes of symbol information in memory.

justanotheranonymoususer:
OK, then maybe have a leveldb database on disk? It should be sorted and optimized for such queries.

mrexodia:
You know you can use the symbols view right?

justanotheranonymoususer:
Yes, but it's not as quick as Ctrl+G, and I need to to select all of the modules and stuff. Olly's (2) goto dialog is nice...

mrexodia:
@torusrxxx perhaps we can cache only exported functions and just include those in the dialog like olly?

justanotheranonymoususer:
Olly includes everything, not only exported functions

torusrxxx:
There is one thing that this code can improve. Qt calls multiple times data() and then DbgSymAutoComplete() and then SymEnumSymbols(). I tried to cache the results for the same text but recieved segment faults, so the code is what it is.

mrexodia:
I think this feature can be merged in after the conflicts are resolved and a setting has been added to disable it...

mrexodia:
Very nice!

mrexodia:
Hm, I just checked and it bugs out pretty hard if you use the down arrow to navigate the suggestions:

torusrxxx:
Press arrow keys → textbox changes(selects a suggestion or a history record?) → candidates changes → qcompleter messed up

PR #1739, opened 2017-10-01
Fix standard conformance issues

danilaml:
Catching temporaries by a non-const reference is a VS extension and it'd fail to compile in a newer VS versions by default (tested in  VS2017).


mrexodia:
Thanks!

PR #1742, opened 2017-10-02
Revert "Fix standard conformance issues"

mrexodia:
Reverts x64dbg/x64dbg#1739


danilaml:
@mrexodia what is the rationale for the revert?
If it's because of https://forum.tuts4you.com/topic/39851-is-it-a-bug/?tab=comments#comment-192434
I think I know what is the issue: I forgot to change (or rather just overlooked):
https://github.com/x64dbg/x64dbg/blob/f0fd4b7acfac7e93346d27054c3abce56549fa38/src/gui/Src/Gui/TabWidget.cpp#L14

It should become `const QPoint &`.
Btw, this problem would be caught/non-existent if the project used new C++11 style for connects (which binds and checks for slot existence at compile time).

Issue #1304, opened 2016-11-28
select many items in bp, symbols, refs view


lstolcman:
1. select many items with shift
 this is currently doable in symbols view only:
![vcxrxxq 1](https://cloud.githubusercontent.com/assets/4583553/20684488/69bf41ce-b5b0-11e6-934c-2dc225ee05a1.png)

and in symbols modules:
![aa3tax4 1](https://cloud.githubusercontent.com/assets/4583553/20684575/dc3fe122-b5b0-11e6-8715-4c3e682b1c01.png)

should be also doable in breakpoints:
![y5e9x3p 1](https://cloud.githubusercontent.com/assets/4583553/20684600/f705c116-b5b0-11e6-9055-ee17296d5e46.png)

may be useful to implement it in references->strings and references->intermodular calls

also:
2. allow ctrl+a (select all) - select all items - not available but should be in these above cases

why? this is usable in case of adding/removing breakpoints groups

mrexodia:
I agree with you. We are open to pull requests and if you're interested take a look at it, should be easy to implement.

mrexodia:
@kkthx I think this issue can be closed. Can you verify?

kkthx:
Hello,
About first issue (select many items by shift) - it is fixed
About second - ctrl-a (select) all - working in "breakpoints" window, but not in "Symbols" window

justanotheranonymoususer:
Multiple items can be selected in symbols and breakpoints view, but not in references. It would be great if it could be added there too.

PR #1743, opened 2017-10-03
GUI: add multiselect hotkeys to StdTable

Atvaark:
This allows selecting multiple rows in e. g. the Breakpoints view
without having to use the mouse.

New hotkeys:
- Select all: (ctrl+a)
- Select first row (ctrl+home)
- Select last row (ctrl+end)
- Expand selection upwards (shift+up)
- Expand selection downwards (shift+down)

capture

Related issue: #1304

### DIFF
--- a/src/gui/Src/BasicView/StdTable.cpp
+++ b/src/gui/Src/BasicView/StdTable.cpp
@@ -7,7 +7,7 @@ StdTable::StdTable(QWidget* parent) : AbstractTableView(parent)
     memset(&data, 0, sizeof(SelectionData_t));
     mSelection = data;
 
-    mIsMultiSelctionAllowed = false;
+    mIsMultiSelectionAllowed = false;
     mIsColumnSortingAllowed = true;
 
     mData.clear();
@@ -46,7 +46,7 @@ void StdTable::mouseMoveEvent(QMouseEvent* event)
 
             if(wRowIndex < getRowCount())
             {
-                if(mIsMultiSelctionAllowed)
+                if(mIsMultiSelectionAllowed)
                     expandSelectionUpTo(wRowIndex);
                 else
                     setSingleSelection(wRowIndex);
@@ -84,7 +84,7 @@ void StdTable::mousePressEvent(QMouseEvent* event)
 
                 if(wRowIndex < getRowCount())
                 {
-                    if(mIsMultiSelctionAllowed && (event->modifiers() & Qt::ShiftModifier))
+                    if(mIsMultiSelectionAllowed && (event->modifiers() & Qt::ShiftModifier))
                         expandSelectionUpTo(wRowIndex);
                     else
                         setSingleSelection(wRowIndex);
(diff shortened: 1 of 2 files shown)


blaquee:
Does this break the context menu?

Atvaark:
It shouldn't. But the context menu will show actions that work with multiple selected rows (e. g. disable breakpoints) and with single rows (e. g. edit breakpoint).
The single row actions will target the last selected row.

src/gui/Src/BasicView/StdTable.cpp
```diff
@@ -195,6 +237,55 @@ void StdTable::expandSelectionUpTo(int to)
...
+void StdTable::expandUp()
+{
+    int wRowIndex = mSelection.firstSelectedIndex - 1;
+    if(wRowIndex >= 0)
+    {
+        if(wRowIndex < mSelection.fromIndex)
+        {
+            mSelection.fromIndex = wRowIndex;
+            mSelection.firstSelectedIndex = wRowIndex;
+            mSelection.toIndex =  mSelection.toIndex ;
```
nooperation: Are the self assignments in expandUp and expandDown intended? They may cause warnings with certain compiler flags or static analysis tools

Atvaark:
> Are the self assignments in expandUp and expandDown intended? They may cause warnings with certain compiler flags or static analysis tools

Thanks for noticing. I pushed a fix that shouldn't cause any compiler warnings.

mrexodia:
It looks good, however I think this will also allow multi-selection in the script view for example there is the `mIsMultiSelctionAllowed` variable to check this (with a nice typo lol)

Atvaark:
I pushed the fix for `mIsMultiSelectionAllowed`.

mrexodia:
Thanks!

justanotheranonymoususer:
Why ctrl+home and not just home? I'd also expect shift+home and shift+end, shift+pgup, shift+pgdn.

mrexodia:
Should be easy to adjust @justanotheranonymoususer, for me personally I have no preference, but things like sublime text use ctrl+home/end to go to the beginning of the file.

justanotheranonymoususer:
A file is different than a list. In a file, home/end go to the start/end of line. In the list there is no caret, so it's not relevant. Compare it to the system lists, you can even see it in explorer by switching to details view.

mrexodia:
Alright, just open a pull request with what you think makes sense 👍

justanotheranonymoususer:
Maybe one day, when I install visual studio...

mrexodia:
You can use the Github web editor and wait for appveyor to build your snapshot zip for the pull request as well 😉

justanotheranonymoususer:
I tried... don't really know how to do page up and page down

mrexodia:
Alright, no problem. Can I help with installing visual studio/qt or whatever @justanotheranonymoususer ?

mrexodia:
@Atvaark was this intended? https://sonarcloud.io/project/issues?id=x64dbg&open=AV8bl8YtVLHgQ8g1xlkd&resolved=false&severities=BLOCKER%2CCRITICAL%2CMAJOR%2CMINOR&sinceLeakPeriod=true&types=BUG

Atvaark:
That wasn't intended, since that warning was mentioned above. I thought I pushed the fix for that.
Would you mind removing the two lines or should I create another PR?

mrexodia:
If removing them is the fix I'll do it real quick tomorrow or whatever...

Atvaark:
Go ahead, Removing the two lines will fix these issues.

justanotheranonymoususer:
@Atvaark can this be used in the references view?

Atvaark:
I don't think so, since the reference view doesn't derive from StdTable, but from StdSearchListView.

PR #1744, opened 2017-10-03
Fix tab detaching

### LINKED ISSUE #1741: Problem with pull request
https://forum.tuts4you.com/topic/39851-is-it-a-bug/?do=findComment&comment=192434

danilaml:
fixes #1741

Btw, it wouldn't go unnoticed if the project used [new signal/slot syntax](https://wiki.qt.io/New_Signal_Slot_Syntax) (which is already quite old ;P).


mrexodia:
Thanks!

PR #1765, opened 2017-10-15
Added setting that controls MaxModuleSize.

### LINKED ISSUE #1396: module name on CALL destination
I notice that on every CALL _function_ instruction, the function is listed as _\<module name\>.sub_xxxxxxxx_ 

This gets really untidy on modules with long file names. Is there anyway to toggle it on/off so it just says CALL _sub_xxxxxxx_

I realise for the windows and external APIs that it may be helpful, but for calls within the same module it is not necessary

shamanas:
Closes #1396.  
Nothing interesting to see here.

### DIFF
--- a/src/gui/Src/Gui/SettingsDialog.ui
+++ b/src/gui/Src/Gui/SettingsDialog.ui
@@ -33,7 +33,7 @@
       <bool>true</bool>
      </property>
      <property name="currentIndex">
-      <number>3</number>
+      <number>0</number>
      </property>
      <widget class="QWidget" name="tabEvents">
       <attribute name="title">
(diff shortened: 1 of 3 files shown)


src/gui/Src/Gui/SettingsDialog.ui
```diff
@@ -33,7 +33,7 @@
       <bool>true</bool>
      </property>
      <property name="currentIndex">
-      <number>0</number>
+      <number>3</number>
```
mrexodia: Minor nitpick here :)

shamanas: Fixed, didn't even know this was a thing :P

mrexodia:
Thanks!

PR #1766, opened 2017-10-16
Remember navigation source in graph history

RaMMicHaeL:
If you follow a jump or a call, and click on '-', you don't go back to the jump or the call (as in the regular CPU view), but to a previous, non-relevant command in the graph. This commit tries to fix this.

I haven't compiled or checked this. Please verify, or I'll do it myself later.

__Edit:__ Tested AppVeyor's build, seems to be working.


mrexodia:
Thank you!

Issue #1757, opened 2017-10-13
support LastStatusValue in register view


mrexodia:
`_TEB->LastStatusValue` should be added to the registers view, with the relevant ntdll exception name.

kichristensen:
@mrexodia Just found this one, and it looks like it could be closed after the #1776 has been merged?

PR #1776, opened 2017-10-21
Show LastStatusValue in register view

Mattiwatti:
This is an attempt at implementing #1757 (show `TEB->LastStatusValue` as a pseudo-register in the same way as `LastError`).

The PEB/TEB and error database commits are more or less free-standing and can be merged as-is I think. I discovered that `offsetof(TEB, LastStatusValue)` was incorrect on x86 so that one is probably not a bad idea.

Questions re: the changes I made to dbg:
- I added members to both `REGDUMP` and `THREADALLINFO` in bridgemain.h (in the same places where there was already a last error member). As this header is part of the SDK, how many plugins does this kill? None of mine seem to be affected, but there are a lot of plugins...
- In `TraceRecordManager` I cut off `REGDUMPWORD::regword[]` (meaning it is the same size as before) because of the comment stating the spare space is reserved for AVX512. I'm fairly sure this is harmless since the last status is never used by Windows itself (it is purely informational), but I'd feel better about this if I knew what this field is actually for. At the moment I don't really have a clue.

The GUI part seems to mostly be working fine, except for this:
- The edit box for the last status seems to want a 64 bit value for some reason; e.g. if the current status is `0xC0000001 (STATUS_UNSUCCESSFUL)`, the edit box value will be `ffffffffc0000001`. This still works correctly (only 4 bytes are written), but it looks a bit odd to say the least. Strangely enough the LastError box doesn't do this, even though I copied the edit box code from there. Someone more knowledgeable about Qt than me can probably tell me what's up here.


src/bridge/bridgemain.h
```diff
@@ -757,6 +757,12 @@ typedef struct
 
 typedef struct
 {
+    NTSTATUS code;
```
mrexodia: If you use DWORD here just like `LASTERROR` it will not display the `ffffffff....`

src/bridge/bridgemain.h
```diff
@@ -765,6 +771,7 @@ typedef struct
     X87STATUSWORDFIELDS x87StatusWordFields;
     X87CONTROLWORDFIELDS x87ControlWordFields;
     LASTERROR lastError;
+    LASTSTATUS lastStatus;
```
mrexodia: The main issue with this is that any plugins using this structure will break so I'd rather not do that...

A solution for this is to let the plugin pass the buffer size so we can make sure no buffers are being overwritten but it's a bit too late for that now 😢

src/bridge/bridgemain.h
```diff
@@ -810,6 +817,7 @@ typedef struct
     THREADPRIORITY Priority;
     THREADWAITREASON WaitReason;
     DWORD LastError;
+    NTSTATUS LastStatus;
```
mrexodia: Same here, plugins will break with this change if you change this structure...

src/bridge/bridgemain.h
justanotheranonymoususer: You can introduce `GetFoobarEx` like WinAPI does, when `GetFoobar` will call `GetFoobarEx(buff, original_size)`

Mattiwatti: Yep, that's what I was afraid of. If only `LASTERROR` had 4 more bytes of padding before `name` :(

What do you suggest to do about this? It is of course possible to add a new bridge export only to retrieve the last status value, but I think we can all agree that would be pretty horrible :P Possibly worse: leave the public header unchanged but secretly give the register view the bigger struct anyway because it knows about the status value. Ergh.

@justanotheranonymoususer's approach sounds like the only reasonable long term solution to me. E.g. `BRIDGE_IMPEXP bool DbgGetRegDumpEx(REGDUMP* regdump, size_t bufferSize);` where `bufferSize` is `sizeof(REGDUMP)`, but this value would of course depend on the plugin SDK version used.

I can also see other reasons for doing this besides adding the last status value: if AVX512 support is added in the future, `REGISTERCONTEXT` will need to be expanded with the zmm registers, which also affects the size of `REGDUMP`.

src/bridge/bridgemain.h
Mattiwatti: Yep. I simply removed this field for now since it isn't used in the threads view (unlike LastError).

src/bridge/bridgemain.h
Mattiwatti: Indeed! I think I'll refrain from asking why this is... fixed regardless.

src/bridge/bridgemain.h
mrexodia: I think `DbgGetRegDumpEx` would be the better solution. Internally we can change `_dbg_getregdump` to take an additional `size_t bufferSize` like you suggested to keep the change as small as possible...

Mattiwatti: Hmm, so you would suggest keeping one `_dbg_getregdump` internally (with an added `bufferSize` parameter) that is used by both `DbgGetRegDump` and `DbgGetRegDumpEx`? Just making sure as it's probably easier if I submit that change as a separate PR.

mrexodia:
Yeah that was my idea :)

Mattiwatti:
OK, I've rebased this on top of the `DbgGetRegDumpEx` commit with some changes to use `REGDUMP_V2`, and it Works For Me (TM).

mrexodia:
Thanks! I'll slightly refactor the `DbgGetRegDumpEx` thing but apart from that it looks great!

PR #1779, opened 2017-10-24
Added winerror format function and show help message for last error i…

torusrxxx:
…n reg view

### DIFF
--- a/src/dbg/formatfunctions.cpp
+++ b/src/dbg/formatfunctions.cpp
@@ -2,6 +2,7 @@
 #include "threading.h"
 #include "value.h"
 #include "memory.h"
+#include "exception.h"
 
 std::unordered_map<String, FormatFunctions::Function> FormatFunctions::mFunctions;
 
@@ -12,18 +13,18 @@ void FormatFunctions::Init()
         duint size;
         if(argc < 2 || !valfromstring(argv[1], &size))
         {
-            strcpy_s(dest, destCount, "Invalid argument...");
+            strcpy_s(dest, destCount, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Invalid argument...")));
             return FORMAT_ERROR_MESSAGE;
         }
-        if(size > 1024 * 1024 * 10) //10mb max
+        if(size > 1024 * 1024 * 10) //10MB max
         {
-            strcpy_s(dest, destCount, "Too much data (10mb max)...");
+            strcpy_s(dest, destCount, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Too much data (10MB max)...")));
             return FORMAT_ERROR_MESSAGE;
         }
         std::vector<unsigned char> data(size);
         if(!MemRead(addr, data.data(), data.size()))
         {
-            strcpy_s(dest, destCount, "Failed to read memory...");
+            strcpy_s(dest, destCount, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Failed to read memory...")));
             return FORMAT_ERROR_MESSAGE;
         }
         auto result = StringUtils::ToHex(data.data(), data.size());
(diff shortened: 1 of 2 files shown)


torusrxxx:
Usage: `{winerror@5}` will display `ERROR_ACCESS_DENIED:(description in your language)`

mrexodia:
Very nice, we can do the same for `NTSTATUS` values after #1776 is merged!

torusrxxx:
I don't think `FormatMessage` supports `NTSTATUS`.

mrexodia:
It does, you have to pass the module handle of ntdll in the first(?)
parameter...

torusrxxx:
So can we add a new `{ntstatus@C0000005}` format function or extend `winerror` to allow one more argument which accepts a module name like `ntdll.dll`? I don't think many DLLs support this operation.

Mattiwatti:
AFAIK kernel32.dll and ntdll.dll are the only DLLs containing any useful message tables for use with `FormatMessageW`. The ntdll ones should be queryable in the exact same way as the kernel32 ones. So IMO `{winerror@xx}` + `{ntstatus@xx}` is the best choice since I don't think there is a point to passing module names other than those two.

mrexodia:
Manually cherry picked in ec0555dc0d162c12d4d264ea7ba361b4e984cfe4

However to make this work properly (error code 574 or 547, it contains `%08X` so it sees `%0` as the end of the string) we have to manually retrieve and parse the message tables on debugger startup.

mrexodia:
These appear to be a few libraries that contain the error codes:

- kernel32.dll
- kernelbase.dll
- ntdll.dll
- netmsg.dll
- mferror.dll
- shell32.dll
- didn't check but probably more

We can use `RtlFindMessage` on the relevant DLL to get the raw error message which solves the issue of `%0` being parsed as the end of the string. I am not entirely clear on how the MUI things work in this case though, is it handled automagically or do we have to do something separate for that?

Mattiwatti:
Not sure about the .mui strings. I **think** that's all done automatically, but I don't have any way to know for sure. Does anyone have a non-English (or at the very least not en-US) Windows installation to try this?

```C++
#include "ntdll/ntdll.h"

#define MESSAGE_RESOURCE_UNICODE 0x1

int main(int, wchar_t**)
{
  UNICODE_STRING NtdllName = RTL_CONSTANT_STRING(L"ntdll.dll");
  PVOID Ntdll;
  NTSTATUS Status = LdrGetDllHandle(nullptr, nullptr, &NtdllName, &Ntdll);
  if (!NT_SUCCESS(Status))
    return -1;

  const NTSTATUS ErrorStatus = STATUS_ACCESS_VIOLATION;
  PMESSAGE_RESOURCE_ENTRY Entry;
  Status = RtlFindMessage(Ntdll,
                        LDR_FORMAT_MESSAGE_FROM_SYSTEM_MESSAGE_TABLE,
                        MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US),
                        ErrorStatus,
                        &Entry);
  if (!NT_SUCCESS(Status))
    return -1;

  if ((Entry->Flags & MESSAGE_RESOURCE_UNICODE) != 0)
    printf("%ls\n", (const wchar_t*)Entry->Text);
  else
    printf("%s\n", (const char*)Entry->Text);

  return 0;
}
```
(obviously changing the arguments to `MAKELANGID` as necessary)

On my en-US installation this returns an entry with `MESSAGE_RESOURCE_UNICODE` set and the text
`The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.`

If I change the language to `MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_UK)` I get `STATUS_RESOURCE_LANG_NOT_FOUND`.

stonedreamforest:
**zh-cn:**
`The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.`

Mattiwatti:
Thanks. Is that with `MAKELANGID(LANG_CHINESE, SUBLANG_CHINESE_SIMPLIFIED)` (which I think is the correct one for ZH-cn, correct, me if I'm wrong) or the one above? If it is for zh-cn, then I don't know what's up there, I would expect it to return Chinese text or an error if the lang ID is incorrect, not English...

Mattiwatti:
Forgot to ask, since the text for `STATUS_ACCESS_VIOLATION` is a rather... famous... ;) When you get this error from Windows itself, is it in English or Chinese? My dad has a Dutch version of Windows installed and I have occasionally (rarely) seen English messages/blurbs in it.

torusrxxx:
MAKELANGID(LANG_CHINESE_SIMPLIFIED, SUBLANG_CHINESE_SIMPLIFIED)
0x%08lx 指令引用的 0x%08lx 内存。该内存不能为 %s。 

MAKELANGID(LANG_SPANISH, SUBLANG_SPANISH)
(none)

MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US)
The instruction at 0x%08lx referenced memory at 0x%08lx. The memory could not be %s.

MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT)
0x%08lx 指令引用的 0x%08lx 内存。该内存不能为 %s。

Mattiwatti:
It looks like `MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT)` already does the right thing automatically then. The only issue I see is if someone is using e.g. a Spanish version of Windows but has x64dbg explicitly set to en-US. That will affect the rest of the UI language but not the output of `RtlFindMessage`. Not sure if this is really a big deal though, I expect other programs will have the same issue with system messages.

It might be worth passing `MAKELANGID(LANG_ENGLISH, SUBLANG_ENGLISH_US)` if the UI has been set to en-US instead of the system default? I expect that will be by far the most common language for people who do override the system default, but that's just my gut feeling. Why doesn't x64dbg collect tons of 'anonymous' user data in the background like Windows 10 does? Then we'd know for sure. It's 2017 guys, jeez.

Issue #1775, opened 2017-10-21
Do not overwrite symbol names with autocomments


Mattiwatti:
Debugger version: any recent
OS/bitness: any

The register content autocomments overwrite symbol names. Example:
![comment](https://i.imgur.com/bHmXtk0.png)

The marked comment was initially the name of the function, `BaseSetLastNTError`.

I know you can disable this behaviour with the 'autocomments only on CIP' option, but I do appreciate the autocomments most of the time. Here though I would definitely prefer to see the function name (perhaps even in bold or black for emphasis) rather than the contents of `edi`.

Manual comments are not affected by this.

mrexodia:
Yeah the symbols are displayed in the address column, you can expand it. This behavior is only to help out people who didn’t notice that 😀

Mattiwatti:
D'oh! Now I feel stupid. I've always had the address column set to the exact minimum width required to contain an address.

I'd still prefer the autocomment not overwrite the symbol name (or otherwise be appended to it); with the address column expanded it feels like a lot of screen real estate is being wasted because most addresses do not have associated symbol names.

mrexodia:
Yeah I agree that an option could be done for this. Also what do you think
about displaying a tooltip when mousing over the address column? I’m not
shre how I feel about that...

Mattiwatti:
Hm, I dunno about tooltips... having them is definitely an improvement over no tooltips, but the address column isn't the first place I'd look for something like this. Of course, after you've seen the tooltip once you'll know where it is so maybe it's not such a big deal. Maybe it could be done in combination with the same mini-arrows that already exist for jump instructions (but pointing sideways)?

`761E79F2 >`

That way you still have the breadcrumb hint for more info but without the address bar taking up 2-3 times the normal width. It also prevents you from hovering over addresses needlessly because you can immediately see whether or not there will be a tooltip for it.

torusrxxx:
We can disable autocomments for this particular instruction “mov edi,
edi”.It appears in system modules frequently and doesn’t have any meaning.

Mattiwatti:
[...]

PR #1780, opened 2017-10-25
Don't add autocomments on "mov edi,edi"(#1775)

torusrxxx:


src/dbg/_exports.cpp
```diff
@@ -337,103 +337,104 @@ extern "C" DLL_EXPORT bool _dbg_addrinfoget(duint addr, SEGMENTREG segment, BRID
...
-
-            //Ignore register values when not on CIP and OnlyCipAutoComments is enabled: https://github.com/x64dbg/x64dbg/issues/1383
-            if(!getregs)
-            {
-                for(int i = 0; i < instr.argcount; i++)
-                    instr.arg[i].value = instr.arg[i].constant;
-            }
-
-            for(int i = 0; i < instr.argcount; i++)
+            if(!cp.IsNop())
```
mrexodia: The diff is extremely unclear, this is the only link you added right?

justanotheranonymoususer: Try this:
https://github.com/x64dbg/x64dbg/pull/1780/files?w=1

PR #1781, opened 2017-10-25
Add tokenizer support for pointer operands

athre0z:
```
<Baboon> hi
<Baboon> it seems that jmp far are not correctly decoded by x64
<Baboon> (10/10/2017 version, didn't test the latest)
[...]
<Baboon> I'll download the last version
<Baboon> last version, same bug : "\xEA\xEF\xB5\x0D\x77\x33\x00"
<Baboon> got an empty string instead of /b meta
```

This fixes the issue.


mrexodia:
Thank you!

PR #1784, opened 2017-10-30
Overriding a non global action hotkey now asks for confirmation.

### LINKED ISSUE #1256: It can specify overlapping shortcuts in the shortcuts dialog

shamanas:
Closes #1256.  

Specifically, the current behavior is:  
- Overriding global hotkeys is not allowed at all.
- Overriding a local shortcut with a global one is silent.
- Overriding a local shortcut by a local one asks for confirmation.


mrexodia:
Looks good, however you cannot update the submodule since you don't have push rights there...

mrexodia:
Thanks!

Issue #975, opened 2016-08-13
Copy cryptographic hash of selected data


torusrxxx:

mrexodia:
generally I have refrained from this to avoid horrible dependency issues, might be a plan to create an impossible to compile module that has an interface for all hash types.

mrfearless:
Could use the extra space beside the dump for something like the checksum option that hex workshop has for checksums
![](https://s20.postimg.org/n7om1xejh/dump.png)
![](https://s20.postimg.org/dbnj2a8rh/hwchecksum.png)

And could be an option to outsource that stuff to a plugin, which if loaded calls a function to display this 'extra' widget beside the dump for this purpose, adding in right click menus to generate checksums / crypto hashes for selection/entire file, which are then shown on the right in the new 'crypto' widget.

torusrxxx:
Closed with #1785. An (orphaned) project "FrozenStorm" is still left there.

PR #1785, opened 2017-10-31
Copy cryptographic hash

torusrxxx:
#975 turned out to be built-in feature of Qt 5.6. Also refactored DataCopyDialog slightly.

### DIFF
--- a/src/gui/Src/Gui/DataCopyDialog.cpp
+++ b/src/gui/Src/Gui/DataCopyDialog.cpp
@@ -1,6 +1,7 @@
 #include "DataCopyDialog.h"
 #include "ui_DataCopyDialog.h"
 #include "Bridge.h"
+#include <QCryptographicHash>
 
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/DataCopyDialog.cpp
```diff
@@ -264,7 +278,12 @@ void DataCopyDialog::printData(DataType type)
         for(int i = 0; i < numIPs; i++)
         {
             if(i)
-                data += ", ";
+            {
+                if((i % mTypes[mIndex].itemsPerLine) == 0)
```
nooperation: I don't really see it being an issue right now, but this and the other two instances of it have the potential  to divide by zero in the future if itemsPerLine is ever set to zero (or someone blindly copy-pastes this code elsewhere)

mrexodia: Good point!

mrexodia:
👍

PR #1789, opened 2017-11-01
MSVC 2017 compile fixes

balintf:
This PR fixes the following 2 compile errors that occur when building using the MSVC2017 toolchain in Qt Creator:

1. Trying to bind an rvalue to non-const reference:
`..\gui\Src\Disassembler\CsQBeaEngine.cpp(274): error C2440: 'initializing': cannot convert from 'QHash<ENCODETYPE,CsQBeaEngine::DataInstructionInfo>::iterator' to 'QHash<ENCODETYPE,CsQBeaEngine::DataInstructionInfo>::iterator &'
..\gui\Src\Disassembler\CsQBeaEngine.cpp(274): note: A non-const reference may only be bound to an lvalue`

2. Declaring a struct inside a lambda function:
`..\gui\Src\Gui\BreakpointsView.cpp(145): error C2327: 'BreakpointsView::reloadData::<lambda_14f74381a4291d4224dc2543f7a47ae4>::__this': is not a type name, static, or enumerator
..\gui\Src\Gui\BreakpointsView.cpp(145): error C2065: '__this': undeclared identifier
..\gui\Src\Gui\BreakpointsView.cpp(145): error C2227: left of '->s' must point to class/struct/union/generic type
..\gui\Src\Gui\BreakpointsView.cpp(145): note: type is 'unknown-type'`

The second one seems to be a compiler error. MSVC gets confused when this is referenced in the member function of a struct that was declared inside a lambda which also captures this.
As a workaround I removed the unnecessary this-> references in the operator<.

### DIFF
--- a/src/gui/Src/Disassembler/CsQBeaEngine.cpp
+++ b/src/gui/Src/Disassembler/CsQBeaEngine.cpp
@@ -271,9 +271,9 @@ Instruction_t CsQBeaEngine::DecodeDataAt(byte_t* data, duint size, duint origBas
     //tokenize
     CapstoneTokenizer::InstructionToken cap;
 
-    auto & infoIter = dataInstMap.find(type);
+    auto infoIter = dataInstMap.constFind(type);
     if(infoIter == dataInstMap.end())
-        infoIter = dataInstMap.find(enc_byte);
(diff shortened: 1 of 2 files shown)


mrexodia:
❤️

PR #1798, opened 2017-11-05
Multiple enhancements

torusrxxx:
NTSTATUS format function is not used in register view currently because C0000005 message is bad.

### DIFF
--- a/src/dbg/formatfunctions.cpp
+++ b/src/dbg/formatfunctions.cpp
@@ -38,11 +38,10 @@ void FormatFunctions::Init()
     {
         std::vector<wchar_t> helpMessage(destCount);
         String errName = ErrorCodeToName((unsigned int)code);
-        if(errName.size() == 0)
-            errName = StringUtils::sprintf("%08X", DWORD(code));
 #ifdef _WIN64
         if((code >> 32) != 0)  //Data in high part: not an error code
         {
+            errName = StringUtils::sprintf("%p", code);
             if(destCount < errName.size() + 1)
                 return FORMAT_BUFFER_TOO_SMALL;
             else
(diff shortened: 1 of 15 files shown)


src/dbg/formatfunctions.cpp
```diff
@@ -38,11 +38,53 @@ void FormatFunctions::Init()
...
+            }
+        }
+    });
+
+    Register("ntstatus", [](char* dest, size_t destCount, int argc, char* argv[], duint code, void* userdata)
+    {
+        std::vector<wchar_t> helpMessage(destCount);
+        String errName = ErrorCodeToName((unsigned int)code);
 #ifdef _WIN64
         if((code >> 32) != 0)  //Data in high part: not an error code
```
mrexodia: The high part of NTSTATUS codes are usually set to something like 0xFFFFFFFFC0000005 so this check would say the code is invalid while it is actually valid...

src/gui/Src/Gui/DataCopyDialog.cpp
```diff
@@ -16,23 +16,23 @@ DataCopyDialog::DataCopyDialog(const QVector<byte_t>* data, QWidget* parent) : Q
     mTypes[DataCWord] = FormatType { tr("C-Style WORD (Hex)"), 12 };
     mTypes[DataCDword] = FormatType { tr("C-Style DWORD (Hex)"), 8 };
     mTypes[DataCQword] = FormatType { tr("C-Style QWORD (Hex)"), 4 };
-    mTypes[DataCString] = FormatType { tr("C-Style String"), -1 };
```
mrexodia: Why are these changes made?

torusrxxx: Because the UI has changed to prevent user from setting items per line to 0.

mrexodia:
👍

PR #1802, opened 2017-11-07
DBG: fix saving CPUDump and mmap dump to file

### LINKED ISSUE #1778: "Failed to write file" when dumping memory to file
Hello, all.
I have recently started using x64dbg. I have a file named "sample.exe", written in C++. Here is the source code, if it helps.

```
#include <iostream>

using namespace std;

int main(){
	int input;
        cout << "Please enter key...\n" << endl;
	cin >> input;
	if (input == 55){
		cout << "Correct!" << endl;
	}
	else{
		cout << "Incorrect!" << endl;
	}
}
```

I am trying to dump the memory located in the .text section. For reasons which I am unaware of, whenever I try to dump memory through the memory map to a .bin file, the result always ends up being that nothing happens and I am given an error that simply states "failed to write file" located at the bottom of the window.
I am at a loss as to what causes this. I am using the latest version of x64dbg, I run it as an administrator, and it does not matter what executable I try this on or what section is dumped, it always results in the same error.
Any help would be appreciated, thanks.

rmbreak:
fixes #1778 

`stringformatinline(argv[1])` was replacing a file path with \\n with a newline. For example, C:\Users\Desktop\Me\nogood.bin would cause `name = "C:\\Users\\Desktop\\Me\nogood.bin"`.

### DIFF
--- a/src/dbg/commands/cmd-memory-operations.cpp
+++ b/src/dbg/commands/cmd-memory-operations.cpp
@@ -165,7 +165,7 @@ bool cbInstrSavedata(int argc, char* argv[])
         return false;
     }
 
-    String name = stringformatinline(argv[1]);
+    String name = argv[1];
     if(name == ":memdump:")
         name = StringUtils::sprintf("%s\\memdumps\\memdump_%X_%p_%x.bin", szProgramDir, fdProcessInfo->dwProcessId, addr, size);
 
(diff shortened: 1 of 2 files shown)


mrexodia:
I think a better fix would be to actually remove the (totally buggy) `StringUtils::ReplaceAll(format, "\\n", "\n")` from the string formatting functions because it's not documented that it can do that anyway 😄

rmbreak:
Ah, that would have been worse than where we started :p. Just added back `stringformatinline(argv[1])` and removed the `ReplaceAll` in both string formatting functions.

mrexodia:
Cheers!

PR #1806, opened 2017-11-08
GUI: show Follow->Constant even if same as Address

### LINKED ISSUE #832: Following a Constant dump
Example if i want to follow dump 0x579d4903 in the instruction line:
57912030 | 0F B6 88 03 49 9D 57     | movzx ecx,byte ptr ds:[eax+579D4903]
that only works when eax is not 0
screenshots:
dump
dump2

rmbreak:
Before this change, `CPUDisassembly::setupFollowReferenceMenu` would not add `Constant` to the list if `Address` had the same value. I think it's okay to show two entries even if they have the same value for consistency. Let me know what you think.

fixes #832

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -169,14 +169,14 @@ void CPUDisassembly::setupFollowReferenceMenu(dsint wVA, QMenu* menu, bool isRef
                 if(arg.segment == SEG_FS)
                     segment = "fs:";
 #endif //_WIN64
-                QString constant = ToHexString(arg.constant);
-                if(DbgMemIsValidReadPtr(arg.constant))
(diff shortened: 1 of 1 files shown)


rmbreak:
Actually this doesn't look very good with instruction like `call [0x00581124]`. In this case the follow menu looks like:

> Address: 0x00581124
> Constant: 581124

Would it make more sense for the code to check if arg contains more than just a constant and if so, then display constant otherwise just display address?

rmbreak:
@mrexodia I just took your advice in issue 832 about inverting the constant and address checks. I didn't find a way to check if a DISASM_ARG has both a constant component and a reg component.

mrexodia:
I'm not sure about this change and I have to do some testing before I can merge it...

mrexodia:
Thanks!

Issue #1805, opened 2017-11-08
Anyone want <CTRL+TAB> working like OD or Qt Creator or Visual Studio?


lynnux:
I feel very comfirtable with the way that <CTRL+TAB>/<CTRL+SHIFT+TAB> working in ollydbg, and not get used the way in x64dbg, although there are enough shortcuts to quickly locate the view where I cared.
After some search, I decide to stole code from the source code from qt creator, and it's almost done:
![1](https://user-images.githubusercontent.com/367894/32537167-9233ad42-c49c-11e7-8e1e-dcbdfab46476.png)
althrough it's just a demo, but can be very quickly integrated in x64dbg, there is only one thing that should take care of is the MHDetachedWindow.

mrexodia:
I don't see the difference between OllyDbg and x64dbg? For Olly it goes to the next window (whatever that means), in x64dbg it goes to the next tab (exactly like chrome).

Another idea is to maybe steal Ctrl+Shift+P from sublime text (something like this: https://github.com/TheCrazyT/x64dbg-plugin-quickaccess) and allow people to search the view they want by name.

mrexodia:
Perhaps @Atvaark has an opinion on this since he implemented Ctrl+Tab in the first place 😄

lynnux:
Yes, currently x64dbg goes the way like chrome, but in ollydbg Ctrl+Tab goto previous popuped window(like the detached window inx64dbg) in history stack. Consider the scenario: I press CTRL+R on a address in CPU, it turns into reference view, if I can't find any useful information, then I should go back, in olly it's just Ctrl+Tab, in x64dbg there has a shortcut for CPU view is Alt+C. But if the previous view is not CPU view, then I should recall where the view I come from, and trigger the corresponding shortcut?

lynnux:
Should close now, implement at pull [1807](https://github.com/x64dbg/x64dbg/pull/1807)

mrexodia:
Will be closed when #1807 is merged

PR #1807, opened 2017-11-09
Tab switch using history stack

lynnux:
Implement [#1805](https://github.com/x64dbg/x64dbg/issues/1805), 
Press <Alt+Right>/<Alt+Left> to give a try. It's maybe a little strange in <Alt+Left>, usually <Alt+Right> is enough, it quickly switch to the previous view.

use the popup code from the qt creator source, like this:

also, detached window was well handled.


lynnux:
wait... maybe I should assign shortcut <Alt+Left>and<Alt+Right> to this, so user can cutomize the shortcuts.

lynnux:
In recent commit, I  rename HistoryViewsPopupWindow to MultiItemsSelectWindow, the idea is that:
I wish there would have some shortcuts for "Follow in Dump/Disassembler/Stack/MemoryMap", but usaully it will have more than one choice to select:

So it's can't to select one without mouse. My idea is binding a shortcut like <CTRL+1> to "Follow in Dump", when press <CTRL+1>, it popup a window like:

to choose one entry you don't need mouse, only continue to press <CTRL+1> it will goto next entry, and you can bind <CTRL+SHIFT+1> to go back. same idea can apply to other "Follow in ..."

mrexodia:
Hm, one of my messages got lost in Github's email (@support)

You shouldn’t do any shortcuts with Alt... I think if we change the behaviour it should replace Ctrl+Tab

mrexodia:
Also that idea for the dump sounds useful! I was thinking of changing the double click so a menu pops up with various options where to follow it...

lynnux:
Don't know which way people like, my init implement is to replace CTRL+TAB, but they can coexist now.  In private, I will bind CTRL+TAB to this.
Double click may encounter a problem that which target it will follow, my idea is that CTRL+1 to follow in disassembler, CTRL+2 to  follow in dump, CTRL+3 to follow in stack, no matter where you locate at.  Of course you can place all options in one menu, but it's may not straight forward, need people to think about which entry he/she should choose.

mrexodia:
Thanks! I swapped the hotkey for this with previous/next tab because it makes more sense, I also made some small style changes but it works fine!

torusrxxx:
x64dbg_translations.pro

mrexodia:
Hm I have to remove that file some day, the translator has support for files from multiple sources...

But I will update it

Issue #1787, opened 2017-10-31
Set company and other metadata for x64dbg executables


torusrxxx:
![x64dbg-modules](https://user-images.githubusercontent.com/15761310/32217130-f9108b4a-be1e-11e7-9e89-40704f44075a.png)

mrexodia:
See https://github.com/x64dbg/x64dbg/pull/1809#issuecomment-344297972

PR #1809, opened 2017-11-09
Companyname

mrfearless:


mrfearless:
Company name info for #1787 but only for x64dbg (excluding x32gui.dll and x64gui.dll - not sure how to add version info to the Qt stuff) 
Also other dll's belong to other open source like capstone.dll, yara etc - should we add meta info for them?
I did add a version manifest to x32_bridge.dll and x64_bridge.dll (compatibilty dlls) but dont think they got added to the dependency @deps folder

src/exe/resource.rc
```diff
@@ -67,17 +67,20 @@ VS_VERSION_INFO VERSIONINFO
...
         BLOCK "040904b0"
         BEGIN
+            VALUE "CompanyName", "x64dbg.com"
             VALUE "FileDescription", "x64dbg"
             VALUE "FileVersion", "0.0.2.5"
+            VALUE "InternalName", "x64dbg.exe"
             VALUE "LegalCopyright", "x64dbg.com"
-            VALUE "ProductName", "x64dbg"
+            VALUE "OriginalFilename", "x64dbg.exe"
+            VALUE "ProductName", "x64dbg.exe"
```
tambry: Isn't the product "x64dbg" instead?

mrexodia:
Sorry for the wasted time, but x64dbg is not run by a company so adding it to version info makes no sense. The only thing it serves is to have crash dialogs show 'x64dbg' instead of the executable name but even that I find doubtful (especially the version info is complete crap)...

PR #1811, opened 2017-11-13
GUI: sortable memory map headers for type and protections

tarterp:
Sorting more of the memory map headers is useful. Finding reflective DLL injection ore other RWE regions is a lot quicker when able to sort.


torusrxxx:
Thank you very much!

mrexodia:
@torusrxxx next time please 'rebase and merge' to keep the history clean...

PR #1812, opened 2017-11-14
Let goto file offset dialog has a init value of current selected address

lynnux:
like IDA, set init value of current selected address. It's convenient when there is no shortcut binding for copy file offset, but goto file offset having a shortcut CTRL+SHIFT+G, you can press CTRL+SHIFT+G and CTRL+C to copy file file offset of current selected address in CPU disassembler or dump window.

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -1003,6 +1003,17 @@ void CPUDisassembly::gotoFileOffsetSlot()
     mGotoOffset->fileOffset = true;
     mGotoOffset->modName = QString(modname);
     mGotoOffset->setWindowTitle(tr("Goto File Offset in ") + QString(modname));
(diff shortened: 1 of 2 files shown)


mrexodia:
You can also press `Alt+Insert`...

lynnux:
@mrexodia , Alt+Insert is copy address, this is copy file offset

mrexodia:
Hm, for that you can also bind a hotkey to copy -> file offset 😄

lynnux:
OK, close the pull, I just like the way it works in IDA. New shortcut binding may have more burden of memory.

mrexodia:
I don't think the pr has to be closed. I was just giving alternatives to this, if the text is selected when the dialog opens I don't have anything against this...

lynnux:
Yes it's auto selected, I've already tested.

mrexodia:
Thanks!

Issue #1759, opened 2017-10-13
0xC0020043 RPC_NT_INTERNAL_ERROR on closing x64dbg


mrexodia:
This issue is extremely hard to reproduce, but does not really impact x64dbg since it only happens on exit. You lose the changed INI settings, but the database should be saved correctly.

Environmental notes:
- Windows 8.1 x64
- Happened to me in lots of x64dbg versions
- x64dbgpy
- I have `Wacom_Tablet.dll` driver installed (for Wacom Intuos Manga), but the thing is not plugged in. Also includes `Wintab32.dll`
- `Start8_64.dll` is also loaded
- Working from total commander

Reproduction (somewhat consistent, but takes a few tries):
- download [AdR.Samples.NativeCallingCLR.ClrAssembly.dll](https://mega.nz/#!a5xwzA5S!8N92YKxhVMPGpClaNvM27KwWPZQXdIoajwRhvr5S7bw)
- Right click -> open in x64dbg
- Try to close x64dbg as fast as possible
- It will crash with `0xC0020043 RPC_NT_INTERNAL_ERROR`

Some analysis results:
- RpcReportFatalError throws this exception when certain asserts are triggered in `LRPC_CASSOCIATION::IoCompleted` (rpcrt4.dll)
- Actual problem: `STATUS_INVALID_HANDLE` is returned from `NtAlpcSendWaitReceivePort` (this might actually be `STATUS_OBJECT_TYPE_MISMATCH` instead)
- WinDbg completely messes up the call stack and leads you to `Ndr64SystemHandleBufferSize` which is not at all relevant
- Lots of threads are just waiting around (and loads of them inside `RpcpReportFatalErrorExceptionFilter`!)
- One of the threads is chilling in `msvcr90.dll!free`, called from x64dbgpy but I think this is just random.
- There appears to be a thread with qt5network
- RPC error also happened: 0xC0000008

mrexodia:
I cannot reproduce the issue anymore, even though I did it like 5 times in a row just now...

Mattiwatti:
I tried this myself expecting it not to reproduce, but I actually got the same results, and (same as you) inconsistently. The code being debugged doesn't seem to matter, I tried it with a
```asm
nop
xor eax,eax
ret
```
executable and could still reproduce the error. Same for x64dbgpy or no x64dbpy.

The correct stacktrace is this:
```
RPCRT4!ReceiveMessage+0x210
[...]

PR #1824, opened 2017-11-27
Do not keep duplicate process and initial thread handles around

Mattiwatti:
See #1759 for discussion (but scroll to the bottom, that ticket discusses like 3 separate bugs in one thread).

In the case that x64dbg starts the debuggee process, it keeps two unneeded handles around with `PROCESS_ALL_ACCESS` and `THREAD_ALL_ACCESS` rights respectively. These are duplicated in `nt!DbgkpOpenHandles` when the first debug event (`CREATE_PROCESS_DEBUG_EVENT`) is sent. The latter two handles are automatically closed by kernel32 after `EXIT_PROCESS_DEBUG_EVENT` is handled. The original handles from `CreateProcess` (done by TitanEngine) are never closed however. You can check this by starting a debug session, stopping the process, restarting the session, and so on: Process Hacker will show a leaked process handle in x64dbg.exe to a non-existent process for each session that was started.

This PR closes the initial handles received from `CreateProcess` just before calling `DebugLoop()`. `cbCreateProcess` then sets `fdProcessInfo->hProcess` and `fdProcessInfo->hThread` to the new handle values before doing anything else. This does mean there is a tiny window during which these handles are NULL, but as they aren't sent to plugins I don't think this is an issue.

The call to `SafeSymCleanup()` has been moved to `cbExitProcess`, so that this always happens before the debuggee process has terminated. This fixes a crash that can occur if you close x32dbg (but strangely not x64dbg?) while the debuggee is still running.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1457,6 +1457,8 @@ static void cbExitProcess(EXIT_PROCESS_DEBUG_INFO* ExitProcess)
     _dbg_animatestop(); // Stop animating
     //unload main module
     SafeSymUnloadModule64(fdProcessInfo->hProcess, pCreateProcessBase);
+    //cleanup dbghelp
+    SafeSymCleanup(fdProcessInfo->hProcess);
     //history
     dbgcleartracestate();
     dbgClearRtuBreakpoints();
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

Issue #1259, opened 2016-11-17
String viewer to view long string


torusrxxx:

akhil117:
It means we can enter the array of strings and with the help of string viewer we should know the long string name ?
Is it?

mrexodia:
No, it would be a simple dialog that is opened when you right click in the dump that allows you to view and copy the long string.

For example:

![](https://i.imgur.com/VXzSHT1.png)

Right click on that byte, 'view string' and then it will show a multi-line text editor that allows you to interact with the string. Perhaps the Data Copy dialog could be used to display the string:

![](https://i.imgur.com/zlwLwq4.png)

PR #1859, opened 2017-12-31
View and copy raw string

torusrxxx:
#1259


mrexodia:
Sorry, I totally forgot about this...

Issue #1880, opened 2018-01-27
FeatureRequest: ASCII column for Address View in Dump


cw2k:
Well the **Address** View in Dump should have an ASCII column. 

Ollydebug is a good example:
![Ollydebug with ASCII column](https://user-images.githubusercontent.com/3834079/35467755-e3aee326-0312-11e8-940a-2b0de503f37a.png)

The ASCII-column is really useful to quickly identity text (without the need to switch the view)
![x64dbg](https://user-images.githubusercontent.com/3834079/35467891-e6156430-0314-11e8-8c88-41882df52ab4.png)
That these hex numbers are text is not that obvious.

Maybe if its not that much work make the ASCII column toggle between **Unicode** and **ASCII** when you click on it. Here is how Ollydbg looks in Unicode mode:
![screenshot - 27 01 2018 03_47_47](https://user-images.githubusercontent.com/3834079/35467948-f211af7c-0315-11e8-9e62-07c9a330ae6c.png)

=> Hmm or maybe just add both columns ASCII and Unicode

PR #1885, opened 2018-01-30
issue #1880 swich bettween ASCII and UNICODE column

aikuimail:


wertyef:
STOP SENDING ME EMAILS

src/gui/Src/Gui/CPUDump.cpp
```diff
@@ -515,6 +515,30 @@ void CPUDump::mouseMoveEvent(QMouseEvent* event)
...
+        if(getColumnCount() && (event->y() <= getHeaderHeight()) &&event->y() >= 0 )
+        {
+            bool wHasCursor = cursor().shape() == Qt::SplitHCursor ? true : false;
+            if(!wHasCursor)
+            {
+                int wColIndex = getColumnIndexFromX(event->x());
+                if( wColIndex == 2)
+                {
+                    QString title = getColTitle(wColIndex);
+                    if(!title.compare("UNICODE"))
```
mrexodia: This is not really a property you can depend on and it would only work on the English version...

mrexodia:
Unfortunately this does not appear to solve #1880, but instead it allows you to switch modes by clicking on a column once(?). I think this could be confusing to users...

hummadhassan:
Unsubsidized me from this shit

src/gui/Src/BasicView/HexDump.h
```diff
@@ -77,6 +77,7 @@ class HexDump : public AbstractTableView
         int separator = 0;
         QTextCodec* textCodec; //name of the text codec (leave empty if you want to keep your sanity)
         DataDescriptor_t data;
+        std::function<void(QObject *obj)> columnSwitch;
```
mrexodia: The `QObject` parameter for `this` is not needed because you can capture `this` (see later comment).

src/gui/Src/Gui/CPUDump.cpp
```diff
@@ -1229,7 +1230,69 @@ void CPUDump::addressSlot()
...
+    wColDesc.itemCount = 8;
+#else
+    wColDesc.itemCount = 4;
+#endif
+    wColDesc.data.itemSize = Byte;
+    wColDesc.data.byteMode = AsciiByte;
+    wColDesc.columnSwitch = [](QObject *obj){
+            CPUDump *pDump = qobject_cast<CPUDump*>(obj);
+            pDump->setView(ViewAddressUnicode);
+    };
```
mrexodia: This would then become:

```c++
wColDesc.columnSwitch = [this]()
{
    this->setView(ViewAddressUnicode);
};
```

src/gui/Src/Gui/CPUDump.cpp
```diff
@@ -1229,7 +1230,69 @@ void CPUDump::addressSlot()
...
+    wColDesc.isData = true;
+    wColDesc.separator = 0;
+#ifdef _WIN64
+    wColDesc.itemCount = 4;
+#else
+    wColDesc.itemCount = 2;
+#endif
+    wColDesc.data.itemSize = Word;
+    wColDesc.data.wordMode = UnicodeWord;
+    wColDesc.columnSwitch = [](QObject *obj){
```
mrexodia: Make sure to run `format.bat` before you commit (will go automatically if you run `install.bat`)

src/gui/Src/Gui/CPUDump.h
```diff
@@ -144,6 +146,7 @@ public slots:
         ViewFloatDouble,
         ViewFloatLongDouble,
         ViewAddress,
+        ViewAddressUnicode,
```
mrexodia: I think it's better to add two new views `ViewAddressAscii` and `ViewAddressUnicode` in addition to the normal `ViewAddress`.

Also please add new enums to the end (because users of old versions will now get a different dump view when they open this updated x64dbg).

mrexodia:
Thank you for the good work 👍

I slightly changed your implementation to make it an option since it looks weird to give it an extra menu item.

PR #1897, opened 2018-02-10
Enhancements to the Data Copy dialog

### LINKED ISSUE #1862: About Data Copy Dialog
Would be nice if there has the functions from wireshark's menu:
default

And it's may be better if droplist change to listbox, like this:
default

ThunderCls:
Closes #1862


ThunderCls:
PR update coming, don't merge yet

ThunderCls:
Ok I guess now its done!

src/gui/Src/Gui/DataCopyDialog.cpp
```diff
@@ -3,6 +3,7 @@
 #include "Bridge.h"
 #include <QCryptographicHash>
 #include <QTextCodec>
+#include <QMessageBox>
```
torusrxxx: Where is this message box used?

torusrxxx:
Hex stream is already available in copy menu. Probably we don't need to duplicate it here? And I have an idea to combine hex editor, multiline string editor and data copy dialog into a dialog with tabs. Is it a good idea?

src/gui/Src/Gui/DataCopyDialog.cpp
ThunderCls: debugging purposes...I forgot to take it out xD

ThunderCls:
+1
It sounds fancy.

mrexodia:
@ThunderCls is this ready to merge?

ThunderCls:
Well I should take that QMessageBox include out, also not sure if I should keep the hex stream or not since this feature in the copy menu adds spaces after every pair of digits and sometimes that could be a bummer. Other than that, yeah I guess it is ready.

mrexodia:
Thanks!

torusrxxx:
x64dbg-combined-hex-editor
I designed a new Data Copy dialog and Hex Editor, but the thick border makes it less appealling to me. Any advice?

mrexodia:
It’s difficult to say without the standard Windows color scheme, but the left one looks alright to me or is that not what you mean?

torusrxxx:
The new dialog on the left is less compact than original one on the right due to additional border and margin. You can try out at https://github.com/torusrxxx/x64dbg/tree/patch00000089

torusrxxx:
latest update of x64dbg hex editor
It feels better now and I have removed the old data copy dialog.

PR #1903, opened 2018-02-20
group bytes of disasm bytes

torusrxxx:


mrexodia:
> <torusrxxx> it supports relocation and patches. The only thing to do is that relocation underline will also underline the space after the pointer.

mrexodia:
@torusrxxx are the bytes supposed to be not added to the HTML clipboard?

mrexodia:
Looks vert good!

s

PR #1911, opened 2018-03-06
Update README.md

in73ct0rd3vil:


mrexodia:
Who is Arpit Jain and why do they get special thanks?

PR #1947, opened 2018-05-19
Load language menu only when needed

torusrxxx:
Saving a directory scanning operation on every x64dbg startup when the hard disk is busy loading x64dbg


mrexodia:
I manually merged this pull request. I removed the commit for customizing the header background because it was impossible to look good. The border color also has to be defined, it should be possible to do based on the text color and background color but I'm not sure how (#1953)

Issue #1961, opened 2018-06-08
when a process is detected in the handles view show pid+name


mrexodia:

torusrxxx:
I left out it when writing the handles view due to concerns of
compatibility between different versions of operating system (type numbers
are different).

mrexodia:
Hm, but somehow it is correctly detected as a process so perhaps we could just try to get the pid after cloning the handle and report that if successful?

PR #1963, opened 2018-06-09
show PID or TID in handles view

torusrxxx:
#1961. Tested on Windows XP. It doesn't show process name yet because it is somewhat more expensive.

### DIFF
--- a/src/dbg/handles.cpp
+++ b/src/dbg/handles.cpp
@@ -58,7 +58,7 @@ static DWORD WINAPI getNameThread(LPVOID lpParam)
 bool HandlesGetName(HANDLE hProcess, HANDLE remoteHandle, String & name, String & typeName)
 {
     HANDLE hLocalHandle;
-    if(DuplicateHandle(hProcess, remoteHandle, GetCurrentProcess(), &hLocalHandle, 0, FALSE, 0))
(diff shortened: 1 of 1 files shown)


torusrxxx:
This assumes the type name "Process" and "Thread" is consistent across Windows versions (seems true). Process names can be retrieved with "GetModuleBaseName" (not done yet). For Windows 10, "GetThreadDescription" might be interesting.

mrexodia:
I tested this on Windows 8.1 and it was working, but on Windows 10 I had some trouble, I will investigate it further...

mrexodia:
Hi, I found a minimal executable that should reproduce the issue I was having:

```
#include <windows.h>

int main()
{
    auto pid = GetCurrentProcessId();
    auto hp = OpenProcess(PROCESS_TERMINATE, FALSE, pid);
    TerminateProcess(hp, 0);
}
```

`GetProcessId` requires `PROCESS_QUERY_INFORMATION`, but duplicating with the same access won't grant it in this case.

torusrxxx:
How do you think to fix this problem? If the debuggee doesn't open a
process with query privileges, its access is very limited, and it is very
likely the debuggee already knows its PID in its memory. I don't want to
fix this bug as almost all process operations require query privilege. A
handle without query privilege can't be useful for viral transmission.

mrexodia:
I think a clean fix would be to re-duplicate the handle with query privileges once we see it’s a process or thread handle. This should be undetectedable and pretty clean (DuplicateHandle has a parameter for access). Maybe today I will try something if I have some time...

Issue #1086, opened 2016-09-14
Can't delete breakpoint if module is not loaded


ViRb3:
Say I am analyzing a program which loads a DLL dynamically. If I breakpoint a function in that dynamically loaded DLL and the DLL gets unloaded, the breakpoint cannot be removed until I load the module again. Upon pressing the button x64dbg says "Breakpoint [address] not found."

ViRb3:
Here's an example with Notepad.
1. Break on EP.
2. Enable "break on DLL load" and continue execution. You should break on the loading of "uxtheme.dll".
3. Set breakpoints in the "uxtheme.dll" module, like so:
https://i.imgur.com/4qPOciQ.png
4. Restart the program and try to delete the breakpoints. Mind the message bottom-left.
https://i.imgur.com/RkEPwZX.png

Nukem9:
This fails because BpGet is using the address of the breakpoint only. The name is blank so it doesn't search for it. Module isn't loaded -> `ModHashFromAddr` fails.

@ViRb3 if you really need to delete them, you can set a name for the breakpoints (then remove em). Otherwise you'll need to wait.

ViRb3:
@Nukem9 Thanks for the nice workaround! I wasn't really bothered by it, just wanted to put it up here. Probably a BP index system should be sufficient to fix this issue.

bkbiradar:
Just

PR #1965, opened 2018-06-15
Allow editing and removing inactive breakpoints

torusrxxx:
This is a tracking branch. Currently it adds a special format DLL:$RVA that BpGet recognizes even if DLL is not loaded, so commands like "bpc" can remove these breakpoints. GUI is under development.

Other bugs fixed:
* Allow going to entrypoint of libstdc++-6.dll in SymbolsView


torusrxxx:
closes #1086

Issue #1841, opened 2017-12-17
Allow sorting FPU registers by logical names


mrexodia:
See #1837

IssuehuntBot:
@oliver-u funded this issue with $2.0. [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1841)

IssuehuntBot:
@0maxxam0 funded this issue with $10. [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1841)

torusrxxx:
fixed.

PR #1968, opened 2018-06-20
Sort FPU registers in ST(X) in registers dialog

torusrxxx:
The user can also edit ST registers, or switch to use x87rX ordering. When in this mode, editing x87SW_TOP doesn't result in content swapping in ST(X), so there might be some more work to do. (#1841)


torusrxxx:
When x87SW_TOP is modified, ST(X) is not swapped, instead x87rX is swapped. This behaviour is also observed in current releases, so is it a problem anymore?

mrexodia:
I don’t know. If you think that is fine behavior it is okay for me :)
probably people expect it to match olly, so if that’s the case I’m all for
it.

I held off merging because I thought you wanted to update the merge request
to address the issue with the swapping. If not I will check out the code in
more depth and do some testing...

torusrxxx:
I'll test it to see if there is anything wrong. There are a few updates to this patch.

torusrxxx:
`FSAVE` instruction saves each ST(x) register to destination, and `FRSTOR` instruction loads each ST(x) register from memory. It doesn't import and export in x87rx. So the swapping behaviour is a normal phenomenon, not a problem.

torusrxxx:
This patch appears to be working. You can try it when the program has used some of x87 register stack, not when it is empty.

mrexodia:
almost ready, just have to remove some 'temp' commits

mrexodia:
also

> <torusrxxx> Please rename "Copy floating point value to clipboard" to "Copy floating point value". I don't see any point to emphasis "to clipboard".

PR #1992, opened 2018-07-20
Build date plugin export

LFriede:
Hi,
I'm the dude with the updater plugin. I used the modified file date of "x64dbg.exe" previously to compare it with the latest snapshots timestamp. I think it would be nicer to comare it against the build date so I added a plugin export for that.


torusrxxx:
Hi, guy. There is a "commithash.txt" file in each release of x64dbg, which you can check against commit history on Github.

LFriede:
Ah, damn I'm so stupid :D It was before my eyes the whole time. Thanks!

Issue #1960, opened 2018-06-07
x64dbg two fatal bug


number201724:
Bug 1:
If the EXE EntryPoint is 0, Entry Breakpoint breaks to the PE Header's IMAGE_DOS_SIGNATURE

Bug2:
If you use ZwMapViewOfSection to load 2 copies of ntdll.dll (path is consistent) then the breakpoint on this module will be wrong. ZwOpenFile -> ZwCreateSection -> ZwMapViewOfSection

number201724:
Some programs use TLS CALLBACK to start.

mrexodia:
Zero is a valid entry point. See https://corkamiwiki.github.io/PE

For bug 2, could you provide some clearer information? Example executable or steps perhaps.

number201724:
Yes 0 is a valid entry point, but the breakpoint of x64dbg is in the PE header instead of the code segment.

number201724:
Bug2 I will write a demo later

abbasfull:
No problem

number201724:
bug example:
https://github.com/number201724/x64dbg_bugreport

number201724:
Example 1: When the entry breakpoint is turned on, he crashes in ntdll

number201724:
For 0 entry point, he should not set the breakpoint to the IMAGE_DOS_SIGNATURE of the PE header.

number201724:
This will cause the ntdll LdrXXXXX function to crash.

Mattiwatti:
Alright alright. Your first PE now runs for me with the above PR applied.

number201724:
After a few months, there is still no repair

torusrxxx:
I support the idea to add number suffix to duplicated DLL.

Mattiwatti:
The problem with adding the image base to the module name (unless this name is only used for display/GUI purposes) is that it would break any user breakpoints and patches upon the next start (due to ASLR), no? That said, for display purposes having the base address appended does seem useful.

If we make the PE malicious enough this seems like a problem that is not solvable in the general case IMO:
1. Genuine `ntdll.dll` gets loaded
2. Process loads a pseudorandom number (say, between 10 and 30) of `ntdll.dll`s based on some internal algorithm and maps them as `SEC_IMAGE`. Note that you can specify a preferred load address, which, assuming the requested address range is not already taken, will work. This 'ASLR-ness' could also be randomized for each module, making some load at the requested base address, and others at a random one.
[...]

PR #1994, opened 2018-07-29
Don't crash in zero entry point PEs if entry breakpoint is enabled

Mattiwatti:
Ref: #1960 (first bug)

0 is a valid entry point for a PE, but ntdll really doesn't like it if you set a breakpoint in the DOS header, and it's not like the breakpoint would have been very useful anyway.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1389,7 +1389,7 @@ static void cbCreateProcess(CREATE_PROCESS_DEBUG_INFO* CreateProcessInfo)
                 dprintf(QT_TRANSLATE_NOOP("DBG", "%d invalid TLS callback addresses...\n"), invalidCount);
         }
 
-        if(settingboolget("Events", "EntryBreakpoint"))
+        if(settingboolget("Events", "EntryBreakpoint") && (duint)CreateProcessInfo->lpStartAddress - (duint)CreateProcessInfo->lpBaseOfImage != 0)
         {
             sprintf_s(command, "bp %p,\"%s\",ss", (duint)CreateProcessInfo->lpStartAddress, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "entry breakpoint")));
             cmddirectexec(command);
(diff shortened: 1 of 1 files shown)


mrexodia:
An entry point of zero is actually valid (there are some examples in the corkami repository). I think a proper fix would be to move setting the entry breakpoint to the system breakpoint callback.

Mattiwatti:
> An entry point of zero is actually valid (there are some examples in the corkami repository)

I know, that's what I said ;)

IMO an entry breakpoint for an image with no entry is nonsensical. Surely there must be an entry before there can be an entry breakpoint? Changing it to the system BP is basically changing the meaning of 'entry breakpoint' to 'system breakpoint' for 0.01% of images. I think it would confuse people more that the behaviour is inconsistent than that it would help anyone. If I want a system breakpoint, I just check the checkbox for that in the options.

The 'no-op' fix is also consistent with how the other breakpoint settings work. If I have 'TLS callbacks' checked, but the image doesn't have TLS callbacks, does x64dbg set a BP on the call to `LdrpCallTlsInitializers` as a replacement instead? No, that wouldn't make any sense.

mrexodia:
There is a misunderstanding here 😀

Currently the entry breakpoint is set on the cbCreateProcess debug event callback, which fucks the loader because the MZ signature is changed (the loader isn’t executing yet at this point)

I proposed to move setting of the entry breakpoint to the cbSystemBreakpoint debug even callback :)

mrexodia:
Oh and when an executable has a zero entry point it will start executing on the MZ signature as if it was code. Some special crafted pe file can then put a jmp right after that and transfer to the real entry point without making the PE file invalid. For DLLs it’s not possible to have a zero entry point.

mrexodia:
I just checked and it appears that you cannot set a breakpoint on the PE header even when doing it from the system breakpoint callback. Probably the only possibility is to fall back to hardware breakpoints if an entry point of zero is detected.

Sample (not malware):
[entry0_.zip](https://github.com/x64dbg/x64dbg/files/2283864/entry0_.zip)

Mattiwatti:
Ahh, I see. Yes, in that case setting the entry breakpoint 'as late as possible' would have been a good fix. Shame it doesn't work then :/

WinDbg has something interesting to say about your sample when I tried setting a hardware breakpoint:
> 0:000> ba e 1 entry0
    ^ Unable to set breakpoint error
The system resets thread contexts after the process
breakpoint so hardware breakpoints cannot be set.
Go to the executable's entry point and set it then.

So it sounds like the create process callback is the only possible moment to do this?

mrexodia:
Yeah I observed some very strange behavior in x32dbg too. I disabled the entry breakpoint to not make everything crash, then from the system breakpoint I put a hardware breakpoint on execute on the entry point. It stopped, but when I started stepping the execution resumed and I saw the message box.

Perhaps a good idea is to throw a big fat message box in the user's face and not care about this too much?

PR #1996, opened 2018-07-30
Update manifest.xml

callmenemo491:
just pretty-print


mrexodia:
Sorry but these kind of changes don’t really have a purpose and only cause the blame to be shifted to you without making any actual changes to the files so I’m closing the pull requests with just whitespace changes.

PR #2001, opened 2018-08-07
Use keys to select in dump

torusrxxx:
Arrow keys are now used to select in dump, not scroll in dump. Dump will scroll when selection is out of viewport. Shift key is to be supported.

### DIFF
--- a/src/gui/Src/BasicView/AbstractStdTable.cpp
+++ b/src/gui/Src/BasicView/AbstractStdTable.cpp
@@ -590,12 +590,12 @@ void AbstractStdTable::setSingleSelection(int index)
     emit selectionChangedSignal(index);
 }
 
-int AbstractStdTable::getInitialSelection()
+int AbstractStdTable::getInitialSelection() const
 {
     return mSelection.firstSelectedIndex;
 }
 
-QList<int> AbstractStdTable::getSelection()
+QList<int> AbstractStdTable::getSelection() const
 {
     QList<int> selection;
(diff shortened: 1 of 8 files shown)


mrexodia:
Looks nice! I think it would be nice to keep some kind of original behavior with Ctrl+arrows though 

Maybe
Ctrl + up: scroll up one line without changing selection
Ctrl + down: scroll down one line without changing selection
Ctrl + left: first visible address - 1 without changing selection
Ctrl + right: first visible address + 1 without changing selection

What do you think?

torusrxxx:
There is a problem with "first visible address ± 1" in floating number view. In development version of x64dbg, when you press ⇔ keys, the floating point numbers change into something meaningless, instead of displaying a next number. Therefore, I decide to shift a "word size" instead.

justanotheranonymoususer:
> when you press ⇔ keys, the floating point numbers change into something meaningless

Or into something meaningful, if the number wasn't aligned to begin with

mrexodia:
@justanotheranonymoususer my idea would be to allow using ctrl + left / right to move the view a single byte left/right. would this address your issue?

justanotheranonymoususer:
Yep. Since it's the less common scenario, maybe make ctrl+arrow move a word, and ctrl+alt+arrow move a byte.

mrexodia:
I tested and it appears to work very nicely @torusrxxx . Feel free to merge :)

PR #2005, opened 2018-08-21
Fixed a crash when tracing into far jump

torusrxxx:
Previously when you trace while recording a trace over a far jump, such as an API call, x32dbg crashes.


mrexodia:
Thanks, I merged both pull requests!

Issue #1972, opened 2018-06-28
FEATURE REQUEST - Add the ability to copy individual x87 registers


deniswisedeniswise:
Please add the ability to copy a specific floating point register by right clicking. I mean the registers ST,ST1,ST2...ST7
-Debugger version May 17 2018, 17:22:24
-Operating system version Windows 10

IssuehuntBot:
@0maxxam0 funded this issue with $4. [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1972)

IssuehuntBot:
@tathanhdinh has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1972)

IssuehuntBot:
@tathanhdinh has submitted [output](https://github.com/x64dbg/x64dbg/pull/2009). [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1972)

mrexodia:
@deniswisedeniswise there is an implementation for your request in #2009 by @tathanhdinh you can download a test snapshot at https://ci.appveyor.com/project/mrexodia/x64dbg/build/1.0.524/artifacts could you try and see if this satisfies your need?

tathanhdinh:
Thanks @mrexodia.

IssuehuntBot:
@mrexodia has rewarded. [See it on IssueHunt](https://issuehunt.io/repos/33791743/issues/1972)

PR #2009, opened 2018-08-27
Support copy individual x87 registers

tathanhdinh:
Hello all,
This PR tries to support #1972.

st_copy

I 'm a fan of `x64dbg` but still not familiar with its code base yet. Many thanks for any comment.


mrexodia:
Thanks!

PR #2010, opened 2018-08-29
Delete failed PDB downloads

Mattiwatti:
The VS2017 15.8 update updated the MS DIA SDK and added a bunch of new internal functions for a `PortablePDB` class, including `PortablePDB::ConvertPortablePDB()` which is called by `IDiaDataSource::loadDataFromIStream()`. I don't know what it's for, but hopefully this means that `/DEBUG:FASTLINK` PDBs are going to give more useful output.

Unfortunately there are two problems:
- `msdia140.dll` is sloppy and doesn't do basic validation like checking whether a PDB has a 0 file size.
- x64dbg is sloppy and creates 0 file size PDBs. (e.g. if the symbol server returned 404)

This means x64dbg will randomly crash on **some** apps: namely if the PDB of the app or a DLL used by it was previously queried unsuccessfully.

This PR expands the cleanup lambda in `downslib_download` to delete the file if the download failed, was cancelled or if the result is incomplete.

Note that this does not add any file size checks for **existing** PDBs, so it's probably a good idea to manually do a search in your symbols directory for .pdb files with size 0 and proactively delete them :smile:

### DIFF
--- a/src/dbg/WinInet-Downloader/downslib.cpp
+++ b/src/dbg/WinInet-Downloader/downslib.cpp
@@ -31,7 +31,18 @@ downslib_error downslib_download(const char* url,
     {
         DWORD dwLastError = GetLastError();
         if(hFile != INVALID_HANDLE_VALUE)
+        {
+            bool doDelete = false;
+            LARGE_INTEGER fileSize;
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #2035, opened 2018-10-04
GUI: Fix build for latest Qt

InKahootz:
I think this was introduced in Qt11. I saw a similar error in RPCS3 repo regarding forward declaration.

If anyone else could validate that it doesn't build on 5.11.1 without this fix would be great as I don't work on C++ that much.

Happy Hacktoberfest!


mrexodia:
Thanks!

PR #2036, opened 2018-10-05
ADDED: trace file can now auto-scroll the disassembly view to current…

wrongway88:
… selection

### DIFF
--- a/src/gui/Src/Tracer/TraceBrowser.cpp
+++ b/src/gui/Src/Tracer/TraceBrowser.cpp
@@ -1396,7 +1396,7 @@ void TraceBrowser::updateSlot()
     }
 }
 
-void TraceBrowser::toggleAutoDisassemblyFollowSelection()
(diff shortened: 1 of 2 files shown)


src/gui/Src/Tracer/TraceBrowser.cpp
```diff
@@ -1373,3 +1395,8 @@ void TraceBrowser::updateSlot()
         reloadData();
     }
 }
+
+void TraceBrowser::toggleAutoDisassemblyFollowSelection()
```
mrexodia: Probably it's better to rename this to `toggleAutoDisassemblyFollowSelectionSlot` to follow the convention of the other slots...

wrongway88: Alright, I renamed it

mrexodia:
Thanks!

PR #2041, opened 2018-10-12
FIXED: Uninitialized variable when calculating registers view height

wschoech:
_InitMappings()_ is now called before _getEstimateHeight()_.

_getEstimateHeight()_ calculates the height of the registers view. The calculation is based on _mRowsNeeded_ which is set in InitMappings(). However, _getEstimateHeight()_  was called before _InitMappings()_ and therefore the calculation was based on an uninitialized variable.


mrexodia:
Looks good, thanks!

PR #2042, opened 2018-10-13
Add support for disassembly popup for more views

torusrxxx:
Disassembly popup has been refactored to allow easy introduction to various views. It is now supported in Disassembly, Breakpoints, Call stack, SEH, Symbols, Sources, References, Threads and Windows view.


mrexodia:
Thanks, it appears to be working good in the references and symbols view!

PR #2043, opened 2018-10-13
add the expression function matchstrA and matchstrW.

cnsheds:
Use conditional breakpoints like this
matchstrW([esp+4], #0:"filename") or matchstrA([esp+4], #0:"filename")
Fixed a read memory failure when the breakpoint was broken because the memory page information was not updated.

### DIFF
--- a/src/dbg/exprfunc.cpp
+++ b/src/dbg/exprfunc.cpp
@@ -11,6 +11,7 @@
 #include "value.h"
 #include "TraceRecord.h"
 #include "exhandlerinfo.h"
+#include <regex>
 
 namespace Exprfunc
 {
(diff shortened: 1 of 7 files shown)


mrexodia:
Hey,

I do like your idea of allocating a string as a value, but I don't think this is a very safe idea. Because now passing something like `matchstrA([esp+4], 123)` will crash x64dbg. In the [strmatch](https://github.com/x64dbg/strmatch) plugin I did I have a separate command to allocate strings in a pool. Perhaps a similar thing could be done here?

Another solution would be to properly support strings in expressions, but that will require a rewrite that won't be pretty...

src/dbg/exprfunc.cpp
```diff
@@ -411,4 +412,59 @@ namespace Exprfunc
...
+                dprintf("matchstrA Error reading [0x%x+0x%.2x*1]\n", src, i);
+                return 0;
+            }
+
+            DbgMemRead(src + (i * step), dbgMemstr + i, step);
+            if(*(dbgMemstr + i) == '\x00')
+                break;
+        }
+
+        std::regex reg1((char*)patstr, std::regex::icase);
```
mrexodia: I was talking about this cast... If you use `matchstrA(rsp, 123)` it will be `(char*)123` or did I miss something?

cnsheds: matchstrA(rsp, 123) is the wrong way to use, if you mean to use the string "123", it should be like this
matchstrA(rsp, #:0"123")

mrexodia: I know it’s the wrong way to use it, but if you use it wrong anyway because of a mistake, will it crash?

cnsheds: Oh, I will add a function to avoid this error.

LuanDevecchi: What about using C++ style cast?

```cpp
const std::regex reg1(reinterpret_cast<char*>(patstr), std::regex::icase);
```

mrexodia: The cast probably won’t compile and if it does will still crash at runtime :)

The only solution here is to create a global string pool and let the patstr be the index in this pool...

mrexodia:
This will be properly implemented soon. The expression functions have been changed to also support string arguments now and implementing this should be easy.

Issue #1997, opened 2018-08-04
Dll ordinal


Portiella:
Hi,

How can I know the ordinal of a dll function? (x86)
In xDbg there is no ordinal column in the symbols tab ...

Thank you

Portiella:
I do not see Ordinal number.
Where is it?
Thx.
![xdbg1](https://user-images.githubusercontent.com/30263226/44142122-84dec568-a07f-11e8-873d-c35119d5a238.png)
![xdbg2](https://user-images.githubusercontent.com/30263226/44142123-84fc3634-a07f-11e8-83f7-13209271cc62.png)

mrexodia:
Local branch. This code is not yet in any release.

Mattiwatti:
FWIW, `dumpbin` does `Ordinal 1234`, IDA does `mfc42_1234`. Normally I'd prefer the IDA format because of database compatibility and such, but putting the module name in the export name seems very redundant IMO.

I prefer `Ordinal1234` (so no space, underscore or hash), simply because it's impossible to mistype that way.

lynnux:
Any update on this?
I see symbols view was huged refactoried, the new handler on ordinal functions of import are in `ReadImportDirectory`(src\dbg\module.cpp) , I modified like this, it got worked:
![grsmhz e 132rgv z2b_xln](https://user-images.githubusercontent.com/367894/49145888-cf386600-f33b-11e8-85e9-3f6a11f3159b.png)

mrexodia:
![Image](https://github.com/user-attachments/assets/7cc3206f-27c3-4cde-8f36-51a321fd7535)

There is now an ordinal column in the symbols view.

Portiella:
> ![Image](https://github.com/user-attachments/assets/7cc3206f-27c3-4cde-8f36-51a321fd7535)
> 
> There is now an ordinal column in the symbols view.

Perfect.

Thx

PR #2054, opened 2018-10-29
Add DLL ordinal to symbol table and fix symbol table comparator

balintf:
This PR fixes issue #1997 and also fixes the symbol table comparator for descending sort. Previously the comparator wasn't a strict weak ordering, and it failed the msvc debug checks on equal keys.

This PR changes the bridge API for DbgGetSymbolInfo by adding a DWORD ordinal member to the SYMBOLINFO struct. I'm not sure how desirable this is, it would at least require the pluginsdk to be updated as well (maybe there is a more compatible way to pass the ordinal that I haven't figured out).

### DIFF
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -598,6 +598,7 @@ typedef struct SYMBOLINFO_
     SYMBOLTYPE type;
     bool freeDecorated;
     bool freeUndecorated;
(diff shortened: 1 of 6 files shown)


mrexodia:
I don't think anybody is using this bridge api, so it shouldn't be an issue to change the SYMBOLINFO struct...

PR #2055, opened 2018-10-29
Remove Capstone references

### LINKED ISSUE #2013: Confused disassembler
I'm playing with the code base of x64dbg, there are many things that make me surprise, e.g. [get Capstone but return Zydis](https://github.com/x64dbg/x64dbg/blob/development/src/gui/Src/Disassembler/capstone_gui.cpp#L205):

```
const Zydis & CapstoneTokenizer::GetCapstone() const
{
    return _cp;
}
```

I believe that is because of some previous decision about using Capstone, then [changed](https://x64dbg.com/blog/2017/10/18/goodbye-capstone-hello-zydis.html) to use Zydis. Should we refactor these codes?

balintf:
This PR fixes #2013  by removing the deprecated references to Capstone. It also renames CapstoneTokenizer to ZydisTokenizer changing the filename to match the other gui filenames' conventions.

The PR also removes the commented out Zydis-Capstone diff code, since the Zydis migration seems to be more or less complete. If this diff code is still needed we can revert the last commit in this PR to get it back.


mrexodia:
Thanks for checking my merge conflict resolving!

PR #2057, opened 2018-10-30
Filter traced instructions

wrongway88:
Added the ability to filter the data displayed in the trace file browser by instruction.

For now it's just filter by instruction as string, there's no validation yet.

Filter can be applied (and removed) via context menu.

### DIFF
--- a/src/dbg/label.cpp
+++ b/src/dbg/label.cpp
@@ -83,11 +83,11 @@ bool LabelGet(duint Address, char* Text)
         if(found == tempLabels.end())
             return false;
         if(Text)
-            strcpy_s(Text, MAX_LABEL_SIZE, found->second.c_str());
+            strncpy_s(Text, MAX_LABEL_SIZE, found->second.c_str(), _TRUNCATE);
         return true;
     }
     if(Text)
-        strcpy_s(Text, MAX_LABEL_SIZE, label.text.c_str());
+        strncpy_s(Text, MAX_LABEL_SIZE, label.text.c_str(), _TRUNCATE);
     return true;
 }
 
--- a/src/dbg/memory.cpp
+++ b/src/dbg/memory.cpp
@@ -324,7 +324,7 @@ static bool IgnoreThisRead(HANDLE hProcess, LPVOID lpBaseAddress, LPVOID lpBuffe
     if(fnQueryWorkingSetEx(hProcess, &wsi, sizeof(wsi)) && !wsi.VirtualAttributes.Valid)
     {
         MEMORY_BASIC_INFORMATION mbi;
-        if(VirtualQueryEx(hProcess, wsi.VirtualAddress, &mbi, sizeof(mbi)) && mbi.State == MEM_COMMIT && mbi.Type == MEM_PRIVATE)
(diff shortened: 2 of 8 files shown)


torusrxxx:
Could you explain why you think this feature would be helpful to most users, and what's your next thing to add into your instruction filter dialog? At a first glance I don't find a use case for it that is better than simply put these instructions into a references view.

mrexodia:
I think it’s good to have a way to filter instructions directly in the trace view. This could help us implement for example viewing all instructions in the current function. Or some kind of taint analysis and only show the relevant instructions.

The downside of the reference view is that it’s not very dynamic and that if we extend the trace view to show more information about the currently selected instruction we won’t have these features.

My only possible concern is how it deals with memory usage and performance with big traces, but I think it will be fine...

torusrxxx:
Filtering instructions just by mnemonic, as it is in the pull request, is
not very helpful.

mrexodia:
Yes but it is easy to test and we need to start somewhere in my opinion.

torusrxxx:
Isn't other more interesting and easier filtering a better way to start, such as filter by user/system attribute? Basically I don't want to add a feature that almost nobody uses and we have to maintain it.

wrongway88:
Like mrexodia says I think it's a good starting point for more sophisticated filtering (the foundation should remain the same for any kind of filter).
The reason I chose filtering by mnemonic is surely based on my own needs (or the needs of my project, so not just me). I however think it would be usefull to anybody who has to deal with large traces.

Anyways, I was planning on adding more filters already, like the one you propose. I can add that and resubmit if you prefer.

Thanks for the feedback

mrexodia:
I fixed the conflicts with the recent Capstone -> Zydis rename that happened.

ZehMatt:
I would actually like to see this. I just ran into the same situation where I went into "Search" and was looking for "Instruction/Mnemonic" but its not there (yet).

There are some examples I can think off:
1. Exotic instructions, not very common yet easy to find via trace.
2. Syscalls, yea I wanna see where it does one.

torusrxxx:
This feature is useful in some cases and even more useful if it can be combined with other filters. I'm just waiting for one more filter so that other contributors can see how to add a filter.

mrexodia:
I squashed+rebased this, but I didn't have time to test yet.

mrexodia:
Fixed build errors on 32 bit

mrexodia:
Resolved conflicts and fixed some minor bugs

mrexodia:
@torusrxxx I will do some more work on this before merging:

- Refactor the filtering functionality to be inside `TraceFileReader` so it's transparent for the `TraceBrowser`
- Add filtering by module
- More testing

torusrxxx:
Wow, it has been a long time! I think filtering inside TraceFileReader is nice, but considering trace searching also uses TraceFileReader, there needs a way to search in unfiltered trace.

mrexodia:
Yeah, I'm trying to keep this branch rebased on the latest development, but it's getting pretty difficult. In general I didn't use tracing very much myself, but whenever I did there would have been a huge benefit in filtering by module so I think some kind of filtering mechanism has merit.

mrexodia:
Tracking in https://github.com/x64dbg/x64dbg/issues/3033

Issue #2015, opened 2018-08-31
Debugger crashed on command 'yara'


cozil:
Debugger crashed while running command 'yara' with only one argument.

I looked into the source code, and found an issue:

```
    if(IsArgumentsLessThan(argc, 2))
        return false;
    duint addr = 0;
    SELECTIONDATA sel;
    GuiSelectionGet(GUI_DISASSEMBLY, &sel);
    addr = sel.start;

    duint base = 0;
    duint size = 0;
    duint mod = ModBaseFromName(argv[2]);
```

argv[2] is not available while running command 'yara' with only one argument.
Should check if argc > 2 before use it.

mrexodia:
To solve this issue: remove yara from x64dbg.

torusrxxx:
To make it clear: `if(IsArgumentsLessThan(argc, 2))` already checks the number of arguments. The real cause is Yara.

balintf:
> To make it clear: `if(IsArgumentsLessThan(argc, 2))` already checks the number of arguments. The real cause is Yara.

This only checks that the `yara` command was given at least one argument, but the cause of the crash is that x64dbg uses the second argument at argv[2], which is only valid if argc > 2. By adding checks before argv[2] usages and some default values for the argc == 2 case, I was able to get the yara command working with only one argument.

> To solve this issue: remove yara from x64dbg.
Do you mean remove it from the 64 bit version, or remove it entirely from x64dbg? Could you elaborate on why do you think yara should be removed? (I don't use it personally, but some might find it useful)

PR #2061, opened 2018-11-01
DBG: fix yara crash when used with a single argument

balintf:
This PR fixes the crash in issue #2015.
When `cbInstrYara` was called with a single argument (argc==2) it tried to dereference the second argument after the end of the argv array in a few cases to try to find the module that should be used.

This PR restores the intended/documented behavior when used with a single value, that
`base = sel.start`, size is the region size from `MemFindBaseAddr(addr, &size)`, and the module name is resolved with `ModNameFromAddr`.


mrexodia:
Thanks!

PR #2062, opened 2018-11-01
Fixed Symbol::GetList() for symbols without undecorated name

homo-programmatis:
* Such symbols have "" instead of nullptr in SYMBOLINFO.undecoratedSymbol
* This fix greatly improves Snowman's decompiled results, example: comctl32!ListBox_SetCurSelHandler


mrexodia:
I think this is fine because nobody is actually using the ScriptApi, but technically it's a breaking change...

mrexodia:
Thanks for your contribution!

PR #2063, opened 2018-11-01
Fixed incorrect comparator used in sorting xrefs

homo-programmatis:
* With old comparator, items {1, 2} and {2, 1} were "less" then each other. This will cause them to sort randomly.


torusrxxx:
Nice

PR #2064, opened 2018-11-01
Fixed incorrect error message when memory could not be allocated

homo-programmatis:
It was printing pointer to stack instead of size


mrexodia:
Lol, what a colossal fuckup... I could’ve sworn I tested this

PR #2068, opened 2018-11-06
Graph zoom mode feature

raiseman:
Implemented as an option in gui tab. When enabled, graph turns to a free scalable object with ability to zoom in/out on any point by (like IDA).

- Zooming to cursor position by Ctrl+Wheel at normal speed, Ctrl+Shift+Wheel at 2x speed. Maximum zoom level is 100%, minimum depends on graph size.

- At some point text becomes hidden and blocks look like "overview" action has been clicked.

- New actions in context menu (with shortcuts) - zoom 100%; fit to window.

- Context menu has been refactored. It now depends on zoom level and cursor position. This makes interaction more clear and predictable. E.g clicking outside of block (especially at large zoom level) will set breakpoint menu hidden as well as any action that needs text to be visible will also be hidden.

graph2


mrexodia:
I think this is a very welcome change, so definitely a big thank you for implementing it!

I don't have a problem merging this (provided it's tested), but in the longer term I would like to replace the graph view with the abstract graph view from Cutter, which is based on this graph view. https://github.com/radareorg/cutter/blob/master/src/widgets/GraphView.cpp

Could you indicate where you got the icons from? And is there any work left to do?

raiseman:
At first I thought porting Cutter graph (in its original form) might be a good idea. But if you try to compare zooming process Cutter vs IDA, you will definitely choose IDA.
Cutter's graph sticks to border. It simply does not allow you scaling/moving to wherever you want. This is particularly visible on large horizontally/vertically oriented graphs.

So my goal was to implement something similar to IDA. 
And I think I did it, at some point it is even better :)

New mode is fully compatible with old calculations. Changes did not affect block coords, edges etc. 
And for those who might prefer old-style mode - it is an option, so you can always uncheck it.

I tested id some time and had not fount any bugs (as of the last commit)
It would be nice if someone tested it on large graph, say > 10k nodes

Icons from [Fuge](http://p.yusukekamiyamane.com/)

PR #2069, opened 2018-11-07
DBG: Fix mapped area overrun in ReadDebugDirectory

### LINKED ISSUE #2065: 32-bit malware sample causes x64dbg crash when opening
Open 32-bit sample

98ece1ba0e548ffb536fffa3e8ea75a2b9d6d8a27e55408d863c5e3d5414890c

With snapshot_2018-10-31_12-52.zip:

screen shot 2018-11-02 at 3 03 21 pm

balintf:
This PR fixes #2065.
`ReadDebugDirectory` could scan past the end of the mapped PE file if the Debug Directory's size is bogus, e.g. 0xfffff000. This PR adds a check for `debugDir + debugDirSize` not pointing out of the mapped image.
There is also an additional check that detects unsigned integer wraparounds, in this case with size 0xfffff000 the calculation would wrap around to 4096 bytes before debugDir which would be inside the mapped area.

### DIFF
--- a/src/dbg/module.cpp
+++ b/src/dbg/module.cpp
@@ -387,7 +387,11 @@ void ReadDebugDirectory(MODINFO & Info, ULONG_PTR FileMapVA)
                     FALSE,
                     IMAGE_DIRECTORY_ENTRY_DEBUG,
                     &debugDirSize);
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #2083, opened 2018-11-22
Graph reopen fix

raiseman:
It seems it is an old bug.

How to reproduce: open graph (press G) -> stop debugging -> start debugging another exe -> click on graph tab (do not press G) -> old graph is still here


mrexodia:
Thanks!

PR #2089, opened 2018-12-11
turn to vs2015

HuangJian1990:
turn to vs2015


mrexodia:
Your merge request is messed up. We want to switch to 2015, but there is currently no real reason to.

PR #2106, opened 2019-01-21
DBG: more robust validation of PE directory sizes

### LINKED ISSUE #2105: Crash on certain input files
ping @Mattiwatti 

Some time ago I got a [sample](https://github.com/x64dbg/x64dbg/files/2776969/x64dbg-vuln.zip) that will crash when loading it into x64dbg. The reason appears to be that `RtlImageDirectoryEntryToData` returns a garbage pointer. Probably the best solution is to rewrite those functions, or perform the same check as the loader does in order to reject the result.

Mattiwatti:
Ref: #2105 #2065

This is essentially the same as PR #2069, but applied to all PE directories. This fixes #2105 (at least the first attachment, likely both) without a need for `try/except` (though keeping the wrapper added in 7d53b1a doesn't hurt).

### DIFF
--- a/src/dbg/module.cpp
+++ b/src/dbg/module.cpp
@@ -117,6 +117,13 @@ static void ReadExportDirectory(MODINFO & Info, ULONG_PTR FileMapVA)
     // Note that we're loading this file because the debuggee did; that makes it at least somewhat plausible that we will also survive
     for(DWORD i = 0; i < exportDir->NumberOfFunctions; i++)
     {
+        // Check if addressOfFunctions[i] is valid
+        ULONG_PTR target = (ULONG_PTR)addressOfFunctions + i * sizeof(DWORD);
+        if(target > FileMapVA + Info.loadedSize || target < (ULONG_PTR)addressOfFunctions)
+        {
+            continue;
+        }
+
         // It is possible the AddressOfFunctions contain zero RVAs. GetProcAddress for these ordinals returns zero.
         // "The reason for it is to assign a particular ordinal to a function." - NTCore
         if(!addressOfFunctions[i])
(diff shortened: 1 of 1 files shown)


mrexodia:
That malware is actually triggering an out of bounds read by setting the number of exports to be very large. Is there some guarantee on data having to be inside the export directory for those RVAs in there?

Adding checks certainly won’t hurt (as long as that massive casting isn’t triggering some UB), but like you said, probably a good idea to do a generic try/except just in case.

Mattiwatti:
I checked out the malware sample; it is indeed quite insidious and I added the range checks in the second commit to fix it. Note that the sample runs OK without the last check added, but it would still be possible to make a PE that exploits the unchecked `addressOfNames[i]` so I added it as well for good measure.

The number of exports itself should be fine I think: it does trigger a 10MB allocation for the vector but that's it. There should still probably be an upper bound on the number of exports though (and imports), but I don't know what. Time to dig up the good ole' PE spec... always assuming it's in the spec, of course...

`void*` -> `ULONG_PTR` is always well-defined AFAIK since `ULONG_PTR` is defined to be an unsigned integral of pointer size. The overflow check is also well-defined because it uses unsigned overflow, not signed overflow which is indeed UB.

Mattiwatti:
Following up - I checked the PE spec and didn't find any mention of a size restriction for either the import or export directories. Even corkami's files don't have something that test this (surprisingly). So I think that 'assume the worst' applies here, i.e. 2^32.

I'm not too worried about a huge import directory - it would of course be possible to kill x64dbg this way, but in the same way you would also kill the loader meaning the image would never load. For the export directory though, there is something to be said for placing a limit on the number of exports, because the loader doesn't normally parse export directories (unless another file imports from $evil_file). What would be a sensible value do you think?

mrexodia:
Sorry for the delay, been really busy recently...

Issue #2131, opened 2019-03-03
Corrupt patches for modules named in certain languages


dl471:
Debugger Version: 3 Mar 2019
Operating System: Windows 7 Professional 64-bit (6.1)
Translation: English and Japanese

When exporting patches for modules with names in certain encodings it seems the patch length gets thrown off and some bytes aren't written. This seems to happen even when using the translations for those languages. This can be reproduced by naming a module テスト or another word in an afflicted language and trying to export a patch for it.

Japanese example:
![japanese_name](https://user-images.githubusercontent.com/32369415/53695838-7c49d100-3db8-11e9-84fe-47bc6b9b18b1.png)

Hindi example:
![hindi_name](https://user-images.githubusercontent.com/32369415/53695841-823fb200-3db8-11e9-96d6-797337b6f1c7.png)

Looking at the patch code I think the issue is how the length of the patch for each module is decided. The patch text gets converted to UTF-8 but the length used when writing to the file is the one from before the conversion. Changing the code to use the length of the text from after the conversion to UTF-8 seems to alleviate the issue.

![name_after_fix](https://user-images.githubusercontent.com/32369415/53695842-853aa280-3db8-11e9-9202-1654d5a8d4d0.png)

This is what things look like after the change. The patch for テスト is now written normally. I've submitted a pull request with the changes I made to get that result.

PR #2132, opened 2019-03-03
Fix filename bug in patch menu

dl471:
Address #2131


mrexodia:
Yes this fix is good, thanks a lot! Probably there are some similar issues in the code in other places too...

dl471:
It's possible though I haven't encountered any similar issues that I can recall. I'll try keep a look out for them though.

mrexodia:
Thank you!

Issue #2002, opened 2018-08-08
Arrow keys to navigate registers view


torusrxxx:
Currently you cannot navigate registers view with keyboard.

PR #2136, opened 2019-03-07
Register view arrow keys

dl471:
I was looking at #2002 and I took a shot at implementing it. This is what I have so far. It covers movements for the general purpose registers and flags for x32. I figured since the positions of the registers in the register view seem to be hardcoded and unlikely to change it'd be best to just hardcode the arrow movements as well instead of a more complicated solution. Not sure if that's a bit weird though so I was wondering what people thought about it before I proceeded further. If you have no major problems with that and think it's an ok approach I can roll it out for the rest of the registers.

### DIFF
--- a/src/gui/Src/Gui/RegistersView.cpp
+++ b/src/gui/Src/Gui/RegistersView.cpp
@@ -209,45 +209,74 @@ void RegistersView::InitMappings()
         {
             mRegisterMapping.insert(x87r0, "x87r0");
             mRegisterPlaces.insert(x87r0, Register_Position(offset++, 0, 6, 10 * 2));
+            mRegisterRelativePlaces.insert(x87r0, Register_Relative_Position(SS, x87r1));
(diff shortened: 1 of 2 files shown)


dl471:
I added the mappings for the rest of the registers. I also added a function which moves the scrollbar to ensure the selected register is always in view so that arrow key navigation doesn't end up with the selected register being off screen. I can't think of anything else that would be applicable to this feature, so I think it's basically done. I would appreciate if a maintainer could take a look at it.

mrexodia:
Will do, thanks for your contribution!

mrexodia:
I reviewed the code and it looks good (especially considering that the current structure is also a lot of calls to `insert`). I have some minor nitpicks that I will fix myself before merging and I have to actually try it out to see if it works as expected.

Thanks a lot and sorry for the delay...

mrexodia:
Thanks!

Issue #2139, opened 2019-03-07
Wrong .DLL Symbols displayed during debugging


willi-neu9:
Hi,

In newer releases of X64DBG 64 Bits (In Version of July 2018 this bug is not displayed) using the attached .DLL the wrong Symbol names are displayed during debugging. See attached .JPEG files. 

If you look into attached "Before_GetProcessAddress_Call.jpeg" you see that EDX points to the symbol Name "ERR_load_crypto_strings"

If you look into attached file: "After_GetProcessAddress_Call.jpeg" you see that address for "ERR_load_crypto_strings" is loaded into RAX register but "libeay321.ERR.get_error" is wrongly displayed.

You can try youself by unpacking attached archive in any directory

Just run "x64_wrong_symbol.exe". It has an embedded breakpoint to load x64dbg on demmand. See name of symbol which address should be loaded and the displayed name of symbol loaded in RAX after you have debuged over GetProcAddress.

Used Windows Version Windows 7 64 Bits.

Best regards

/Willi/

PS. If you need more information, just let me know!

[x64wrong_symbols.zip](https://github.com/x64dbg/x64dbg/files/2942885/x64wrong_symbols.zip)

mrexodia:
Yeah confirmed, something appears to be wrong with the export table parsing here...

PR #2158, opened 2019-04-27
Fix export name parsing

StarsunYzL:
if (IMAGE_EXPORT_DIRECTORY.AddressOfFunctions[i] == 0), the export name after "i" would be wrong. similar to issue https://github.com/x64dbg/x64dbg/issues/2139. here is the fix.


Mattiwatti:
Can confirm this fixes #2139.

mrexodia:
Thanks a lot for your contribution!

PR #2159, opened 2019-04-28
Fix ThreadGetSuspendCount if the suspend count limit is reached

Mattiwatti:
Due to a bug in the kernel, threads that are at a suspend count of `MAXIMUM_SUSPEND_COUNT` (= 127) will have their previous suspend count value reported as 0 by `NtSuspendThread`, and thus also x64dbg since it uses suspend/resume to get the count.

On Windows >= 8.1 there is a better way to get the suspend count without suspending and resuming the thread, so I've added that as well. If this method fails the old one is used.


mrexodia:
Test comment

mrexodia:
Sorry

mrexodia:
Thanks!

PR #2162, opened 2019-05-03
Fix export name is not parsed correctly

StarsunYzL:
There is another bug that is not fixed in PR https://github.com/x64dbg/x64dbg/pull/2158. 

Info.exports.size() may be less than exportDir->NumberOfFunctions, so the export name is not displayed correctly when addressOfNameOrdinals[i] is greater than or equal to Info.exports.size().

Look into the issue https://github.com/x64dbg/x64dbg/issues/2139, view the last export symbol of libeay321.dll in x64dbg(ordinal 4789), the correct one should be "ossl_safe_getenv".


mrexodia:
Thanks!

PR #2179, opened 2019-05-22
added findasmrex and rexasmfind, and also somethings more

D1W0U:
Please, review it if you have enough time and you can, because I duplicated functions to remove some code, and I believe it can be logically conditional determinated, to reduce code. And also, there's a crash, when using \ in the regexp instead of \ while hardcoding the argv[1] parameter. For example: std::string str(".+18.") instead of ".\+18." (x64dbg crashes). And/or when giving some combination of characters I can not remember.


src/gui/Src/Gui/BreakpointsView.cpp
```diff
@@ -660,6 +662,97 @@ void BreakpointsView::editBreakpointSlot()
...
+            exec(QString("SetHardwareBreakpointCondition %1, \"%2\"").arg(addrText).arg(newBp.breakCondition));
+            exec(QString("SetHardwareBreakpointLog %1, \"%2\"").arg(addrText).arg(newBp.logText));
+            exec(QString("SetHardwareBreakpointLogCondition %1, \"%2\"").arg(addrText).arg(newBp.logCondition));
+            exec(QString("SetHardwareBreakpointCommand %1, \"%2\"").arg(addrText).arg(newBp.commandText));
+            exec(QString("SetHardwareBreakpointCommandCondition %1, \"%2\"").arg(addrText).arg(newBp.commandCondition));
+            exec(QString("ResetHardwareBreakpointHitCount %1, %2").arg(addrText).arg(ToPtrString(newBp.hitCount)));
+            exec(QString("SetHardwareBreakpointFastResume %1, %2").arg(addrText).arg(newBp.fastResume));
+            exec(QString("SetHardwareBreakpointSilent %1, %2").arg(addrText).arg(newBp.silent));
+            exec(QString("SetHardwareBreakpointSingleshoot %1, %2").arg(addrText).arg(newBp.singleshoot));
+        case bp_memory:
```
nooperation: case `bp_hardware` falls through to `bp_memory` due to a lack of a break - I'm not sure about the specifics of setting hardware breakpoints so i'm not sure if this is or is not intended

rmgopaul:
Stop emailing me

mrexodia:
Skill issue

Issue #2107, opened 2019-01-22
feature request: Pushing "Enter" on a empty command prompt should repeat the last command


ianklatzco:
I've searched through existing issues and haven't found anything relevant.

This is the behavior in gdb, lldb, and windbg — seems like it might make sense for it to be in x64dbg as well?

I am relatively inexperienced with windows dev, but this seems like the sort of thing I would be able to add without much effort. Would it be welcome?

mrexodia:
Personally I would not use this feature. Generally you're not using the command line to do things like stepping, since there are hotkeys. However like you said, it would be easy to implement so a pull request with setting wouldn't be rejected. On the other hand, the code related to autocomplete/history of the command bar is pretty hacky already so the implementation shouldn't make that even worse 😄

PR #2184, opened 2019-05-31
Fetch mCmdHistory on an empty mCmdLineEdit

lyflow:
Hi 👋, 

I made some change in accordance with issue #2107 
Please, keep in mind that that I am _not_ used to Qt!


torusrxxx:
I would accept this change without any options, anyway it is better than nothing on ENTER.

mrexodia:
I don't think I agree. When you press enter with an empty textbox I expect nothing to happen, like most other textboxes.

torusrxxx:
Anyway I don't want to see it stalled for possibly another year, I'll add the option and push it forward.

mrexodia:
Thanks, this feature has been integrated with #2429

Issue #2176, opened 2019-05-17
Mapped Shortcuts Do Not Match Keys Pressed


connah0047:
- Version: May 9, 2019
- Windows 7 Professional, SP1, x64
- When customizing shortcuts, some keys pressed do not correspond to the keys mapped

**Steps to Reproduce:**
1) Start x32dbg
2) Click on Options -> Shortcuts
3) Scroll down to the instruction "Debug -> Step Into" and click once on it
4) Click the "Clear" button to clear the default shortcut of "F7"
5) Click in the "Shortcut" box and press the "F7" key on the keyboard in an attempt to set the shortcut back to "F7"
6) Observe that "Shift+F6" appears in both the "Shortcut" box and under the "Shortcut" column instead of the expected "F7". There is no readily apparent way to reset it back to default values or to configure "F7" for use as a shortcut now. "F7" is not the only problem key as shown immediately below.

Other keys that reproduce this problem and the errant shortcuts they map to:
F3 maps as Shift+F2
F4 maps as Shift+F2
F8 maps as Shift+F6
F11 maps as Shift+F10
F12 maps as Shift+F10
Scroll Lock maps as Shift+NumLock
Insert maps as Shift+Enter
Page Up maps as Shift+Down
Page Down maps as Shift+Down
Delete maps as Shift+Enter
Up Arrow maps as Shift+End
Left Arrow maps as Shift+End

In an attempt to resolve this issue myself, I have tried:

1) Searching the  bug tracker for similar issues
2) Clearing ALL existing shortcuts before trying to map a problem shortcut
3) Downloading and running a freshly installed copy without making any changes or customizations apart from trying to map a shortcut key

**_PLEASE NOTE:_** This is my first bug report and I have done my very best to make sure it adheres to specifications and that there was not a pre-existing bug describing my problem. If I have overlooked something, please forgive me and point me in the right direction. Thank you!

kdma:
I am no mantainer but looks like this is the problem https://github.com/x64dbg/x64dbg/blob/development/src/gui/Src/BasicView/ShortcutEdit.cpp#L67-L68
For example with F7
Qt::Key_Backtab | 0x01000002
Qt::Key_F7 |          0x01000036
Since F7 "contains" Backtab (which I never heard of) it replaces that part with shift tab and it goes haywire. I am not sure what is the rationale for that substitution.

connah0047:
[...]

PR #2189, opened 2019-06-15
Removed backtab key substitution

kdma:
Like discussed in the issue [#2176](https://github.com/x64dbg/x64dbg/issues/2176) the Qt keys are not bitflags so we can't do binary logic. Moreover the TAB key changes focus so it shouldn't be even possibile to have it bound to some action if said binding is set through the UI (same reasoning applies to BACKTAB).


mrexodia:
Thank you!

PR #2195, opened 2019-06-26
improved sorting in handles tab

affelwafro:
Noticed that sorting in the handles tab is kinda strange.

I came across a ReadFile WinAPI call and wanted to know which file belongs to the handle ID.
Take a look at this picture (clicked the 'Handle' column to sort by handle id's)
handles

The application itself doesn't do anything wrong (sorting the strings) but I would never expect 1C to be at this position.

The only thing left are the lines 433 and 435
I would like to change line 433 from
`mWindowsTable->setCellContent(i, 4, ToHexString(windows[i].threadId));`
to
`mWindowsTable->setCellContent(i, 4, ToPtrString(windows[i].threadId));`

But I don't know if this would be best practice.
For line 435, numbers should be sorted correctly when interpreted as strings?
`mWindowsTable->setCellContent(i, 4, QString::number(windows[i].threadId));`

### DIFF
--- a/src/gui/Src/Gui/HandlesView.cpp
+++ b/src/gui/Src/Gui/HandlesView.cpp
@@ -20,9 +20,9 @@ HandlesView::HandlesView(QWidget* parent) : QWidget(parent)
     mHandlesTable->setDisassemblyPopupEnabled(false);
     int wCharWidth = mHandlesTable->getCharWidth();
     mHandlesTable->addColumnAt(8 + 16 * wCharWidth, tr("Type"), true);
-    mHandlesTable->addColumnAt(8 + 8 * wCharWidth, tr("Type number"), true);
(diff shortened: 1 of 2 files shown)


mrexodia:
If I recall correctly there should already be a mechanism for this. I will check when I’m home, but basically you can just tell stdtable that the contents are actually a number and it should sort by that.

affelwafro:
Oh that would be nice.
I would really like to close this pull request because I'm not able to compile it on my own machine and I thought it was an easy thing to implement.
But as I see now, HandlesView.cpp is using 'StdSearchListView(this, true, true);', so it can't take 5 arguments. I'm not sure if it would be nice to change this to 'StdTable()'.

mrexodia:
You can check here if you are interested how it was done: https://github.com/x64dbg/x64dbg/commit/bf413de6de26342ee72ac6443ce5dcdf26710924

Basically the columns themselves have a property that specifies how they are sorted, just had to change the api of StdSearchListView a bit to expose that.

Issue #1957, opened 2018-06-06
can be a good feature


iofirag:
There will be a very good and used feature
if you can create one more column in the bottom right rectangle view (=stack)
that will show the hex as ascii readable text,

like at "ollydbg" and "Immunity Debugger" that do the same

for example in ollydbg:
![image](https://user-images.githubusercontent.com/5690885/41036862-20414ee4-699a-11e8-9183-a3c8985eb0d3.png)
(as you can see there is already one more column of "Address with ASCII dump")

Also in Immunity Debugger
![image](https://user-images.githubusercontent.com/5690885/41037140-066cc42a-699b-11e8-818f-4f1cf9895b10.png)
(as you can see there is already one more column of "text dump", i can change it manually by right click in this block and toggle between "show Unicode dump" or "show ASCII dump")

and in yours app (it's missing):
![image](https://user-images.githubusercontent.com/5690885/41036541-0a216064-6999-11e8-9e8b-7f391c767dec.png)
i can see it just if i will mark one or more lines and hit right click -> Binary -> Edit.

it will be easier if i can see it without it like the others.

mrexodia:
Settings -> GUI -> Show ASCII/Unicode in address dump mode

iofirag:
This flag add this column at the left bottom of the application and not in the stack view.

In the stack that was more comfortable to check first

mrexodia:
> Probably the best way to implement this would be to add a (per default hidden) column to the stack view and allow users to view it with the 'Edit columns' option

PR #2203, opened 2019-08-18
GUI: add ascii/unicode column on cpu stack widget

ab22:
Feature for #1957

There's a few small things that could be added here. Currently, no settings are saved when the columns are shown/hidden. I am not sure if this is something I should add.

Also, to properly show the ascii/unicode strings on stack, I had to use `std::reverse`. Since performance is a top priority here, I am not sure if this is the way to go with the strings. Another option would be to implement reverse inserts on the `getColumnRichText()` functions based on parameters to avoid having to call `std::reverse`.

Let me know what you think.

### DIFF
--- a/src/gui/Src/BasicView/HexDump.h
+++ b/src/gui/Src/BasicView/HexDump.h
@@ -73,6 +73,7 @@ public:
     struct ColumnDescriptor
     {
         bool isData = true;
+        bool invertData = false;
         int itemCount = 16;
         int separator = 0;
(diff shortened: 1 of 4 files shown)


mrexodia:
My apologies, I saw this at the time and tried it, but then it completely got lost.

Probably the best way to implement this would be to add a (per default hidden) column to the stack view and then expose the 'Edit columns' via the context menu. The latter will likely be done anyway, so just the former is left.

PR #2204, opened 2019-08-20
GUI: #1999 - Added restart required label for system font change

### LINKED ISSUE #1999: Application Font
Latest version as of this post, on Windows 10 (64bit like anybody uses 32).

I adjusted the font sizes for each of the panes without a problem. They all resized right away. However, when I tried resizing the 'Application Font', nothing happened. until I restarted the application that is. I must have set an overly large value, so when I stated up the app again it was a mess. A larger problem was navigating back to the font window. The window has a static size but the inner controls were too large to fit, causing an overflow, which prevented me from seeing/clicking the "defaults" button. I fixed the issue by shift tabbing.

I now notice an asterix/star after the "Application Font" title. Usually this would mean an accompanying message, however there is none (there is enough space under the buttons to contain one but it's empty). I'm guessing something like "Requires a relaunch" should be written there.

Also, would it be possible to resize the button bar according to the font size. Currently they are just too small. One must be very careful not to click the wrong one.

ahkrichards:
This resolves #1999 per @mrexodia comment to open PR for adding a label.

This does not address @twifty's request to "would it be possible to resize the button bar according to the font size. Currently they are just too small. One must be very careful not to click the wrong one.".
 
Added label to bottom of appearance dialog font tab that explains that "*" denotes setting that requires app restart to take effect.
Label is word wrapped and vertical expanding to help with larger font choices.
Only English locale.

Example:
image

### DIFF
--- a/src/gui/Src/Gui/AppearanceDialog.ui
+++ b/src/gui/Src/Gui/AppearanceDialog.ui
@@ -1711,6 +1711,37 @@ border-width: 1px 2px 2px 1px;</string>
       <string>Log:</string>
      </property>
     </widget>
+    <widget class="QWidget" name="layoutWidget9">
+     <property name="geometry">
+      <rect>
+       <x>10</x>
+       <y>250</y>
+       <width>500</width>
+       <height>34</height>
+      </rect>
+     </property>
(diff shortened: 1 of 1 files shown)


mrexodia:
Sorry for the delay, thanks for your contribution!

PR #2231, opened 2019-10-22
DBG: reject wildcard only patterns

### LINKED ISSUE #2212: Issues with new find-patterntransform - reject invalid pattern code
Hi mrexodia,

I think the reason invalid patters are still accepted in find cause your reject invalid pattern is not working. Would you please check the type of string passed. I think you need a cast from System::string to std::string.

I could be wrong. Please check. You are so talented you will solve it soon.

Thanks,
VV

babywashbear:
closes #2212


mrexodia:
Thanks!

Issue #2164, opened 2019-05-05
Do not query process cookie when attaching


Mattiwatti:
Steps to reproduce:
- Enable process cookie querying.
- Attach to a running process (e.g. another instance of x64dbg) and do some stuff in the debuggee.
- Whenever `NtQueryInformationProcess` is called, a "breakpoint reached not in list!" message is generated.

IMO it would be fine to only call `HandleNtdllLoad` (which is responsible for setting the initial bp on `NtQueryInformationProcess`) if the process was created and not attached to. The cookie is only guaranteed to be queried in `LdrpInitializeProcess` -> `RtlSetUnhandledExceptionFilter` -> `RtlEncodePointer`. All other ways of reaching `RtlEncodePointer`/`RtlDcodePointer` happen only under specific circumstances, like the debuggee adding a VEH, or when ntdll calls the exception handlers, so they are not reliable anyway.

Another thing about `HandleNtdllLoad` is that it could be changed to return without setting a breakpoint if the OS is >= Windows 8, because in that case the cookie can just be queried by the debugger. Both of these changes would reduce conflicts with other code that reads `NtQueryInformationProcess` and finds an unexpected breakpoint byte (e.g. ScyllaHide).

Related: #1747

Edit: I see that this was already mentioned in #1964, but no fix was given other than to disable the cookie setting.

balintf:
I already have a simple patch for this, but I want to explore the possibilities of improving it by querying the process cookie and optionally using hardware breakpoints as suggested in #1747 .

> Another thing about HandleNtdllLoad is that it could be changed to return without setting a breakpoint if the OS is >= Windows 8, because in that case the cookie can just be queried by the debugger.
Do you mean to query it simply by using  `NtQueryInformationProcess` with the debuggee's Process Handle? (I've tested it under Windows 10, and it seems to work well, but I don't have a Windows 7 to verify that it fails and if `NtQueryInformationProcess` returns an error code in this case).

Mattiwatti:
> Do you mean to query it simply by using NtQueryInformationProcess with the debuggee's Process Handle? (I've tested it under Windows 10, and it seems to work well, but I don't have a Windows 7 to verify that it fails and if NtQueryInformationProcess returns an error code in this case).

[...]

PR #2237, opened 2019-10-30
Fix process cookie on attach

balintf:
This PR fixes issue #2164 .
CookieQuery will first try querying the process cookie when creating a process or attaching, using NtQueryInformationProcess, which should succeed on Windows >= 8. If it does not succeed we add the breakpoints to NtQueryInformationProcess only when attaching.


mrexodia:
Great, thanks a lot!

PR #2238, opened 2019-10-30
DBG: fix invalid handle exception on terminating attach after run

balintf:
On terminating an attached process after terminating a created process will result in invalid handle exception. This is due to calling CloseHandle on the previous process's hProcessToken.


mrexodia:
Thanks!

Issue #2169, opened 2019-05-09
Debugger crashes after using "analadv" x32dbg only


partoftheworlD:
Before submitting an issue, make sure it has not already been reported.

Then, make sure the following points are included in your issue:

- Debugger version:  Apr 30 2019.
- Windows 10.0.17763.437 x64
- Debugger crash after using "analadv" only in x32dbg.
- Reproduction steps:
1. Attach to process (for example Steam.exe)
2. After attaching write "analadv" in ntdll.dll module.
3. After completion of the analysis - crash

Additional information can help solve the issue quicker, such as:

![image](https://user-images.githubusercontent.com/6013494/57449334-44518500-7264-11e9-89d4-1417401a555a.png)

[Mini-dump](https://mega.nz/#!AotW2aTA!19-oxicG-nRcwMmkF8DjQdfKb42rQ1Oreqfn-Lfrvws)

balintf:
I could reproduce this problem with the latest 32 bit build. Open any 32-bit exe, for example calc.exe, select ntdll.dll and issue "analadv". I tried it on Win 10.0.18362.418 x64.

The crash is caused by overflowing the mEncMap array in AdvancedAnalysis. The code doesn't properly handle if a function candidate or referred data doesn't end before the end of the memory section.

PR #2239, opened 2019-10-31
DBG: fix advanced analysis crash

balintf:
This PR fixes issue #2169 .

Added range checks to analyzeFunction, so it stops when reaching the end of the memory range regardless whether it is data byte, an incomplete instruction or a complete instruction right at the end of the range.

Added range checks to writeDataXrefs for operands that are bigger than one byte, and simplified handling multi-byte operands.


mrexodia:
Thank you!

PR #2255, opened 2019-11-23
DBG: Fix function boundary computation when getting x64 stack frames

### LINKED ISSUE #2254: x64 call stack: Off-by-one error when determining function boundaries
The function end address in the `RUNTIME_FUNCTION` struct points to one byte beyond the last instruction of the function (like STL ranges).

The current lookup method assumes an inclusive range, resulting in incorrect stack frames on entry into a function "B" that starts immediately after another function "A". In such a case, function A's unwind metadata incorrectly gets used to unwind the stack instead of function B's.

The code in question:
https://github.com/x64dbg/x64dbg/blob/824100eea88c38c5a8c7f84d17f832bf6611e26d/src/dbg/stackinfo.cpp#L156-L178

**What the code currently does:**
Find `func` where:
`func.BeginAddress <= rva <= func.EndAddress`

**What the code should do:**
Find `func` where:
`func.BeginAddress <= rva < func.EndAddress`

**Simple test case:**
```cpp
__declspec(noinline) void fnC() {
    printf("fnC()\n");
    printf("fnC() END\n");
}

__declspec(noinline) void fnB() {
    printf("fnB()\n");
    fnC();
    printf("fnB() END\n");
}

__declspec(noinline) void fnA() {
    printf("fnA()\n");
    fnB();
    printf("fnA() END\n");
}

int main() {
    fnA();
    return 0;
}
```
(When compiled on VS 2017 in x64 Release, `fnC` is 32 bytes long and `fnB` immediately follows it.)

Screenshot of erroneous behavior on entry into `fnB`:
callstack_before

Expected call stack: fnA -> fnB
Observed call stack: `unknown` -> fnB

now-raymond:
Fixes #2254.

The function end address in the `RUNTIME_FUNCTION` struct points to one byte beyond the last instruction of the function (like STL ranges).

The current lookup method results in incorrect stack frames on entry into a function "B" that starts immediately after another function "A", resulting in function A's unwind metadata incorrectly getting used to unwind the stack instead of function B's.

**Existing behavior:**
Find `func` where:
`func.BeginAddress <= rva <= func.EndAddress`

**New behavior:**
Find `func` where:
`func.BeginAddress <= rva < func.EndAddress`

### DIFF
--- a/src/dbg/stackinfo.cpp
+++ b/src/dbg/stackinfo.cpp
@@ -162,15 +162,12 @@ static PVOID CALLBACK StackSymFunctionTableAccess64(HANDLE hProcess, DWORD64 Add
         return nullptr;
 
     DWORD rva = DWORD(AddrBase - info->base);
-    RUNTIME_FUNCTION needle;
-    needle.BeginAddress = rva;
-    needle.EndAddress = rva;
-    needle.UnwindData = 0;
-    auto found = binary_find(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), needle, [](const RUNTIME_FUNCTION & a, const RUNTIME_FUNCTION & b)
+    auto found = std::lower_bound(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), rva, [](const RUNTIME_FUNCTION & a, const DWORD & rva)
     {
-        return a.EndAddress < b.BeginAddress;
+        return a.EndAddress <= rva;
     });
-    if(found != info->runtimeFunctions.end())
+
+    if(found != info->runtimeFunctions.end() && rva >= found->BeginAddress)
         return &found->BeginAddress;
 #endif // _WIN64
 


src/dbg/stackinfo.cpp
```diff
@@ -162,15 +162,12 @@ static PVOID CALLBACK StackSymFunctionTableAccess64(HANDLE hProcess, DWORD64 Add
...
     DWORD rva = DWORD(AddrBase - info->base);
-    RUNTIME_FUNCTION needle;
-    needle.BeginAddress = rva;
-    needle.EndAddress = rva;
-    needle.UnwindData = 0;
-    auto found = binary_find(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), needle, [](const RUNTIME_FUNCTION & a, const RUNTIME_FUNCTION & b)
+    auto found = std::lower_bound(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), rva, [](const RUNTIME_FUNCTION & a, const DWORD & rva)
     {
-        return a.EndAddress < b.BeginAddress;
+        return a.EndAddress <= rva;
```
mrexodia: Thanks for your pull request, but wouldn't it be easier to simply change the population of `info->runtimeFunctions` to put in EndAddress - 1?

mrexodia: The reason I'm commenting here is that https://en.cppreference.com/w/cpp/algorithm/lower_bound says:

> The range [first, last) must be partitioned with respect to the expression element < value or comp(element, value), > i.e., all elements for which the expression is true must precede all elements for which the expression is false. A fully-> sorted range meets this criterion.

And it is very difficult to tell if this is really the case, so I'd prefer sticking to `binary_find`

now-raymond: Thanks for x64dbg!

> wouldn't it be easier to simply change the population of info->runtimeFunctions to put in EndAddress - 1?

Of course, this is possible, but I think it is not such a good idea. `info->runtimeFunctions` is a `vector<RUNTIME_FUNCTION>`. imo, we shouldn't change the meaning of values in system structs like `RUNTIME_FUNCTION` since they may be accessed by others who may have certain expectations of the data. It is also quite standard in CPP code (e.g. STL ranges) to have half-open ranges.

> https://en.cppreference.com/w/cpp/algorithm/lower_bound says: ...

Actually, `binary_find` uses `std::lower_bound`, so `binary_find` has at least the same or more requirements as `std::lower_bound`. `std::binary_search`'s requirements are listed [here](https://en.cppreference.com/w/cpp/algorithm/binary_search) -- it has the same requirement that you've quoted above.

Imo using `binary_find` is slightly misleading here in terms of code-intent since the needle isn't really a RUNTIME_FUNCTION, it's a single address (rva). The existing code creates a stand-in RUNTIME_FUNCTION needle with both start and end addresses set to the RVA to satisfy the comparator predicate's argument requirements:

https://github.com/x64dbg/x64dbg/blob/824100eea88c38c5a8c7f84d17f832bf6611e26d/src/dbg/stackinfo.cpp#L164-L168

The simple comparison predicate `return a.EndAddress < b.BeginAddress` used with `binary_find` works only for the inclusive range.

As mentioned, since `binary_find` uses `std::lower_bound` anyway (`std::lower_bound` is essentially already binary search), we could do less work by using `std::lower_bound` directly, comparing with the RVA directly instead of with a synthesized RUNTIME_FUNCTION from the RVA.

A comparator with `std::lower_bound` has the signature: `[](const RUNTIME_FUNCTION & a, const DWORD & rva)->bool` and returns true if the function range `a` is ordered before (and thus does not include) `rva`.

In the commit I wrote the predicate as `return a.EndAddress <= rva` but perhaps it would have been clearer to write it equivalently as `return (a.EndAddress - 1) < rva`, i.e. `a` < `rva` if the last byte of `a` is before `rva`.

```cpp
DWORD rva = DWORD(AddrBase - info->base);
auto found = std::lower_bound(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), rva, [](const RUNTIME_FUNCTION & a, const DWORD & rva)
{
    return (a.EndAddress - 1) < rva;
});
if(found != info->runtimeFunctions.end() && rva >= found->BeginAddress)
    // found
```

mrexodia: The intention is to do a binary search and I would expect an interface similar to `std::find`, where it either returns an iterator or the end. `std::lower_bound` has this super awkward and (in my opinion) error prone interface where you have to do stuff like `if(found != info->runtimeFunctions.end() && rva >= found->BeginAddress)`, which is why I wrote `binary_find`.

All I want to do is a binary search, but whenever using `std::lower_bound` you have to think about these pointless details that are super easy to mess up (as in: I have to open cppreference every time I try to use it to make sure that's how it works).

I do agree though that creating the fake element isn't ideal, so I'm a bit torn between either solution...

now-raymond: If you would like to stick to `binary_find`, to fix the off-by-one error, the needle can be defined as follows:

```cpp
RUNTIME_FUNCTION needle;
needle.BeginAddress = rva;
needle.EndAddress = rva + 1;
needle.UnwindData = 0;
```

And the predicate changed to:
`return (a.EndAddress - 1) < b.BeginAddress;`

Full code:
```cpp
DWORD rva = DWORD(AddrBase - info->base);
RUNTIME_FUNCTION needle;
needle.BeginAddress = rva;
needle.EndAddress = rva + 1; // changed
needle.UnwindData = 0;
auto found = binary_find(info->runtimeFunctions.begin(), info->runtimeFunctions.end(), needle, [](const RUNTIME_FUNCTION & a, const RUNTIME_FUNCTION & b)
{
    return (a.EndAddress - 1) < b.BeginAddress; // changed
});
```

I know you just want to do a binary search, but since the needle (pointer) and the haystack (function ranges) differ in type, both `binary_find` and `lower_bound` require extra work to handle the search - for `binary_find`, it is in the formulation of the needle - the synthesized RUNTIME_FUNCTION must compare equivalent to the corresponding actual RUNTIME_FUNCTION that it is a part of, i.e. satisfy (!(a<b) && !(b<a)). In that sense, the complexity is hidden in the setup for the needle.

For `lower_bound`, the extra work is ensuring `rva >= found->BeginAddress` since `lower_bound` will return the first element that compares greater or equal to the RVA, i.e. the `rva < func.EndAddress` half of the equation.

Ultimately I suppose it is a manner of preference! If you would like and agree with the `binary_find` change I can update the PR with that change instead.

mrexodia: I don't think either solution is much better than the other, so I will merge it as-is, thanks!

PR #1843, opened 2017-12-21
Add CTRL+1/CTRL+2, which popup a window to follow in disam/dump

lynnux:
if there are more than one choice, just continue CTRL+1/CTRL+2 to loop selection.

PR #2262, opened 2019-12-08
Add CTRL+1/CTRL+2, which popup a window to follow in disam/dump

lynnux:
Previous closed pr is here: [pr](https://github.com/x64dbg/x64dbg/pull/1843)
I use this patch on my own personal fork from 2017 to now, it's really useful, and I don't want to build my fork to add this utillity any more, as the git clone in china become more and more slow...


lynnux:
actually, my keyboard layout is different from others, I switched left ctrl and alt, so alt+1/alt+2 is my key stroke on other keybord, you can assign any other shortcut for this.

lynnux:
Similar things has already implemented in my private plugin, so closed it now

Issue #1908, opened 2018-02-27
Feature: specify export function in DLL loader with arguments


Te-k:
x64dbg does not currently have a feature to debug a specific export function of a DLL (as of 27th of February 2018). So if you want to debug a DLL export function, you have either to use another tool or to update the DLL entry point.

A feature like this exist in OllyDbg : when a DLL is loaded, you can specify export function and arguments in a specific window : 

![img](https://techtalk.pcpitstop.com/wp-content/uploads/OllyDbg_Dll_Loader.png)

The x64dbg team mention on Twitter that the feature does not exist because [no one wants to implement it](https://twitter.com/x64dbg/status/968527006156632064), which is totally fair, people should spend their time on developing feature they need. But I think it is interesting to have an issue opened to document what can be developed by other people and see how interesting it would be for some person.

So here is description of the feature as I see it:

- When a DLL is loaded, have the possibility to call a specific function exported
- Have the possibility to add parameters

Is it a feature other people are interested in? (please add +1) Is there anyone interested to develop it? Any idea on how this feature should be implemented?

GauriSpears:
> you have either to use another tool or to update the DLL entry point
Could you explain the workaround? I am debugging .cpl file - it is dll with exported CplApplet function. When I change entry point RVA to RVA of the function, I've got exceptions.

JusticeRage:
So, I've been looking into this for a couple of hours today, following our brief Twitter exchange.
I would like to ask a few questions before I go further with the implementation.

The first thing I did was add a new setting to control this behavior:
![image](https://user-images.githubusercontent.com/6743798/72207370-c4398a00-3498-11ea-99d1-58c87449f025.png)

After browsing around in the code, I'm unsure what is the best place to implement this feature. I don't really want to change code inside the Titan Engine (which includes the DLLLoader.exe program) ; i feel that this would be better suited to reside in the debugger itself.
My idea is to change the `debugLoopFunction` in `debugger.cpp` to add a callback to the `InitDLLDebugW` call. When the entry point BP is triggered, I intend to change the EIP, push optional arguments on the stack, and so on.

The problem I have currently is that while the callback is called as expected, the debugger doesn't stop but instead keeps running untill the program exits. Is there a way to have a callback executed, while still keeping the breakpoint at EP?
More generally, is this the right approach, or is there a better place where this could be done?

stevemk14ebr:
[...]

PR #2272, opened 2020-01-12
Feature to specify export function in DLL loader with arguments 

JusticeRage:
See https://github.com/x64dbg/x64dbg/issues/1908

Currently a work in progress. The objective is to add a dialog that allows users to select which export they want to run when debugging a DLL.

It's still a work in progress at the moment. The general idea is:
- Add a new setting in Misc to control whether the user wants to see the dialog.
- If so, add a callback to the DLL's entrypoint BP to show the dialog.
- Let the user select the import and (in the future) add parameters.
- Change EIP to the requested DLL, allocate/push arguments and give back control to the user.

In the future, the dialog could also be extended to call arbitrary functions in a program during the debugging session.

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -1735,6 +1735,11 @@ BRIDGE_IMPEXP void GuiGetCurrentGraph(BridgeCFGraphList* graphList)
     _gui_sendmessage(GUI_GET_CURRENT_GRAPH, graphList, nullptr);
 }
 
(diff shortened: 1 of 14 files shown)


mrexodia:
Closing this because it's not complete. Right now you can pretty much mirror how OllyDbg's loader works by replacing `loaddll.exe` with whatever you desire. This feature is nice to have, but it just doesn't seem that relevant to me since in most cases you'll either debug `DllMain` to unpack something or debug the DLL via the application that actually uses it.

Issue #211, opened 2015-04-12
int3 (debug interrupt 'CC') not easily skipped


Nukem9:
**Problem**: After int3 (debug interrupt 'CC') is encountered during debugging, execution stops (this is good), but no simple keystrokes allow execution to continue. To continue, I have found that if I press 'DownArrow' and then Ctrl-*, that will sometimes make the next instruction the current (Set New Origin Here).  More than half the time, this does not work, and the RIP goes back to the int3.  Then, debugging can continue.  Also, clicking on the next line, then right-clicking and selecting Set New Origin Here does the same, and most of the time I have to repeat the process.

**What is wanted**: Immediately after execution stops due to 'int3', be able to press F7 or F8 and continue (as in Ollydbg, similar in other debuggers) by skipping over the 'int3' instruction and setting RIP at the next instruction.

**Background**:  I debug my own code by inserting 'int3' statements wherever I want the debugger to stop so I can debug at that point.  But x64_dbg behaves differently from the other debuggers I also use (Visual Studio, Windbg, and Ollydbg).  With the others, pressing F7/F8 (or F10/F11) causes the program to immediately skip over the 'int3' instruction and resume with the next line.  x64_dbg feels awkward compared to the others.

----------------------------------------
- Bitbucket: https://bitbucket.org/mrexodia/x64_dbg/issue/211
- Originally reported by: [ejamesr](http://bitbucket.org/ejamesr)
- Originally created at: 2014-10-25T22:13:56.221

Nukem9:
DBG: resolved issue #211 (added a 'skip' command to skip the next instruction)


-> <<cset 3b755c927914>>

----------------------------------------
Original comment by: [mrexodia](http://bitbucket.org/mrexodia)

Nukem9:
You are right about that it "feels awkward", but in my opinion x64_dbg is one of the few debuggers handling this thing correctly. Various debuggers handle INT3 commands as if they are breakpoints, meaning they swallow the exception, leaving the debugger open for detection mechanisms like:


```
#!c++

BOOL TestDebuggerSwallowsException()
{
    __try
    {
        RaiseException(STATUS_BREAKPOINT, 0, 0, 0);
    }
    __except(1)
    {
        return FALSE; //all good, debugger did not swallow the exception
    }
    return TRUE; //debugger detected
}
```

Above code inspired from: http://www.autosectools.com/anti-debugging-with-exceptions.pdf

[...]

PR #2283, opened 2020-01-24
Enable int3 skipping swallow exception

stevemk14ebr:
This relates to my comment on https://github.com/x64dbg/x64dbg/issues/1908 as well as https://github.com/x64dbg/x64dbg/issues/211. The functionality implemented by the patch for 211 makes it so that Step Over/Step Into (swallow exception/pass exception) no longer function as expected. Instead i have removed this option, and made it so that all pass/swallow exception step modes now handle int3 just as any other exception would be in these modes AND normal stepping modes do not automatically swallow int3. This make the features behave as expected while still protecting x64dbg from the int3 anti-debug tricks.


stevemk14ebr:
bump

mrexodia:
This had been fixed in a different way, but thanks for the PR :)

PR #2295, opened 2020-02-18
more useful data in trace browser 

sage444:
Hello 

Added registers and memory columns to Trace tab. I done this just for my personal use since context menu with the same data not always handy.

What is done: 
- visualisation of updated registers and values 
- visualisation of memory and updated values 
- highlight 
- copy 
- might broke something, don't know yet:)

Will appreciate any feedback to this PR.


mrexodia:
Sorry for the extreme delay, I totally missed this PR. I pushed a fix for the code formatting and will review. But it looks good on first glance!

sage444:
Hey, it's ok, no need to sorry.
Late is better than never;)

mrexodia:
Thank you!

PR #2321, opened 2020-04-13
Replace x32 with x86

paulens12:
Obviously x32 was a typo and someone just went on with it...


mrfearless:
Its to represent 32bit and 64bit usage - and if you add 32 and 64 together you get 96 which is the name of the helper app x96dbg.exe, so it is by design that it uses the existing naming convention.

mrexodia:
It was not a typo, just the project was called x64dbg so it made sense to call it x32dbg for 32 bits. I’m aware of the x32 ABI but at this point it’s too late.

paulens12:
First of all, it's not at all apparent that 96 is gotten by adding those two together. I just thought it was a typo in x86.

Secondly, the 32bit variant of x86 is NOT called x32, that is a completely incorrect term coined by kids who don't know the difference between "x64" and "64 bits" and hence incorrectly translate "32 bits" to "x32". This makes the project look extremely unprofessional and cringe-inducing.

The correct name would be either plain x86 or IA-32, or i386. Similarly, the 64bit variant could be called AMD64 or x86_64, all of which are correct terms. As for the x96 thing, it could simply be called x86 if you decide to go with IA_32 or i386 as it represents the whole family of architectures.

mrexodia:
>  I’m aware of the x32 ABI

PR #2346, opened 2020-05-20
take advantages of function-boundary-information from runtimeFunctions

TanakaYasen:
ModInfo::runtimeFunctions can help to identify boundary of a function. (non-leaf function)

### DIFF
--- a/src/dbg/function.cpp
+++ b/src/dbg/function.cpp
@@ -83,8 +83,32 @@ bool FunctionAdd(duint Start, duint End, bool Manual, duint InstructionCount)
 bool FunctionGet(duint Address, duint* Start, duint* End, duint* InstrCount)
 {
     FUNCTIONSINFO function;
-    if(!functions.Get(Functions::VaKey(Address, Address), function))
+    if (!functions.Get(Functions::VaKey(Address, Address), function))
+    {
+#ifdef _WIN64
+        auto info = ModInfoFromAddr(Address);
+        if (nullptr == info)
+            return false;
+
+        DWORD rva = DWORD(Address - info->base);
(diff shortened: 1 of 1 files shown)


src/dbg/function.cpp
```diff
@@ -83,8 +83,32 @@ bool FunctionAdd(duint Start, duint End, bool Manual, duint InstructionCount)
 bool FunctionGet(duint Address, duint* Start, duint* End, duint* InstrCount)
 {
     FUNCTIONSINFO function;
-    if(!functions.Get(Functions::VaKey(Address, Address), function))
+    if (!functions.Get(Functions::VaKey(Address, Address), function))
+    {
+#ifdef _WIN64
```
mrexodia, line 88: I like this idea! However this source file is responsible for loading/saving/getting functions from the database, so this isn't the right location for this change.

Would be nice if you can move it here https://github.com/x64dbg/x64dbg/blob/063ce31ec6c264fda5b2a22e592874e44b38cbe9/src/dbg/_exports.cpp#L261 and also name it `rtfunc_xxx` (or something)

TanakaYasen: I found some defects and limitations of above implementation. E.g. leaf node functions & x32 PEs are not supported. 
um, I find a fantasy function of IDA, which is Function Window and function list. I wonder how it implemented. I tried to erase runtimeFunction(a.k.a Exception Directory) of a PE file, but IDA can recognize functions still. Then I tried to patch a function such as "sub_180001000" filling with int3, and then IDA failed to recognize the function and not list "sub_180001000" in function list window. It seems that IDA recognize function boundary basing on analyzing byte codes. Do you have research on IDA implementation??? I think this function is charming but unnoticed by people.

mrexodia:
Running the command `exanalyse` already does the same.

PR #2351, opened 2020-05-26
GUI: Sanitizing input expression before assemble.

jackreez:
BUG: When assembling an instruction, if we copy and paste an instruction from somewhere with whitspaces [ \r\n] it results in invalid parsing in asmjit engine selected and throws an error as "Error writing to memory".

SCENARIO: Copy and pacste an instruction from a text editor with a newline prefix to the assemble dialog and try to assemble.
If there is newline in the prefix the validation don't even care about the expression after the newline prefix.

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -934,18 +934,21 @@ void CPUDisassembly::assembleSlot()
             if(exec != QDialog::Accepted)
                 return;
 
+            //sanitize the expression (just simplifying it by removing excess whitespaces)
+            auto expression = assembleDialog.editText.simplified();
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -934,18 +934,21 @@ void CPUDisassembly::assembleSlot()
             if(exec != QDialog::Accepted)
                 return;
 
+            //sanitize the expression (just simplifying it by removing excess whitespaces)
+            auto expression = assembleDialog.editText.simplified();
```
jackreez, line 938: variable to reduce extra function calls.

mrexodia:
Thanks!

PR #2359, opened 2020-06-05
Styling improvements

JustasMasiulis:
The main parts of styling that were removed:
* Tables headers
* Status bar
* CPU info scroll bars

Still a bit of work left to do but this fixes the main offenders.

Included there is a `style.css` file that more or less recreates the original look.


justanotheranonymoususer:
Maybe this caused the regression? https://github.com/x64dbg/x64dbg/issues/2402

PR #2362, opened 2020-06-08
Upgrade Zydis

torusrxxx:


mrexodia:
Tracking in https://github.com/x64dbg/x64dbg/issues/3030 instead

PR #2367, opened 2020-06-13
DBG: Fixed variables entries contains invalid garbage names.

jackreez:
[FutureProofFix]: Force copy of variables from the variables list. Sometimes auto copy used in the for loop results in reference to the local stack (i guess and there may be some more like this, noticeable in latest compiler builds). Occured in build, compiled with msvc 19.16 (VS2017) not the official (VS2013) builds.

{D3A54040-4BE4-42E7-8D9B-69E4AFE41627}

### DIFF
--- a/src/dbg/variable.cpp
+++ b/src/dbg/variable.cpp
@@ -386,10 +386,9 @@ bool varenum(VAR* List, size_t* Size)
     }
 
     // Fill out all list entries
-    for(auto & itr : variables)
+    for(const auto &entryIter : variables)
     {
-        *List = itr.second;
-        List++;
+        *(List++) = VAR(entryIter.second);
     }
 
     return true;


jackreez:
I'm think this is not the exact fix for this issue. I wonder why some of the variable names only pointing to a garbage string!

Need to re-investigate that

jackreez:
Anyway no problems with the official release compiler. So, it's not an issue until release compiler changes.

mrexodia:
Generally this code is real 💩. Likely something is going wrong with undefined behavior and the compiler is optimizing that. Can you reproduce this in debug mode on vs2019?

jackreez:
I checked this with VS2017 and it happens in debug mode as well.

jackreez:
After some low level digging i've found that there is a bug in the STL string's **operator=** overloaded method.

TL;DR

This happens only for small string length less than 16 (size of the local buffer of the string structure).

```cpp
// Memory Structure of an STL String Object
struct String {
    size_t padding;  // Only appears in debug builds. just for safety. i guess.
    union _data {
        char *ptr;
        char buffer[16];
    } data;
    size_t size;
    size_t capacity;
};
```
**String memory structure from a debug build of x32dbg.** 
mem_ref

Even the initial capacity (0xF: local char array buffer to save mem allocs for small strings < 16[last byte is reserved for the null terminator if strlen is 15]) is not overflown the cstr is mem allocated and the ptr is pointing to the cstr. Since the capacity is still 0xF and and the buffer has a ptr and not a cstr the value of the string is invalid until the capacity is incremented. In this case, further calls to **c_str()** method will return a ptr to the buffer and not the actual ptr that pointing to the cstr.

It seems the string allocator allocates a new blk of memory of capacity and copies the data in there even the size is < capacity.

This happens when we call **String::operator=()** before the string ctor **String::String()**.

How my fix for the issue works is.

The original implementation.
```cpp
for (auto & itr : variables)  // where typeof variables: map<String, VAR>, List: VAR*
{
    *List = itr.second;
    List++;
}
``` 
What this does is it directly calls **VAR::operator=** on List with itr.second which
internally calls the **String::operator=** on its members.
```cpp
for (auto & itr: variables)
{
    VAR::operator=(List, itr.second);
    List++;
}

My Fix
```cpp
for (auto & itr : variables)
{
    *List = VAR(itr.second);
    List++;
}
``` 

What this does is it firstly calls the ctor on List and the calls the **VAR::operator=**
```cpp
for (auto & itr: variables)
{
    VAR::VAR(List);
    VAR::operator=(List, itr.second);
    List++;
}
```

If this was the case with **operator=** then all assigned short string will have garbage value.

I don't know the underlying details of the string allocator when operates on an non initialized string.
But, It seems to be an undefined behavior when called this way.

TODO: // To come with a better soln for the bug. 
1) Check with different Tools version. (Currently: v141_xp).
2) Check with different version of compiler.

PR #2369, opened 2020-06-14
DBG: Fixed DLL breakpoints.

jackreez:
DLL breakpoints can now be enabled/disabled and removed!


jackreez:
Have some problems with git.
made another PR #2375.

PR #2372, opened 2020-06-18
Programatic way to show "References" tab

morsisko:
These changes allow to show the "References" tab either via plugin or using new command `showref`


mrexodia:
Looks good, thanks a lot!

PR #2374, opened 2020-06-25
DBG: Added memory breakpoint's types in BpToBridge conversion.

jackreez:
Fixes: Memory breakpoints summary always shows type as access for all types.
@BreakPointView.


mrexodia:
Thank you!

PR #2380, opened 2020-06-27
fix "Trace already active" after "ticnd ())"

torusrxxx:
fix "Trace already active" when start tracing after "ticnd ())"
"ticnd []]" is now invalid


mrexodia:
Thanks!

Issue #1794, opened 2017-11-03
References tab not following the address on double-click


reliasn:
[Video showing the problem](https://sendvid.com/0dq4rmj8)

Steps to reproduce it:
1 - CTRL+B to search for a binary pattern. In the example of the video, ``F3 A4`` was used.
2 - Results are shown in the "References" tab.
3 - Double-click one of the results.
4 - x32dbg doesn't follow the address in the "CPU" tab. If you right-click and select "Follow disassembler", it works.

Unless I'm using the debugger incorrectly, I suppose this is not intended.

mrexodia:
It probably follow the reference in the dump window.

mrexodia:
Just watched the video and that is definitely what happened. This keeps popping up so I think it is a good idea to somehow attract the user's attention to the fact that the stuff was followed in dump instead of the disassembler...

mrfearless:
Not sure if its possible with the qt widgets, but maybe flash the border of the dump, alternate colors red-black-red-black for 1 second if follow dump=true, or the background of the dump tab itself (red background, white text / normal background, normal text)

mrexodia:
Currently the border of the dump is flashed 😄 flashing the background is a good idea though!

torusrxxx:
Create a short scrolling animation

torusrxxx:
Please open a separate issue

PR #2387, opened 2020-07-08
Flashing background instead of border when following reference in dump

morsisko:
This pull request implements idea covered in #1794 issue. Now it is easier (at least in my opinion) to focus on the dump widget, instead of the disassembly widget which doesn't change when following some addresses


mrexodia:
Looks great!

Issue #2332, opened 2020-05-07
[FEATURE REQUEST] Jump to address/jump into function with middle mouse click


enkeyz:
This is a feature request, for adding functionality for the middle mouse button. 

If you click on a command with middle mouse button, and that command contains a jump or a function call, then x64dbg jumps to that address/or jumps into the function.

Yeah, I know you can do it with the Enter button on the keyboard, but you always use the mouse, and the middle mouse button isn't used for anything.

enkeyz:
And what about CTRL/SHIFT/ALT + middle mouse?

mrexodia:
I honestly think it would be better to allow changing the behavior of double click, because I don't think anyone ever needs to assemble when doing double click...

enkeyz:
Ty!

PR #2398, opened 2020-07-22
Changed default double-click behavior to follow jumps and calls in disassembly and stack widgets

morsisko:
Implemented #2332 - Now you can select option that can change double mouse click behavior - after double-click jump/call in the disassembly window the jump/call destination will be followed, the same action will be taken when double-clicked the "return to xxx" in the stack widget.

Not sure if my implementation is the best way to check if address on stack is return address

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -103,14 +103,10 @@ void CPUDisassembly::mouseDoubleClickEvent(QMouseEvent* event)
     // (Disassembly) Assemble dialog
     case 2:
     {
-        duint followJumpsAndCalls = 0;
         duint dest = DbgGetBranchDestination(rvaToVa(getInitialSelection()));
 
-        if(BridgeSettingGetUint("Gui", "FollowJumpsAndCalls", &followJumpsAndCalls) && followJumpsAndCalls && DbgMemIsValidReadPtr(dest))
+        if(DbgMemIsValidReadPtr(dest))
             gotoAddress(dest);
-
-        else
-            assembleSlot();
     }
     break;
 
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -102,8 +102,17 @@ void CPUDisassembly::mouseDoubleClickEvent(QMouseEvent* event)
 
     // (Disassembly) Assemble dialog
     case 2:
-        assembleSlot();
-        break;
+    {
+        duint followJumpsAndCalls = 0;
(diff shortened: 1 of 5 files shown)


src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -102,8 +102,17 @@ void CPUDisassembly::mouseDoubleClickEvent(QMouseEvent* event)
...
-        break;
+    {
+        duint followJumpsAndCalls = 0;
+        duint dest = DbgGetBranchDestination(rvaToVa(getInitialSelection()));
+
+        if(BridgeSettingGetUint("Gui", "FollowJumpsAndCalls", &followJumpsAndCalls) && followJumpsAndCalls && DbgMemIsValidReadPtr(dest))
+            gotoAddress(dest);
+
+        else
+            assembleSlot();
```
mrexodia, line 113: For this I'd be happy to change the default behavior to `gotoAddress` and just get rid of `assembleSlot` altogether.

src/gui/Src/Gui/CPUStack.cpp
```diff
@@ -532,7 +532,14 @@ void CPUStack::mouseDoubleClickEvent(QMouseEvent* event)
 
     default:
```
mrexodia, line 533: I think you should add `case 1: // value` and then call `modifySlot()` as it is now, that way none of the behavior disappears

src/gui/Src/Gui/SettingsDialog.cpp
```diff
@@ -250,6 +250,7 @@ void SettingsDialog::LoadSettings()
     GetSettingBool("Gui", "ShowExitConfirmation", &settings.guiShowExitConfirmation);
     GetSettingBool("Gui", "DisableAutoComplete", &settings.guiDisableAutoComplete);
     GetSettingBool("Gui", "AsciiAddressDumpMode", &settings.guiAsciiAddressDumpMode);
+    GetSettingBool("Gui", "FollowJumpsAndCalls", &settings.guiFollowJumpsAndCalls);
```
mrexodia, line 253: Just get rid of the setting, it's better to have double click be intuitive and you are not removing functionality since you can still do the actions by other means...

morsisko:
To clarify, by "get rid of the setting" you mean to remove the possibility to select the default double-click behavior for user, and just make the "follow jump/call" default one?

And for the changes in `CPUDisassembly.cpp` - When the user double-click non jump/call instruction like "mov" or "cmp" nothing should happen?

mrexodia:
Yes, just change the default behavior:)

Yeah I think if you double click and there is nothing to follow doing
nothing is fine

morsisko:
I think it should be better now. I wasn't sure what should happen when the third column in stack widget is double clicked when there is no return address, so nothing happens now (similar to the disassembly widget)

PR #2400, opened 2020-07-23
Follow source operand

torusrxxx:
When I used x64dbg I found this incorrect, it should follow source operand. Most often destination operand doesn't have anything meaningful.


mrexodia:
What about `FXRSTOR`?

torusrxxx:
It just follows whatever can be followed. I just tested MOV, FXRSTOR, NOP, ADD, and PUSHAD and they behave as expected.

mrexodia:
Right, I misread the code 🤦

mrexodia:
Thanks!

PR #2403, opened 2020-07-25
Toggle between graph and CPU with G

justanotheranonymoususer:


mrexodia:
Unfortunately this isn't really enough, because hotkeys cannot conflict. Probably a better solution is to get rid of one of the actions so there is only a single hotkey bound.

mrexodia:
I'll do this by merging the actions, but thanks for the pull request :)

torusrxxx:
I wanted to add a default shortcut for `ActionFollowDisasm` (in `CPUDump` and `TraceBrowser`), but should this `ActionGraphFollowDisassembler` `Shift+Return` merged with `ActionGraph` `G`, what hotkey will be `ActionFollowDisasm` ?

PR #2404, opened 2020-07-25
Set Copy Address to Ctrl+X

justanotheranonymoususer:
Alt+insert is quite clunky, and it's a common operation


mrexodia:
I don't really want to make this change. I'm personally comfortable with Alt+Insert and I think users can bind a different hotkey if they really use it that commonly. Additionally `Ctrl+X` has intuitive behavior of cutting which this doesn't do.

PR #2406, opened 2020-07-26
Updated string searching algorithm to support string in any language

torusrxxx:
String searching is always a headache for most x64dbg users. Most users need to find a plugin or special version for their language, and almost none of them are free and open source. I try to fix that by improving the algorithm to work with all people at the cost of some more useless strings being found, which is acceptable for me.
The algorithm used here is very simple: it converts to and from Unicode, and checks whether the original string is recovered after round-trip conversion without errors. It works for me, and users in other languages are encouraged to test it.

### DIFF
--- a/src/dbg/disasm_helper.cpp
+++ b/src/dbg/disasm_helper.cpp
@@ -280,10 +280,36 @@ void disasmget(duint addr, DISASM_INSTR* instr, bool getregs)
         memset(instr, 0, sizeof(DISASM_INSTR)); // Buffer overflow
 }
 
+bool isunicodestring(const WString & data)
+{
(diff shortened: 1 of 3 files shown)


mrexodia:
Thanks!

mrexodia:
People have been reporting issues with this new functionality, so I will revert it until I find out what causes the issues (because in principle I think these changes are useful).

emckee2006:
Could those differences be caused by looking at up to 60 bytes compared to
the 11 bytes of the previous version? You also removed the memcpy into test
(but I'm not sure what it's purpose is here)...

mrexodia:
image

image

Possible, I didn't look into it further yet.

Issue #2397, opened 2020-07-22
Alt+T doesn't open Threads view


justanotheranonymoususer:
That's because it opens the Trace view in the menu - hotkey collision

PR #2420, opened 2020-08-07
Highlight user modules in modules view & memory map

torusrxxx:
- Highlight user modules in modules view & memory map
- Add example text in color settings
- Fixed bugs with Alt+G and GuiFocusView
- Restore Alt+T hotkey(#2397)


src/gui/Src/Gui/AppearanceDialog.ui
```diff
@@ -629,9 +629,23 @@ border-width: 1px 2px 2px 1px;</string>
       </property>
      </widget>
     </widget>
+    <widget class="QLabel" name="exampleText">
```
mrexodia, line 632: I'm not sure if it's really useful to add this. I tried it but the real example is already updated live in the disassembly view and often the label can be misleading/confusing (for example with a transparent background or when there is no background color at all).

However if adding this is really necessary for you I think it should be improved by centering the text and using the same font as the disassembly, or at least a monospace font to get a better match.

src/gui/Src/Gui/DisassemblerGraphView.cpp
```diff
@@ -1129,6 +1128,9 @@ void DisassemblerGraphView::mouseMoveEvent(QMouseEvent* event)
 
 void DisassemblerGraphView::mouseReleaseEvent(QMouseEvent* event)
 {
+    // Bring the user back to disassembly if the user is stuck in an empty graph view (Alt+G)
+    if((!this->ready || !DbgIsDebugging()) && event->button() == Qt::LeftButton)
```
mrexodia, line 1132: Perhaps add `|| button() == Qt::BackButton` (and maybe escape?)

src/gui/Src/Gui/MemoryMapView.cpp
```diff
@@ -315,6 +310,12 @@ QString MemoryMapView::paintContent(QPainter* painter, dsint rowBase, int rowOff
             painter->drawText(QRect(x + 4, y, getColumnWidth(col) - 4, getRowHeight()), Qt::AlignVCenter | Qt::AlignLeft, wStr);
             return QString();
         }
+        else if(DbgFunctions()->ModGetParty(getCellUserdata(rowBase + rowOffset, 0)) == 0)
```
mrexodia, line 313: I think maybe we should add some enum in `bridgemain.h` to not keep hardcoding the constants all around

src/gui/Src/Utils/Configuration.cpp
```diff
@@ -239,6 +239,8 @@ Configuration::Configuration() : QObject(), noMoreMsgbox(false)
     defaultColors.insert("BreakpointSummaryKeywordColor", QColor("#8B671F"));
     defaultColors.insert("BreakpointSummaryStringColor", QColor("#008000"));
     defaultColors.insert("PatchRelocatedByteHighlightColor", QColor("#0000DD"));
+    defaultColors.insert("SymbolUserTextColor", QColor("#FF0000"));
+    defaultColors.insert("SymbolSystemTextColor", QColor("#000000"));
```
mrexodia, line 243: The dark theme needs to be updated to have at least neutral colors for this.

src/gui/Src/Gui/AppearanceDialog.ui
torusrxxx: Hi, this example is used to configure some color not visible in disassembly, for example "SymbolUserTextColor"

mrexodia:
I made user and system have the same color, but fixed the color in the memory map sorry for the wait!

PR #2429, opened 2020-08-15
Fetch mCmdHistory on an empty mCmdLineEdit and able to disable it

torusrxxx:
This improves #2184 by adding a setting to disable the behaviour


src/gui/Src/Gui/MainWindow.cpp
```diff
@@ -2267,8 +2267,8 @@ void MainWindow::on_actionDefaultTheme_triggered()
     Config()->Colors = Config()->defaultColors;
     Config()->writeColors();
     // Reset [Fonts] to default
-    Config()->Fonts = Config()->defaultFonts;
```
mrexodia, line 2270: Why this change and why is it combined with the history change?

mrexodia:
Thanks!

PR #2430, opened 2020-08-16
New system for expression functions

xenocidewiki:
Modified the way expression functions are handled, and implemented a few new functions that will be useful when setting conditional breakpoints among other things:

- strstr i.e. `strstr(utf8(addr), "abc")`
- strcmp i.e. `strcmp(utf8(addr), "abc")`

Where both can use utf8() or utf16() to specify what sort of string format is being read from memory.

Example plugin: https://github.com/xenocidewiki/x64dbgplugintest


mrexodia:
Thanks man!

Issue #2422, opened 2020-08-07
Switching between Default and Dark Theme reset your font preferences


blaquee:
Any way to persist the font settings? 
Use case: Some folks use lighter theme during daylight hours and switch to dark in the evenings for eye strain.

Can theme settings be exported?

mrexodia:
Yes, the font is part of the style.ini

blaquee:
When you set the fonts it doesnt seem to be saving them in the themes style.ini. Id imagine that switching back and forth would retain the fonts you set.

blaquee:
I see now, User font settings is saved in x64dbg.ini, but when switching styles it will override the font with the styles.ini in the theme folder. I think most applications choose user settings as the default override. We may consider doing this with regards to font settings.

mrexodia:
In theory if you load the theme and then change the font it will respect
your user settings.

PR #2431, opened 2020-08-17
Don't override user font when using dark theme (#2422 #2426)

torusrxxx:
It doesn't need to set font when switching to dark theme, and the font used in dark theme shouldn't override the user setting.


mrexodia:
Thanks!

PR #2439, opened 2020-08-30
Fixed memory leak; properly clean QThread after it finishes

morsisko:
It seems that at the moment the FlickerThread and BackgroundFlickerThread objects are never deleted (despite fact that they are allocated on the heap using "new" operator)


mrexodia:
I think they are deleted once the whole system is down (because of the QObject hierarchy), but probably better to delete them indeed 👍 

I’ll do a bunch of merging once I’m done with my vacation so it might be a while

mrexodia:
Thanks!

PR #2448, opened 2020-09-17
Fix hardware breakpoint replacement slot display

ishitatsuyuki:
Replacement candidates used to show "Slot X (Free)" before sometimes when there are disabled hardware breakpoints. Fix it and change the placeholder to "Unknown" since it's logically invalid to display the menu when there's a free slot.


ishitatsuyuki:
Downloaded the build artifact and briefly tested, seems to be working.

ishitatsuyuki:
Review ping!

mrexodia:
It’s on my list of stuff to do :)

mrexodia:
I still don't completely understand how to reproduce the initial issue, but the fix looks fine so I'll merge it. Thanks!

ishitatsuyuki:
It's easy to encounter the case where the addresses displayed in the slots are wrong, but to actually see the "Free" case it takes a little bit of effort. Anyway, here's a repro video with notepad.exe as the target:

https://www.loom.com/share/422064f861f0416fad2161525ea9bf4d

PR #2456, opened 2020-09-19
Fix analysis blocking the command queue

ZehMatt:


mrexodia:
Tracking this in https://github.com/x64dbg/x64dbg/issues/3031. It needs a more generalized approach to allow commands to be executed synchronously when desired.

PR #2462, opened 2020-09-26
Add python on edit data

viniciusfeitosa:
I did this PR based in a suggestion from this video https://www.youtube.com/watch?v=we-BGBYIUZo


mrexodia:
Isn’t that the same as one of the C-style strings with a “b” prepended?

viniciusfeitosa:
It is like this one DataCShellcodeString

mrexodia:
Thanks!

PR #2465, opened 2020-09-27
Toggle "Show Suspected Call Stack Frame" with space

justanotheranonymoususer:
Not sure if gonna work, tried according to this:
https://github.com/x64dbg/x64dbg/blob/58825d4522f06986e09ab17b141204a66e68dd15/src/gui/Src/Utils/ActionHelpers.h#L240

Let's see what the build says.

### DIFF
--- a/src/gui/Src/Gui/CallStackView.cpp
+++ b/src/gui/Src/Gui/CallStackView.cpp
@@ -37,7 +37,9 @@ void CallStackView::setupContextMenu()
     mFollowFrom->setShortcut(QKeySequence("enter"));
     connect(this, SIGNAL(enterPressedSignal()), this, SLOT(followFrom()));
     mMenuBuilder->addSeparator();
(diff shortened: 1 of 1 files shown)


mrexodia:
I tested and it's not working. Likely an issue that needs some `keyReleaseEvent` to override the behavior manually.

justanotheranonymoususer:
I don't know how to do that unfortunately. I thought it would be easy. Every time something changes, like loading symbols, I need to refrsh this. The only way now is right clicking and changing view, and then back again. So space could make it simpler. But I failed to implement.

mrexodia:
Hm perhaps it’s better to call the refresh view at the right time? Could
you list all the actions that are not refreshing the table?

justanotheranonymoususer:
Loading symbols is one, most common for me. If thread is paused I think it refreshes. But if it's running and I want to force refresh, there needs to be a way to trigger refresh.

mrexodia:
I think the core problem was fixed. Tracking the hotkey in https://github.com/x64dbg/x64dbg/issues/3034

Issue #2438, opened 2020-08-26
CTRL+C doesn't work in search list


torusrxxx:
Currently CTRL+C in search list is broken, it works in other lists.

mrexodia:
Of course!

G3ph4z:
Just to confirm the corresponding PR I've tested it, and works fine, so I believe its fixed.

PR #2474, opened 2020-10-11
Fix Ctrl+C in the search list

aikuimail:
bug description：https://github.com/x64dbg/x64dbg/issues/2438


torusrxxx:
Thanks for your help, however I don't think this approach is the proper fix, because the shortcut is configurable in the shortcuts dialog, and other copy actions support a shortcut too if a shortcut for them are configured. The best way to fix it would be `addAction`.

mrexodia:
Closing this since the solution is too specific. The original bug is still open and might need some clarification.

Issue #2449, opened 2020-09-17
Problem with "String references" search.


dhanzek:
On newer snapshots I encounter problem with string search (Search for -> Current module -> String references),
which manifest as not finding all strings in module.

Example:
As of "snapshot_2020-07-29_04-15" the X32dbg yield no result for searching string "This feature"
in "VB Decompiler.exe" (VB Decompiler Lite v11.5, https://www.vb-decompiler.org/download.htm).
Last good snapshot which can find string was "2020-07-25_13-18".

torusrxxx:
The string that is not found is longer than current max value (60 bytes) and therefore not found. I'll try to fix it.

PR #2482, opened 2020-10-19
Fix string detection

torusrxxx:
There was a bug when the string is long. This has been fixed and it can find 20,000+ strings in sample mentioned in #2449 including the desired string.


mrexodia:
Thanks!

Issue #2308, opened 2020-03-16
Editing watches brings up empty text box


gyorokpeter:
When right-clicking on a watch and "Edit..." I get a blank text box instead of having the current watch expression filled in. I can work around by doing "Copy -> Expression" but this is inconvenient.
![kép](https://user-images.githubusercontent.com/1097079/76752649-14261d00-6778-11ea-8646-a19208768ce8.png)

PR #2484, opened 2020-10-19
GUI: Improved editing expressions in watch view #2308

morsisko:
Now the input box contains current watch expression, instead of showing empty input box.


mrexodia:
Thanks!

PR #2485, opened 2020-10-20
Add separate event settings for system module;Add debug engine selection

torusrxxx:


src/gui/Src/Gui/SettingsDialog.h
```diff
@@ -109,6 +116,12 @@ private slots:
         calc_unsigned = 1
     };
 
+    enum EngineType
```
mrexodia, line 119: There is already an enum for this in the bridgemain.h and locally I have a very similar change but I will see which one is better and manually merge :)

mrexodia:
I made some changes (such as the NtTerminateProcess breakpoint not being the default), but otherwise it works great!

Issue #2309, opened 2020-03-16
Watches: can't set/change type from window


gyorokpeter:
Adding or editing a watch brings up a window with only one text box for the value, and no place to set the type.
![watch](https://user-images.githubusercontent.com/1097079/76753414-9d3d5400-6778-11ea-96df-2882e3ac0ddc.png)

PR #2486, opened 2020-10-21
GUI: Possibility to change watch type using GUI (context menu)

morsisko:
-Added new command `SetWatchType`, not sure if it is necessary but I think it's more clean this way
-Added possibility to change watch type from context menu (lack of possibility to change the watch type from gui was reported in #2309)

x32dbg_2020-10-22_01-56-49


mrexodia:
Thanks a lot!

PR #2487, opened 2020-10-23
Improved call stack tab, displaying multiple threads 

morsisko:
-Added new function (GetCallStackByThread)
-Change the call stack view, it displays call stacks from multiple threads now

Hope it looks better now. Not sure if the internals of that function looks good for you.

Before: 
x32dbg_2020-10-23_17-10-33

After:
x32dbg_2020-10-23_15-39-43


lynnux:
This is useful, good job, thanks!

morsisko:
Changed the thread number to thread id, moved the current thread's callstack to the top of the table, and highlighted the current thread, now it looks like this:

PR #2490, opened 2020-10-26
In some cases, the call stack information for a 64-bit process cannot be obtained correctly.

wenchen1995:
In some cases, the call stack information for a 64-bit process cannot be obtained correctly.
reference: https://docs.microsoft.com/en-us/windows/win32/api/dbghelp/ns-dbghelp-stackframe

AddrFrame

An ADDRESS64 structure that specifies the frame pointer.

x86:  The frame pointer is EBP.

Intel Itanium:  There is no frame pointer, but AddrBStore is used.

x64:  The frame pointer is RBP or RDI. This value is not always used.


mrexodia:
Awesome, thanks a lot!

PR #2497, opened 2020-10-30
Changes related to WatchView

morsisko:
-Added column enum to WatchView
-Added current type to `SetWatchExpression` executed from `editWatchSlot`, this is because otherwise if no type is specified this resets the watch type back to `uint`
-Added "force refresh" feature. Without this change, after you edit watch expression, the value of this particular watch is always "0" until you refresh it by for example stepping to next instruction.


mrexodia:
Thanks!

PR #2505, opened 2020-11-03
Fix flag setting when flag is already set

ioncodes:
This fixes commands such as `_ZF=1` where ZF is already set to 1 by checking whether the bit is already set.


mrexodia:
Awesome, thanks! ❤️

PR #2509, opened 2020-11-08
Added in Log windows values for addresses of functions after module w…

IevgenKulykDev:
**List of function found during analysis of debugging module**

After user pressed Ctrl-A combination, in the Log it's possible to see not only the statistics but functions' addresses. It's possible to click on that address for fast moving to that area and analyse functions that were found.


IevgenKulykDev:
View->Functions (Ctrl + Alt + F) will create this list. PR is ambigous

PR #2526, opened 2020-11-28
Fix appearance, assemble and close dialog when using ultra large font

torusrxxx:
#2524
x64dbg-appearance-ultra-large-font
x64dbg-interface-ultra-large-font


mrexodia:
Thanks!

PR #2527, opened 2020-11-28
auto follow operand in stack, load data from file

torusrxxx:
* Auto follow operand in stack
* Load data from file (TODO: What did ODbgScript do? Should the "savedata" command be renamed to "DM", same as ODbgScript?)


mrexodia:
Thanks! I disabled the option to be enabled per default after discussing with a few people because they say it could be confusing.

However I think a better implementation would be to underline the relevant data and do that unconditionally. I am working locally on this, but it might take a while 🙂

PR #2541, opened 2020-12-10
Allow animation interval greater than 100ms

torusrxxx:
Allow animation interval greater than 100ms
Don't remove Xref when pressing "A"


src/dbg/analysis/recursiveanalysis.cpp
```diff
@@ -51,7 +51,7 @@ void RecursiveAnalysis::SetMarkers()
         {
             FunctionDelRange(start, end, false /* Do not override user-defined functions */);
             LoopDeleteRange(start, end); // clear loop range in function
-            XrefDelRange(start, end); // clear xrefs in function
+            //XrefDelRange(start, end); // clear xrefs in function
```
justanotheranonymoususer, line 54: Maybe it's better to remove only xrefs for which both the source and destination are within the range. The real problem is that xref with source outside the range is removed.

PR #2555, opened 2020-12-25
Fix transparent favourite toolbar hovering over the File menu

torusrxxx:
The favourite toolbar is now placed in main toolbar, not at (0,0) when first used.


mrexodia:
Thanks!

PR #2563, opened 2020-12-29
Return RUNTIMEFUNCTION from DbgFunctionGet for better GUI

torusrxxx:
Now almost every function is clearly bounded in the disassembly view.

### DIFF
--- a/src/dbg/function.cpp
+++ b/src/dbg/function.cpp
@@ -90,7 +90,24 @@ bool FunctionGet(duint Address, duint* Start, duint* End, duint* InstrCount, dui
 {
     FUNCTIONSINFO function;
     if(!functions.Get(Functions::VaKey(Address, Address), function))
+    {
+#if _WIN64
+        duint base = ModBaseFromAddr(Address);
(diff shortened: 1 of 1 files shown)


mrexodia:
I don't really want to merge this change without a setting because it's confusing and gives a false impression that analysis has been done. In addition there is no guarantee this data is valid at all and even in valid executables the ranges can be split up.

torusrxxx:
So what about a new analysis called "runtimefunctionanalysis" that simply copies RUNTIMEFUNCTION to x64dbg functions?

mrexodia:
I’m pretty sure this command already exists :)

torusrxxx:
So let's add `exanalyse` to the analysis menu? Definitely better than other analyse.

mrexodia:
I think in some ways it's worse. But if `exanalyse` is extended to merge multiple `RUNTIME_FUNCTION` entries that are part of the same function I think it's a really good idea!

mrexodia:
I added a follow-up issue, but I will close this one.

PR #2567, opened 2020-12-30
Create/Open system breakpoint script

torusrxxx:
The system breakpoint script can be created/opened from the system breakpoint script browse dialog.


mrexodia:
Thanks!

PR #2573, opened 2021-01-05
Go to function return in trace view

torusrxxx:


torusrxxx:
This fixed a bug which caused singleshoot hardware breakpoint not removed from TitanEngine, and then you cannot use a singleshoot hardware breakpoint more than 4 times.

PR #2587, opened 2021-01-30
Use PointerToRawData in debug directory if no RVA is present

### LINKED ISSUE #2574: x32dbg unable to load a PDB 2.0 file
When loading TRAOD.exe in x32dbg it's as if it doesn't even see the adjacent TRAOD.pdb file:

alt text

Loaded the same exe file in both x32dbg and OllyDbg 1.10, showing the same location in both:

alt text

You can see there are no symbols shown in x32dbg, while OllyDbg does show them.

I've checked that the .exe and .pdb match with ChkMatch tool:

alt text

Tried as well with the latest snapshot (snapshot_2021-01-08_02-43.zip) and the same result. So this is something that probably hasn't been reported yet.

Link to files for testing: https://download1077.mediafire.com/1s7h7tcek2ng/i8ltb9wqnujoo61/TRAOD.zip.

BR,
Sun

Mattiwatti:
This fixes #2574.

### DIFF
--- a/src/dbg/module.cpp
+++ b/src/dbg/module.cpp
@@ -463,7 +463,11 @@ static void ReadDebugDirectory(MODINFO & Info, ULONG_PTR FileMapVA)
     const auto supported = [&Info](PIMAGE_DEBUG_DIRECTORY entry)
     {
         // Check for valid RVA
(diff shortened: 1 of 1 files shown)


mrexodia:
❤️

PR #2592, opened 2021-02-04
Change NtTerminateProcess breakpoint to a generic 'exit breakpoint'

Mattiwatti:
I like the `NtTerminateProcess` breakpoint that can be enabled in the settings, but it has some issues compared to what I'm used to from WinDbg. Consider the following program:
```C++
int main()
{
    return TerminateThread(GetCurrentThread(), 0);
}
```
This will bypass the `NtTerminateProcess` breakpoint, since terminating the last thread of a process will also terminate the process (but without invoking `NtTerminateProcess` directly.) A probably more useful example is some other process (task manager or maybe something more malicious) terminating the debuggee process. In cases such as this it is currently not possible to do a proper post mortem to see what caused the debuggee to exit, because the entire GUI state is wiped.

This PR renames the `NtTerminateProcess` breakpoint to simply 'Exit Breakpoint', and moves the implementation to the exit process callback. This means the process state (current EIP/RIP, call stack etc.) can still be inspected regardless of how the process was terminated.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1489,9 +1489,23 @@ static void cbCreateProcess(CREATE_PROCESS_DEBUG_INFO* CreateProcessInfo)
 static void cbExitProcess(EXIT_PROCESS_DEBUG_INFO* ExitProcess)
 {
     dprintf(QT_TRANSLATE_NOOP("DBG", "Process stopped with exit code 0x%X\n"), ExitProcess->dwExitCode);
+    const bool breakHere = settingboolget("Events", "NtTerminateProcess");
+    if(breakHere)
+    {
+        // lock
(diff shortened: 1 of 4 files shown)


mrexodia:
Initially I did it this way, but if I remember correctly things get really weird at this stage of the process (all DLLs unloaded etc). What do you think @torusrxxx ?

Mattiwatti:
This doesn't seem to be the case for me: running the AppVeyor binary from this PR and the sample code above, x64dbg behaves basically as if there was a breakpoint set on the return of `NtTerminateThread` (or some other syscall if the process was killed externally). DLLs are still loaded and visible in the symbols tab, memory map and the dump view. You can even still dump the process.

torusrxxx:
As you may know, I'm having holiday and doesn't have time to try and comment on this.

mrexodia:
I didn’t know, but I hope you’ll enjoy it!

mrexodia:
❤️

PR #2606, opened 2021-03-07
Allow Option for assembling Instruction on Double Click

Yakov5776:


Yakov5776:
Oops, didn't add every commit

PR #2610, opened 2021-03-09
Add copy menu actions to trace register view

torusrxxx:
You can copy registers from the trace view


mrexodia:
Thanks! Hope you’re doing well 😊

PR #2619, opened 2021-03-28
Fix typo in TraceBrowser.cpp

eltociear:
occured -> occurred


mrexodia:
Thank you! Perhaps it would be a good idea to run some kind of source code spell checker over the whole codebase and fix all the issues at once

Issue #2598, opened 2021-02-15
A few suggestions of UI


czghzc:
I am a person who pursues details and perfection. I will give you some more demanding suggestions. Please understand.

1,When the program is opened, the initial interface size and position are displayed first, and then it jumps to the last memory size and position.
Suggestion: hide or skip the initial interface and jump directly to the last memory position and size.
![gif1](https://user-images.githubusercontent.com/23645596/107971769-db2ca300-6fed-11eb-9c49-5678f5f57777.gif)

2,Hope that the register window can be adjusted smaller. It is useful when the entire interface is small. But it is limited.
![gif2](https://user-images.githubusercontent.com/23645596/107971828-ef70a000-6fed-11eb-8c09-a3473b447e32.gif)

3,The distance between the interval marking line and the sideline of these two places is different. The distance between the second place is too small and it feels a bit crowded. It would be great if it can be the same distance as the first place.
![gif3](https://user-images.githubusercontent.com/23645596/107971946-14651300-6fee-11eb-9226-0f2a042ab0ad.png)

4,The scroll bar type marked 2 is different from other scroll bar types. It is recommended to use the scroll bar type marked 1 uniformly. And this register window cannot be adjusted to the minimum, similar to the second question.
![gif4](https://user-images.githubusercontent.com/23645596/107971969-1f1fa800-6fee-11eb-8f44-c6906fbf5061.png)

5,The font size in these places are too small to see.
![gif5](https://user-images.githubusercontent.com/23645596/107971989-26df4c80-6fee-11eb-9de6-5313a201f675.png)

mrexodia:
Thanks for the suggestions! I'm pretty sure that 1/2/4/5 are quite easy to fix, 3 is some kind of limitation in Qt that I'm not sure can be bypassed. I would suggest you take a look and try to fix some of these issues yourself if you have some time. Let me know if you need any help getting things to compile, but I will gladly merge improvements like this!

czghzc:
Sorry, I am not too familiar with QT programming, I may not be able to contribute.

PR #2620, opened 2021-03-28
Allow registers to be resized smaller & use default scrollbar style

torusrxxx:
#2598
x64dbg


mrexodia:
This is awesome, thanks!

PR #2623, opened 2021-03-31
enhance the FPU support for trace info box and trace registers view

torusrxxx:
- The content of FPU registers now show in the trace info box as numbers, easing the reverse engineering of floating point code.
- Double click on EIP in trace registers will follow in disassembly, double click on XMM register in trace registers will open the View XMM registers dialog.
- FPU operands in memory will now display in the trace info box.
- Moved Thread ID from "Information" menu to trace info box
- Increase output precision for FPU register in info box
- Hide scrollbars in trace info box


mrexodia:
Thanks!

PR #2634, opened 2021-04-10
Add module name to import symbols

torusrxxx:


mrexodia:
Thanks!

lynnux:
Thanks @torusrxxx ,but the sort seems still by name without dll.
图片
Is this need to change?

PR #2640, opened 2021-04-14
add support module+offset

bashen946:
Follow in Disassembler: ntdll.dll+100

Double click the address
CPU:
ntdll.dll+100 | 0000 | add byte ptr ds:[eax],al |
ntdll.dll+102 | 0000 | add byte ptr ds:[eax],al |
ntdll.dll+104 | E0 00 | loopne ntdll.77BE0106 |
ntdll.dll+106 | 0221 | add ah,byte ptr ds:[ecx] |
ntdll.dll+108 | 0B01 | or eax,dword ptr ds:[ecx] |

### DIFF
--- a/src/gui/Src/BasicView/Disassembly.cpp
+++ b/src/gui/Src/BasicView/Disassembly.cpp
@@ -6,11 +6,13 @@
 #include "CachedFontMetrics.h"
 #include "QBeaEngine.h"
 #include "MemoryPage.h"
+#include "public.h"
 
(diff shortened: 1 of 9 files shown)


mrexodia:
Hello and thanks for your contribution! Is this different from #2639, because one of them should be closed so I can start reviewing the other one :)

mrexodia:
The syntax that achieves the same is either `ntdll:$100` or `ntdll:0 + 100`

PR #2645, opened 2021-04-17
GUI: Disabled possibilty to sort content in call stack tab, fixed #2637

### LINKED ISSUE #2637: Call StackFunction module data display is abnormal
Test version ：snapshot_2021-04-13_22-19(Windows server2019 17763 x64)
image
Exception demonstration ：

As shown in the figure above, when you click on the list name, the data will be displayed abnormally 
Steps to reproduce ：
1. Use x64dbg to load a program 
2. Click the "call stack" function module (the function module needs to have data at this time) 
3. Click on a list name (such as "Thread" in the dynamic picture demonstration)

morsisko:
I think that in future it could be a good idea to implement sorting via Thread ID, but for now I think disabling is ok, to avoid the possibility of breaking the widget.

### DIFF
--- a/src/gui/Src/Gui/CallStackView.cpp
+++ b/src/gui/Src/Gui/CallStackView.cpp
@@ -6,8 +6,8 @@ CallStackView::CallStackView(StdTable* parent) : StdTable(parent)
 {
     int charwidth = getCharWidth();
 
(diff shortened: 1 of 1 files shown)


mrexodia:
Awesome, thank you!

PR #2678, opened 2021-07-04
Update SettingsDialog.ui

CaledoniaProject:
Use the same title "Preferences" as in the main window menu


mrexodia:
Awesome, thank you!

Issue #2665, opened 2021-05-25
Bug：Error in change memory data windows


zmrbak:
![QQ图片20210525201141](https://user-images.githubusercontent.com/7257543/119495882-78d60200-bd95-11eb-83ae-0bbb8f939e53.png)
![QQ图片20210525200308](https://user-images.githubusercontent.com/7257543/119495070-948cd880-bd94-11eb-9ead-b40ffe723966.png)
![QQ图片20210525200438](https://user-images.githubusercontent.com/7257543/119495078-96569c00-bd94-11eb-90cc-44e1de4f707a.png)

PR #2690, opened 2021-07-18
MemoyDump - "modify" dialog fix

morsisko:
In the current version of x64dbg the value is always treated as duint, thus for example when modifying the value in 1 byte mode, the signed/unsigned values are converted to duint/int as in the #2665. Also in the current version, the "Signed" field input value will never be less than zero when double clicking the value in dump in signed word/byte mode.
This pull request changes this behavior.


mrexodia:
Thanks a lot!

Issue #2317, opened 2020-04-01
Locking window filter and refreshing view: Filter is removed


AntonVonDelta:
I can't lock the filter in the Handles>Windows filter and refresh the view because the filter will be removed.

**Before:**
![image](https://user-images.githubusercontent.com/25268629/78144726-a6951480-7438-11ea-99d1-961e26500701.png)

**After:**
![image](https://user-images.githubusercontent.com/25268629/78144765-b3196d00-7438-11ea-9b06-e6dbb9314d06.png)

- Version:  Feb 23 2020 on x32dbg
- Windows 10 x64 1909

Is there a quick fix I could implement myself? I need this feature.

mrexodia:
Probably this won’t be so difficult to fix. Check the code responsible for filling the table (iirc HandlesView.cpp).

PR #2693, opened 2021-07-19
Don't clear filter on list refresh (#2317)+ fix icon table text overflow

torusrxxx:


src/gui/Src/BasicView/StdIconSearchListView.cpp
```diff
@@ -18,6 +18,6 @@ void StdIconSearchListView::setIconColumn(int c)
 
 void StdIconSearchListView::setRowIcon(int r, const QIcon & icon)
 {
-    clearFilter();
+    //clearFilter();
```
mrexodia, line 21: Wouldn’t this mess up the results because the filter isn’t updated and the search not repeated?

torusrxxx: Sometimes (for example, AttachDialog) when the table is refreshed there is a call to refreshSearchList to repeat the search. Sometimes there isn't (//NOTE: DO NOT CALL mModuleList->refreshSearchList() IT WILL DEGRADE PERFORMANCE!).
If there is a filter, you don't need to call reloadData() at all since that will be invisible, if we keep the filter properly instead of clear it. Maybe StdSearchListView::reloadData() should be modified to do what refreshSearchList() does?
As for other methods, such as setRowCount , setCellContent and setRowIcon I have no idea why the filter must be cleared as the filter will be cleared or retained when reloadData() is called.

mrexodia:
Thx

Issue #2654, opened 2021-04-29
Comment Crash in 32/64-bit


bitRAKE:
Using the latest nightly build, April 17, 2021.
(Error also in prior version.)

Crash is easy to reproduce:
1. debug in 32/64-bit
2. view comments
3. right-click on non-code comment and view in disassembler.
4. {crash}

Since not all comments produce the error, I've attached a very small example.

The display gets filled with this message box until crash:
```
---------------------------
Error
---------------------------
Invalid parameter passed to CRT function! Program will now generate an exception.

File: ???
Function: ???
Expression: ???
---------------------------
OK   
---------------------------
```
Happy to provide a 64-bit example - it errors the same.
[template.zip](https://github.com/x64dbg/x64dbg/files/6397393/template.zip)

mrexodia:
Thanks for the reproduction case, really appreciated!

morsisko:
Is this already fixed? Can't reproduce it on the build from 1st July

bitRAKE:
AFAICT, works beautifully - some happy fasm coders.
https://board.flatassembler.net/topic.php?t=20753

bitRAKE:
> Is this already fixed? Can't reproduce it on the build from 1st July

Try to right-click on 0x403000, View in Disassembler in example.
Still crashes.

morsisko:
Strange, I'm running your database with your .exe on x32dbg using 64 bit version of windows 7, everything seems to work fine 
![2021-07-26_22-20-59](https://user-images.githubusercontent.com/11231925/127054115-d0a6589f-a77f-4f8c-acb0-23884ef587dc.gif)

bitRAKE:
On Win10 all mouse movement accelerates the cascade to crash. So, maybe it's related to mouse capture in the event loop? Yet, "Follow in Dump" works -- there would need to be some difference in how "Follow in Disassembler" is handed that causes the crash.

bitRAKE:
I was also able to produce the crash from the Dump. So, it's probably not specifically comment related. Right-clicking on addresses in the '.idata' section [from anywhere] and "Follow in Disassembler" is sufficient to crash.

morsisko:
Found the issue. This is caused by rendering a comment that is longer than 512 characters (the maximum length of comment). Should work fine after my pull request.

PR #2700, opened 2021-07-28
DBG: Fixed crash when comment longer than 512 characters

morsisko:
Fixed crash mentioned in #2654


mrexodia:
Awesome, thanks a lot!

Issue #2515, opened 2020-11-16
More actions in memory map


torusrxxx:
- [ ] Load from file
- [ ] Follow in modules (or symbols) window
- [ ] Bookmark

PR #2701, opened 2021-07-31
Added "Follow in Symbols" to the Memory Map context menu

morsisko:
-Added new command "symfollow" to follow base address in the symbols tab (not sure if this is good idea)
-Implemented feature suggested in #2515 - added "Follow in symbols" to Memory Map context menu


mrexodia:
Missed this somehow, but thanks a lot!

PR #2713, opened 2021-08-17
Added CB_STOPPINGDEBUG plugin callback in order to let plugins perform clean-up

thejanit0r:
Added CB_STOPPINGDEBUG plugin callback in order to let plugins perform clean-up


src/dbg/_plugins.h
```diff
@@ -256,6 +261,7 @@ typedef enum
 {
     CB_INITDEBUG, //PLUG_CB_INITDEBUG
     CB_STOPDEBUG, //PLUG_CB_STOPDEBUG
+    CB_STOPPINGDEBUG, //PLUG_CB_STOPDEBUG
```
mrexodia, line 264: You cannot add the enum here because it will break binary compatibility

mrexodia (changes requested):

src/dbg/_plugins.h
thejanit0r: Correct, forgot about that. Moved the enum to the bottom, before `CB_LAST`.

mrexodia (approved):

mrexodia:
Awesome, thank you!

PR #2714, opened 2021-08-17
Adding support for MEMORY PAGE_NOACCESS breakpoints

thejanit0r:
Adding support for MEMORY PAGE_NOACCESS breakpoints.

The rest will have to be implemented in TitanEngine (PR will follow).


mrexodia:
What is the difference between `access`?

thejanit0r:
> What is the difference between `access`?

`access` is based on `PAGE_GUARD`, which in some cases on multi-core, multi-threaded applications its handling will cause race conditions when used on code pages. With `PAGE_NOACCESS` the permissions on the page remain static.

mrexodia:
I would say this is an implementation detail of TitanEngine (I think
GleeBug implements it with PAGE_NOACCESS) and for the user there is no
difference.

Perhaps the implementation can be improved though (eg use noaccess when an
executable page is breakpointed).

PR #2717, opened 2021-09-02
Added optional flag to enable the alternative memory breakpoint (PAGE_NOACCESS)

thejanit0r:
Currently the flag is only settable from x64dbg.ini / x32dbg.ini

[Engine]
MembpAlt=1

If it proves to be worth the change, it may also be worth it to add it directly to the GUI in the future, or replace it altogether.


mrexodia:
Thanks!

PR #2722, opened 2021-09-06
Don't delete 'unknown exceptions' filter in settings dialog

### LINKED ISSUE #2708: I set the Ignore range in Exceptions , but the break on\Logging\Exception handled by; always reset my setting every time when i restart the software
- Debugger version (can be found on right hand side of menu bar of debugger).
-     version Jul 1 2021.
- Operating system version and Service Pack (including 32 or 64 bits).
-     32 and 64 bits.
- Brief description of the issue.
-     I set the Ignore range in the Exception, but every time I restart the software, the Break on\Logging\Exception handled by; always resets my settings.
-  Elaborate reproduction steps for the bug/issue being reported.
-      execute the software
-      set Ignore range in Exception, and select Do not break, then save the setting.
-      close the software and restart.
-      when I go back to the settings the Break On resume First chance
 
Any issues that do not include these will be put on halt until clarification.

If you have questions, be sure to check:
- http://faq.x64dbg.com
- http://help.x64dbg.com/en/latest/introduction/index.html
- http://x64dbg.com/#contact

Additional information can help solve the issue quicker, such as:

- A POC or sample executable that causes the issue.
- A short clip or a screenshot clearly showing the issue.
- Any supporting information that could help solve/understand the issue.
- The translation you are using. Make sure to also test without translation.
- Plugins you have installed. Make sure to also test without plugins.

Examples of good issues:

- https://github.com/x64dbg/x64dbg/issues/1302
- https://github.com/x64dbg/x64dbg/issues/1407
- https://github.com/x64dbg/x64dbg/issues/1448

Thank you for your interest in improving this community project!

x64dbg team

Mattiwatti:
What happens: the 'unknown exceptions' filter. which is a special filter that is not supposed to be deletable in settings, can be accidentally deleted. This was reported to me by someone who added an exception filter for `00000000-FFFFFFFF`, but adding any filter that includes 0 in its range will cause this. The result is a broken settings dialog that will stay broken even if x64dbg is restarted, because the invalid list will be persisted in the INI.

Cause of above is an off-by-one error; this PR fixes it.

Related (maybe, the issue is not overly descriptive but it seems to match): closes #2708

### DIFF
--- a/src/gui/Src/Gui/SettingsDialog.cpp
+++ b/src/gui/Src/Gui/SettingsDialog.cpp
@@ -536,7 +536,7 @@ void SettingsDialog::AddExceptionFilterToList(ExceptionFilter filter)
     unsigned long start = filter.range.start;
     unsigned long end = filter.range.end;
 
-    for(int i = settings.exceptionFilters->size() - 1; i > -1; i--)
+    for(int i = settings.exceptionFilters->size() - 1; i > 0; i--)
     {
         unsigned long curStart = settings.exceptionFilters->at(i).range.start;
         unsigned long curEnd = settings.exceptionFilters->at(i).range.end;


mrexodia:
Awesome, thanks!

Issue #2510, opened 2020-11-11
How can I print the comment from specific address using command line?


jyoonab:
By using command line, I wish to print the comment(I assume this is correct term) from a specific address,
but I have had no chance find the correct command for doing this.

![tempsnip](https://user-images.githubusercontent.com/37427419/98810285-5700d780-2462-11eb-8052-f402ee6388ef.png)

I have tried to use some commands(such as commentlist etc...) but not able to find the solution.

May I ask if there is any correct command line which will show the comment?

Thanks

mrexodia:
These are so-called autocomments which are based on the registers and
memory. If you want to find the strings you can just use the search strings
button.

jyoonab:
Thanks for prompt reply and letting me know the term = ] 
I am trying to print the autocomments by using the log function like (log "{autocomments@address}").
Is there any way I can do this?
I know I can print the disassembly info by using disasm@dis.sel().

tamersaeed:
Write echo

mrfearless:
There is also a manual process - you can also press ";" (semi-colon) to edit the comment - then select it and copy to clipboard

mrexodia:
You can write a plugin that does this for you. Here is an example of `{timestamp@0}`: https://github.com/mrexodia/TimeStampFormat/blob/master/TimeStampFormat/plugin.cpp

takhyun12:
The purpose of the question is to ask if he can get value through String Formating Method (e.g. {mem;size@address}).

mrexodia:
Im well aware, that’s why I gave an example plugin that allows you to
create your own format functions. There you can call DbgGetCommentAt to get
the comment you want (or return any value).

jyoonab:
@mrexodia Thanks for letting me know. It seems I need to dig into the DbgGetCommentAt function first, and do what you suggested. Let me close this thread = ]

mrexodia:
https://github.com/x64dbg/x64dbg/blob/d422af7dd49074094bce6b92d1590bd9cb14793f/src/gui/Src/Gui/CPUDisassembly.cpp#L796-L803

PR #2740, opened 2021-09-28
DBG: add comment and label format functions

yunzheng:
This adds `comment` and `label` to format functions. This can be handy to log automatic comments or labels such as symbol names or comments that are automatically identified by x64dbg.

Example:

```
log "{comment@eip}"
log "{label@eax}"
```

inspired by #2510


mrexodia:
Thanks a lot!

PR #2744, opened 2021-10-01
set default log and command conditions

torusrxxx:
If there is only log/command text but no condition, then logCondition or cmdCondition would not be NULL and the default value provided in `return logCondition ? logCondition->Evaluate() : 0;` would not be used.


mrexodia:
Thanks for the fix!

PR #2760, opened 2021-10-14
Fixed a bug that overwrites last character of DBCS string with NUL

torusrxxx:
The bug corrupts some DBCS strings and thus x64dbg will not able to find them. With the patch x64dbg will find more valid strings.


mrexodia:
In practice you are mostly seeing garbage blocks and weird sequences, likely because it’s extended ascii and almost any sequence of bytes is a “valid” string (or at least decodes)

PR #2764, opened 2021-10-16
Secretly allow the user to mod background image

torusrxxx:
Do you want to try x64dbg with background IMAGE instead of color? Set background color in your INI to #XXXXXX and add following to style.css

```
Disassembly,
TraceBrowser,
QPlainTextEdit,
StdTable {
    background-repeat: no-repeat;
    background-image: url(foobar.jpg);
    background-attachment: fixed;
}
```

### DIFF
--- a/src/gui/Src/BasicView/AbstractTableView.cpp
+++ b/src/gui/Src/BasicView/AbstractTableView.cpp
@@ -245,7 +245,8 @@ void AbstractTableView::paintEvent(QPaintEvent* event)
     }
 
     // Paints background
-    wPainter.fillRect(wPainter.viewport(), QBrush(mBackgroundColor));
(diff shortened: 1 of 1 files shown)


mrexodia:
Just as a super-delayed follow-up for people searching, you can use `background-position` to move the image to the right:

```css
Disassembly {
    background-repeat: no-repeat;
    background-image: url(./rc/image.png);
    background-position: bottom right;
    background-attachment: fixed;
}
```

image

Another [trick](https://forum.qt.io/topic/40151/solved-scaled-background-image-using-stylesheet/3) is to use `border-image`:

```css
border-image: url(:/res/background.jpg) 0 0 0 0 stretch stretch;
```

The [Qt documentation](https://doc.qt.io/qt-6/stylesheet-customizing.html) is pretty good about explaining the process of drawing the CSS background.

PR #2778, opened 2021-10-28
dump TEB on double clicking FS/GS register

torusrxxx:


mrexodia:
That's a nice quality of life improvement, thanks!

PR #2783, opened 2021-10-30
README.md: HTTP => HTTPS

Schweinepriester:
Checked every link, skipping the redirect HTTP => HTTPS this way :)

Two HTTP-only links remain:
- http://translate.x64dbg.com (HTTPS cert is for crowdin.com)
- http://plugins.x64dbg.com (redirect only works with HTTP)


mrexodia:
Thanks!

PR #2790, opened 2021-11-19
Fix the enter key on the keypad not working in CPUStack widget

starsunyzl:
The built-in "enter" shortcut of the CPUStack widget only works with Qt::Key_Return, and doesn't work with Qt::Key_Enter, both should work.

### DIFF
--- a/src/gui/Src/BasicView/AbstractTableView.cpp
+++ b/src/gui/Src/BasicView/AbstractTableView.cpp
@@ -647,7 +647,7 @@ void AbstractTableView::leaveEvent(QEvent* event)
 void AbstractTableView::keyPressEvent(QKeyEvent* event)
 {
     int wKey = event->key();
-    if(event->modifiers())
+    if(event->modifiers() && event->modifiers() != Qt::KeypadModifier)
         return;
 
     if(wKey == Qt::Key_Up)
--- a/src/gui/Src/BasicView/AbstractTableView.cpp
+++ b/src/gui/Src/BasicView/AbstractTableView.cpp
@@ -647,7 +647,7 @@ void AbstractTableView::leaveEvent(QEvent* event)
 void AbstractTableView::keyPressEvent(QKeyEvent* event)
 {
     int wKey = event->key();
-    if(event->modifiers() && event->modifiers() != Qt::KeypadModifier)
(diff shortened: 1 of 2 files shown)


src/gui/Src/BasicView/AbstractTableView.cpp
```diff
@@ -647,7 +647,7 @@ void AbstractTableView::leaveEvent(QEvent* event)
 void AbstractTableView::keyPressEvent(QKeyEvent* event)
 {
     int wKey = event->key();
-    if(event->modifiers())
+    if(event->modifiers() && event->modifiers() != Qt::KeypadModifier)
```
mrexodia, line 650: Shouldn’t this also use `Qt::NoModifier` for consistency?

src/gui/Src/BasicView/HexDump.cpp
```diff
@@ -642,9 +644,9 @@ void HexDump::keyPressEvent(QKeyEvent* event)
...
         }
     }
-    else if(modifiers == Qt::ShiftModifier)
+    else
     {
-        //TODO
+        AbstractTableView::keyPressEvent(event);
     }
     /*
         Let's keep the old code for a while until nobody remembers previous behaviour.
```
mrexodia, line 652: Feel free to get rid of this too

mrexodia:
Thanks for the contribution! I left some minor suggestions but as far as I can tell the code is correct.

src/gui/Src/BasicView/AbstractTableView.cpp
starsunyzl: Yes, you're right.

src/gui/Src/BasicView/HexDump.cpp
starsunyzl: Not sure if this is useful, now it's removed.

mrexodia:
Thanks again!

src/gui/Src/BasicView/HexDump.cpp
starsunyzl: > Feel free to get rid of this too

> Not sure if this is useful, now it's removed.

Doesn't work after removing this, my fault, sorry.

PR #2792, opened 2021-11-19
Fix #2790 mistake

starsunyzl:
My fault, sorry for duplication of work.

see: 
https://github.com/x64dbg/x64dbg/pull/2790#pullrequestreview-810733354


mrexodia:
Thanks for the fix!

PR #2805, opened 2021-12-21
Eenhanced floating point support

torusrxxx:
Add string format to print floating point number pointer:

The user can print a float number with format `{f:pointer}` or double number with format `{F:pointer}`

Add the UI to modify floating point number in dump:

The user can edit the floating point number in dump by double-clicking.


torusrxxx:
We can have an x64dbg update on Xmas!

PR #2807, opened 2021-12-25
Add support for printing floating point numbers in XMM/YMM

torusrxxx:
Now only ST(...) is missing (ldconvert required)

### DIFF
--- a/src/dbg/stringformat.cpp
+++ b/src/dbg/stringformat.cpp
@@ -21,12 +21,103 @@ enum class StringValueType
     FloatingPointDouble
 };
 
+template<class T> String printFloatValue(FormatValueType value)
+{
+    assert(std::is_same<T, double>() || std::is_same<T, float>());
+    String result;
+    char buf[16];
+    memset(buf, 0, sizeof(buf));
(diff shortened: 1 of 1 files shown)


src/dbg/stringformat.cpp
```diff
@@ -21,12 +21,103 @@ enum class StringValueType
     FloatingPointDouble
 };
 
+template<class T> String printFloatValue(FormatValueType value)
+{
+    assert(std::is_same<T, double>() || std::is_same<T, float>());
```
mrexodia, line 26: This should be static_assert

torusrxxx: There seems to be a problem with static_assert but I don't use VS2013 anymore.

mrexodia:
I’ll check it out, seems like a very strange error…

src/dbg/stringformat.cpp
mrexodia: I fixed it, had to use `std::is_same<T, double>::value`

mrexodia:
Thanks!

PR #2808, opened 2021-12-26
Add autocomment for conditional breakpoint

torusrxxx:
There's something useful in conditional breakpoint, the user shouldn't have to copy them to comments.


mrexodia:
Thanks!

PR #2810, opened 2021-12-28
copy and paste breakpoint conditions

torusrxxx:
It works, but I don't have more time to polish it right now.


mrexodia:
Thank you!

PR #2817, opened 2022-01-06
Adding search action in LogView

rajkumarananthu:
To increase flexibility for user to search for a pattern
inside the log instead of copying to some other tool.

### DIFF
--- a/src/gui/Src/Gui/LogView.cpp
+++ b/src/gui/Src/Gui/LogView.cpp
@@ -40,6 +40,8 @@ LogView::LogView(QWidget* parent) : QTextBrowser(parent), logRedirection(NULL)
     connect(Bridge::getBridge(), SIGNAL(setLogEnabled(bool)), this, SLOT(setLoggingEnabled(bool)));
     connect(Bridge::getBridge(), SIGNAL(flushLog()), this, SLOT(flushLogSlot()));
     connect(this, SIGNAL(anchorClicked(QUrl)), this, SLOT(onAnchorClicked(QUrl)));
+    dialogSearchInLog = new LineEditDialog(this);
+    dialogSearchInLog->setWindowTitle(tr("Seach For"));
 
     duint setting;
     if(BridgeSettingGetUint("Misc", "Utf16LogRedirect", &setting))
@@ -105,6 +107,7 @@ void LogView::setupContextMenu()
     menuCopyToNotes->addAction(actionCopyToDebuggeeNotes);
     actionAutoScroll->setCheckable(true);
     actionAutoScroll->setChecked(autoScroll);
+    actionSearchInLog = setupAction(DIcon("search-for.png"), tr("Search Log"), this, SLOT(searchInLogSlot()));
 
     refreshShortcutsSlot();
     connect(Config(), SIGNAL(shortcutsUpdated()), this, SLOT(refreshShortcutsSlot()));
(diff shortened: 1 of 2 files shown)


mrexodia:
It looks like the files aren’t formatted properly. You can run format.bat manually or install.bat to automatically format on commit

src/gui/Src/Gui/LogView.cpp
```diff
@@ -415,6 +420,27 @@ void LogView::copyToDebuggeeNotes()
     emit Bridge::getBridge()->setDebuggeeNotes(Notes);
 }
 
+void LogView::searchInLogSlot() {
+    dialogSearchInLog->show();
+
+    if (dialogSearchInLog->exec() == QDialog::Accepted) {
+        QList<QTextEdit::ExtraSelection> extraSelections;
+        QColor highlight = QColor(Qt::blue);
```
mrexodia, line 428: Here it would be good to reuse the existing ConfigColor for highlighting and match the highlighting behavior in the disassembler (change text color instead of background).

src/gui/Src/Gui/LogView.cpp
```diff
@@ -415,6 +420,27 @@ void LogView::copyToDebuggeeNotes()
...
+
+        // finding all occurances matching the regex given from start
+        while(find(QRegExp(dialogSearchInLog->editText))) {
+            QTextEdit::ExtraSelection extra;
+            extra.format.setBackground(highlight);
+            extra.cursor = textCursor();
+            extraSelections.append(extra);
+        }
+        // highlighting all those selections
+        setExtraSelections(extraSelections);
```
mrexodia, line 440: It would also be good to at least scroll to the first occurrence (unless I missed it) and perhaps if you search the same again start searching from the selection (similar to the notepad find).

mrexodia (changes requested):
Thanks for your contribution!

rajkumarananthu:
Hi, I have a few questions here:
1. Notepad is not wrapping the search after it reaches the end of the file, should I do what notepad is doing. Or can we wrap it, like once the last occurrence of the search string is found and user still wants to search for it, should I wrap it point to the first occurrence? 
2. To find the next occurrence notepad is having a button saying 'Find Next', here we have to do 'Ctrl+F' (search) again for same string or empty (because that is how LineEditDialog is working), is that fine?
3. And the last searched pattern will remain highlighted till there is an unsuccessful search or a restart of application, is that okay?
4. And the highlighting is like this: all the matches are highlighted with change of text color, but the current selection is highlighted with blue background.
highlighting_changed

Let me know what you think of this update? If you have any suggestions please let me know, I am open for learning :)

mrexodia:
Does the usage feel good to you? I was thinking and perhaps you can implement it similar to how the disassembly references work:

1. `Ctrl+F` to open the search dialog, internally this will find all occurrences
2. Navigate to the first occurrence found from the selection (to replicate a text editor-like experience)
3. Use `Ctrl+L` to go to the next occurrence, use `Ctrl+L` to go to the previous one `ActionGotoPreviousReference` and `ActionGotoNextReference`

mrexodia:
Will check the rest later

rajkumarananthu:
Hi @mrexodia, I thought I have already committed the changes on this, my bad. Please review the updates that I did as per your suggestions, let me know if there are any changes required.

Thanks
Rajkumar Ananthu.

mrexodia (approved):

mrexodia:
Thanks for your contribution!

PR #2821, opened 2022-01-18
added some C++ datatypes to HexEditDialog

SirHerpDerp:
Dear x64dbg team,

I've just added two new datatypes to the HexEditDialog,
std::string = "\\x90\\x90"
std::vector<char> = { '\x90', '\x90' }

Why?
I'm currently writing a cpp application and wanted to go away from these old C-datatypes (eg. char *) and use some modern things. I'm not a professional coder so I read much things on the web. Most people say that std::vector is commonly used as a flexible data storage. So I think that adding std::string and std::vector<char> could be useful for someone.

### DIFF
--- a/src/gui/Src/Gui/HexEditDialog.cpp
+++ b/src/gui/Src/Gui/HexEditDialog.cpp
@@ -57,7 +57,7 @@ HexEditDialog::HexEditDialog(QWidget* parent) : QDialog(parent), ui(new Ui::HexE
     mTypes[DataCString] = FormatType { tr("C-Style String"), 1 };
     mTypes[DataCUnicodeString] = FormatType { tr("C-Style Unicode String"), 1 };
     mTypes[DataCShellcodeString] = FormatType { tr("C-Style Shellcode String"), 1 };
-    mTypes[DataCppString] = FormatType { tr("C++-Style String"), 16 };
+    mTypes[DataCppString] = FormatType { tr("C++-Style String"), 1 };
     mTypes[DataCppCharVector] = FormatType { tr("C++-Style Char Vector (Hex)"), 16 };
     mTypes[DataASMByte] = FormatType { tr("ASM-Style BYTE (Hex)"), 16 };
     mTypes[DataASMWord] = FormatType { tr("ASM-Style WORD (Hex)"), 12 };
--- a/src/gui/Src/Gui/HexEditDialog.cpp
+++ b/src/gui/Src/Gui/HexEditDialog.cpp
@@ -57,6 +57,8 @@ HexEditDialog::HexEditDialog(QWidget* parent) : QDialog(parent), ui(new Ui::HexE
     mTypes[DataCString] = FormatType { tr("C-Style String"), 1 };
     mTypes[DataCUnicodeString] = FormatType { tr("C-Style Unicode String"), 1 };
     mTypes[DataCShellcodeString] = FormatType { tr("C-Style Shellcode String"), 1 };
+    mTypes[DataCppString] = FormatType { tr("C++-Style String"), 16 };
+    mTypes[DataCppCharVector] = FormatType { tr("C++-Style Char Vector (Hex)"), 16 };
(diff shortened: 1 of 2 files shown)


mrexodia:
Thanks, but I think the first one is already covered by C shellcode string. The second one is already covered by a C array.

Although perhaps there options should be renamed?

SirHerpDerp:
Thanks for your fast reply.
This is strange, I looked through all existing types again right now, but mine does not appear there.

C Shellstring: `"\x90"`
C++ std::string: `"\\x90"`
(double backslash)

C-Array: `{ 0x90, 0x90, 0x90 };`
C++Vector: `{ '\x90', '\x90', '\x90' };`

Or am I missing something?

SirHerpDerp:
> > C++ std::string: `"\\x90"`
> 
> You can use a raw string (`R"(\x90)"`)

Aaah, didn't know that this is possible. Thanks for this tip. So if the std::string doesn't make it to x64dbg I can use this instead. I just tested it with MSVC and `R"(\x90\x90)"` gets converted to `"\\x90\\x90"` automatically.

mrexodia:
The idea is to get `\x90` though so you will have the byte 0x90 in your std::string...

SirHerpDerp:
You are absolutely correct.
It took me a while for figuring out why I haven't simply used "\x90". I tested this and ran into some trouble because `std::string` seems to be null terminated, `\x90\x00\x90` becomes `\x90`.
But justanotheranonymoususer solution works like a charm.

So I would remove the code for the `std::string`, but how about the `std::vector<char>`?

mrexodia:
The `R"(\x90)"` will not emit a `0x90` byte, it just emits the string `\x90`: https://godbolt.org/z/167495qra

For `vector<char>` you can use `C-Style BYTE (Hex)` and prepend `std::vector<char> v =`. It will work without any special changes.

I really appreciate your contribution, but as it is now I don't want to merge it.

SirHerpDerp:
I thought about this the last few days, too and agree that implementing `std::string` is unnecessary since you could achieve the same by using a raw string with the `C-Style Shellcode String`.
On top of that this string seems to be useless until parsed into a `char*` for example.

But I have to disagree with you about the vector.
Since you are, compared to me, a much better coder I think the problem relies on my end. Sorry for stealing you time with this. :)

I really tried to use the `C-Style BYTE (Hex)` in combination with a char vector but with no success.

Tried to compile the following code:
```
#include <vector>

int main(){
    std::vector<char> v = { 0x55, 0x8B, 0xEC };
    return 0;
}
```

it won't compile on MSVC and GCC for me, since it will be interpreted as integers.
When using `std::vector<char> v = { "0x55", "0x8B", "0xEC" };` it compiles without any problems.
It's also directly usable with `WriteProcessMemory(pHandle, address, v.data(), v.size(), NULL);`.

ZehMatt:
Something like this can be also be achieved with plugins. I wrote something like this for DotX64Dbg

See: https://gist.github.com/ZehMatt/7bd7dae19ccbb9fa5412a34ae2236462

mrexodia:
I will try another time, but maybe try vector<unsigned char>

I’m pretty sure that was working for me, with char the value 0x8B isn’t valid

merces:
> with char the value 0x8B isn’t valid

Exactly. I think that's the problem, @SirHerpDerp. With `char`, you can't have a value greater than `CHAR_MAX` (`#include <limits>`), which is `127`, or `0x7f`, is most architectures. It should compile if you change to `vector<unsigned char>`, for example.

SirHerpDerp:
Oof.

Yes, you are absolutely right!
Well I never used an `unsigned char` till today, never thought that this exists, thought `char` ignores the two's complement.
Seems like I have to change all `char` to `unsigned char` in my project the next days.

@ZehMatt thanks for sharing this code, I will definitely take a look. But writing a plugin seems like alien technology for me right now.

Thanks to everyone who has commented here and for taking your time. I definitely learned some new things here.

So since this PR is useless I will close it right now.

PR #2834, opened 2022-02-12
Extra Log method for HTML

Montycarlo:
Allow developers to add HTML to the displayed log window, for instance to highlight certain parts of the log messages.

```
Log.msg("<span style="color:red;">This message will be colored</span>);
```

Original PR;
> 
> Allow developers to add color styling to the displayed log window, to highlight certain parts of the messages.
> 
> ```
> Log.msg("[color@red] This message will be colored. [stopcolor]")
> ```
> 
> 2022-02-12-18_36-NKtTQNO8JA
> 
> 
> -----
> 
> This is probably not the syntax we would ideally want, but we cannot directly enter HTML into the log due to the escaping in LogView.cpp on line 263:
> ```   
> msgUtf16 = msgUtf16.toHtmlEscaped();
> ```
> 
> Additionally, we cannot use the `{stringformat}` syntax as described [here](https://help.x64dbg.com/en/latest/introduction/Formatting.html) as this is processed prior to the HTML escaping.
> 
>

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -1130,6 +1130,11 @@ BRIDGE_IMPEXP void GuiAddLogMessage(const char* msg)
     _gui_sendmessage(GUI_ADD_MSG_TO_LOG, (void*)msg, 0);
 }
 
+BRIDGE_IMPEXP void GuiAddLogMessageHtml(const char* msg)
+{
+    _gui_sendmessage(GUI_ADD_MSG_TO_LOG_HTML, (void*)msg, 0);
+}
+
 BRIDGE_IMPEXP void GuiLogClear()
 {
     _gui_sendmessage(GUI_CLEAR_LOG, 0, 0);
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -1089,6 +1089,7 @@ typedef enum
     GUI_DISASSEMBLE_AT,             // param1=(duint)va,            param2=(duint)cip
     GUI_SET_DEBUG_STATE,            // param1=(DBGSTATE)state,      param2=unused
     GUI_ADD_MSG_TO_LOG,             // param1=(const char*)msg,     param2=unused
+    GUI_ADD_MSG_TO_LOG_HTML,        // param1=(const char*)msg,     param2=unused
     GUI_CLEAR_LOG,                  // param1=unused,               param2=unused
     GUI_UPDATE_REGISTER_VIEW,       // param1=unused,               param2=unused
     GUI_UPDATE_DISASSEMBLY_VIEW,    // param1=unused,               param2=unused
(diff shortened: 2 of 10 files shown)


justanotheranonymoususer:
How about the more HTML-like:
`Log.msg("[color=red] This message will be colored. [/color]")`

torusrxxx:
If the debuggee prints "[stopcolor][stopcolor][stopcolor]" via OutputDebugString, conditional logging breakpoint or something else, your patch would not work. This is why there is call to toHtmlEscaped() in the first place.
I don't mind emitting some HTML into log, as long as it is not possible to misinterpret strings generated by the debuggee as format markup. HTML templates and prepared SQL statements are good examples where user-generated content is escaped by default and then formatted. This may require new APIs so that older plugins don't break.

Montycarlo:
Thanks for your comments  - I'll take your suggestion @torusrxxx and provide a safer alternative.

Montycarlo:
> If the debuggee prints "[stopcolor][stopcolor][stopcolor]" via OutputDebugString, conditional logging breakpoint or something else, your patch would not work. This is why there is call to toHtmlEscaped() in the first place. I don't mind emitting some HTML into log, as long as it is not possible to misinterpret strings generated by the debuggee as format markup. HTML templates and prepared SQL statements are good examples where user-generated content is escaped by default and then formatted. This may require new APIs so that older plugins don't break.

With this in mind, would you open to accepting a new log method that explicitly accepts HTML? It is then up to the plugin developer to HTML-escape their input where appropriate (but are then free to style spans, etc).

mrexodia:
I would be totally fine with that!

Montycarlo:
Ready for your review @mrexodia

torusrxxx:
Actually I want x64dbg to take care of escaping the string to be formatted
with "%s" format (and all other formats that could emit the "<" character).
This way plugins don't have to write HTML escape code and at the same time
it is guaranteed to be safe unless they put unsafe strings directly into
the format string. Presently: a plugin could format "<font
color=red>%s</font>" and it works, but a security problem exists silently.
Also I think it's a great time to add support for other changes to the log
with the new API, for example add the log filter system (requires
additional arguments to specify the type and source of log message).

Montycarlo:
> Actually I want x64dbg to take care of escaping the string to be formatted with "%s" format (and all other formats that could emit the "<" character). This way plugins don't have to write HTML escape code and at the same time it is guaranteed to be safe unless they put unsafe strings directly into the format string.

Could you get an example of how a client would use such an interface? I didn't want to go down the route of having to create and expose a heap of methods for `logStringAsColor` or `logStringWithClass` etc.

mrexodia:
If that doesn’t work for you @torusrxxx we can also only leave `GuiAddLogMessageHtml` and get rid of the printf html function. The number of users is probably 1 so it doesn’t make sense for me to write a whole custom printf implementation.

torusrxxx:
I don't think it would be used only be 1 user. Many plugin developers would
be delighted to add color to their log. If you don't like to write a secure
printf version, yeah it can be left out. GuiAddLogMessageHtml is required
anyway no matter whether we decide to add printf support.

mrexodia:
So we can merge basic HTML support now and then add the printf overload down the line 👍

Montycarlo:
Changed exported function name to `_plugin_lograw_html` taking a single string and removed the parameters.

mrexodia:
Sorry for the long wait, but it looks great and I'll merge it.

Thanks a lot!

PR #2844, opened 2022-03-10
Script waits too long or freezes while running. x64dbg#2843

qwertyu131:
A problem arising from the simultaneous execution of script and patch operations.

Until the script file is finished, I made sure that it does not do the patch operations. The problem was solved. It was enough to call the patch list once after the script process was finished.


qwertyu131:
I made it active again.

ZehMatt:
Still empty

PR #2849, opened 2022-03-14
Fix for printable characters

mobile46:
Hi, some users like me can't use (non-ascii) characters like ç, ğ, ş, ö in some text boxes. I think this patch fixes that.


mrexodia:
Thanks!

PR #2872, opened 2022-05-05
Copy a memory address formatted as a raw data. (edit)

Rat431:
Implement a memory address copying in the clipboard as a raw-data format from the x64dbg's context menu.

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -288,13 +288,12 @@ void CPUDisassembly::setupRightClickContextMenu()
     mMenuBuilder->addMenu(makeMenu(DIcon("binary.png"), tr("&Binary")), binaryMenu);
 
     MenuBuilder* copyMenu = new MenuBuilder(this);
-    QAction* rawAddrAction = makeAction(DIcon("copy_address.png"), tr("Address to pattern"), SLOT(copyRawAddressSlot()));
     copyMenu->addAction(makeShortcutAction(DIcon("copy_selection.png"), tr("&Selection"), SLOT(copySelectionSlot()), "ActionCopy"));
     copyMenu->addAction(makeAction(DIcon("copy_selection.png"), tr("Selection to &File"), SLOT(copySelectionToFileSlot())));
     copyMenu->addAction(makeAction(DIcon("copy_selection_no_bytes.png"), tr("Selection (&No Bytes)"), SLOT(copySelectionNoBytesSlot())));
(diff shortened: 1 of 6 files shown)


src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -288,11 +288,13 @@ void CPUDisassembly::setupRightClickContextMenu()
     mMenuBuilder->addMenu(makeMenu(DIcon("binary.png"), tr("&Binary")), binaryMenu);
 
     MenuBuilder* copyMenu = new MenuBuilder(this);
+    QAction* rawAddrAction = makeAction(DIcon("copy_address.png"), tr("Address to pattern"), SLOT(copyRawAddressSlot()));
```
mrexodia, line 291: You can directly inline this in the `addAction` below

src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -1556,6 +1558,25 @@ void CPUDisassembly::copyAddressSlot()
     Bridge::CopyToClipboard(clipboard);
 }
 
+void CPUDisassembly::copyRawAddressSlot()
```
mrexodia, line 1561: What exactly does this copy? The byte pattern `AA BB CC DD` or am I misunderstanding?

image

If I understand correctly this should do the same.

mrexodia (changes requested):
Thanks for the contribution. I didn't have time to try it (on the run), but added some comments for clarification.

src/gui/Src/Gui/CPUDisassembly.cpp
Rat431: Hi, go in the "Copy" section and look for "Address to pattern". This option will just take the memory address formatted as a raw pattern, ready to be written (pasted) in memory resulting as a valid pointer without having to manually decompose the address.

Rat431:
> ✅ [Download x64dbg 1.0.1336](https://ci.appveyor.com/project/mrexodia/x64dbg/builds/43465432/artifacts) (commit [f69c3932aa](https://github.com/x64dbg/x64dbg/commit/f69c3932aa) by @Rat431)

Just made the change you requested.

mrexodia:
Thanks for your contribution, but I don't think this use case is frequent enough to warrant a new menu item. It should be pretty straightforward to do this with a plugin or you can do something like `bswap(dis.sel())`. It could be great to add a `Bytes:` column to the calculation like the register edit dialog though because I use the `bswap()` function quite a bit when using x64dbg as a calculator.

mrexodia:
I implemented this feature in the calculator dialog, thanks for the idea!

image

PR #2889, opened 2022-06-20
Case insensitive string functions

justanotheranonymoususer:


src/dbg/exprfunc.cpp
```diff
@@ -616,6 +627,23 @@ namespace Exprfunc
...
+        assert(argc == 1);
+        assert(argv[0].type == ValueTypeString);
+
+        size_t len1 = ::strlen(argv[0].string.ptr);
+        size_t len2 = ::strlen(argv[1].string.ptr);
+        auto it = std::search(
+            argv[0].string.ptr, argv[0].string.ptr + len1,
+            argv[1].string.ptr, argv[1].string.ptr + len2,
+            [](char ch1, char ch2) { return std::toupper(ch1) == std::toupper(ch2); }
+        );
```
justanotheranonymoususer, line 641: Based on https://stackoverflow.com/a/19839371

src/dbg/exprfunc.cpp
```diff
@@ -13,6 +13,8 @@
 #include "exhandlerinfo.h"
 #include <vector>
 #include <regex>
+#include <string>
```
justanotheranonymoususer, line 16: Was already used in the file

mrexodia:
Thank you!

PR #2892, opened 2022-06-26
API GuiSetFavouriteCommand

yjd:
Solve the problem of writing plug-ins to use GuiAddFavouriteCommand, x64dbg repeatedly adds items every time it starts

### DIFF
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -1171,9 +1171,8 @@ typedef enum
     GUI_UPDATE_GRAPH_VIEW,          // param1=unused,               param2=unused
     GUI_SET_LOG_ENABLED,            // param1=bool isEnabled        param2=unused
     GUI_ADD_FAVOURITE_TOOL,         // param1=const char* name      param2=const char* description
-    GUI_SET_FAVOURITE_TOOL_SHORTCUT,// param1=const char* name      param2=const char* shortcut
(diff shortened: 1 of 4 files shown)


src/bridge/bridgemain.h
```diff
@@ -1171,8 +1171,9 @@ typedef enum
     GUI_UPDATE_GRAPH_VIEW,          // param1=unused,               param2=unused
     GUI_SET_LOG_ENABLED,            // param1=bool isEnabled        param2=unused
     GUI_ADD_FAVOURITE_TOOL,         // param1=const char* name      param2=const char* description
-    GUI_ADD_FAVOURITE_COMMAND,      // param1=const char* command   param2=const char* shortcut
     GUI_SET_FAVOURITE_TOOL_SHORTCUT,// param1=const char* name      param2=const char* shortcut
+    GUI_ADD_FAVOURITE_COMMAND,      // param1=const char* command   param2=const char* shortcut
+    GUI_SET_FAVOURITE_COMMAND_SHORTCUT,      // param1=const char* command   param2=const char* shortcut
```
mrexodia, line 1176: You cannot add messages in the middle. It breaks binary compatibility (just move it to the end)

yjd: > You cannot add messages in the middle. It breaks binary compatibility (just move it to the end)

Since I don't know QT, the function I want has not been tested successfully :(
src/gui/Src/Gui/MainWindow.cpp checksum error

Issue #2589, opened 2021-02-01
how to set BP on the selected commands in list? 


reverser69:
hi all
i search for "all intermodular call" then I'm not able to select a portion of codes in the list(shift+click) and set BP on them. the only option is to set BP on all. I'm using Jan 14 2021 ver
sometimes 64dbg won't respond anymore after setting many BPs. is there any way to solve this issue?
any help?

mrexodia:
For anyone interested: it shouldn't be too difficult to extend the reference view to allow multi-selection.

PR #2901, opened 2022-07-03
Add ability to set breakpoints on multiple selected lines in the References tab

eltimen:
issue #2589

Also I fixed array out of bounds assertion failure at this line:
https://github.com/x64dbg/x64dbg/blob/56eef9553ec2d45f45b5036078be9cff013e64df/src/gui/Src/BasicView/SearchListView.cpp#L380
Return value of `QKeyEvent::text()` can be empty string for modifier key press event (Ctrl, Shift, ...)


mrexodia:
Thanks!

PR #2903, opened 2022-07-05
Fix Shift key combination bug

yjd:


mrexodia:
Thanks for your contribution! It would help if you could describe the bug in words so it's easier to understand what this actually fixes.

yjd:
@mrexodia 


I type in the Shift+H key combination and it pops up with an error ignoring the Shift key.
Testing any Shift+xxx combination doesn't work.

mrexodia:
Thanks for the info, I will give it a try and let you know!

mrexodia:
Sorry for the late reply, thanks for the fix!

PR #2905, opened 2022-07-05
API GuiSetFavouriteCommandShortcut

yjd:
Solution：
1， write plug-ins to use GuiAddFavouriteCommand, x64dbg repeatedly add items every time you start the problem
2，add API GuiSetFavouriteCommandShortcut for plug-in use

Problems
Tested. Writing plugins to register hotkeys using GuiSetFavouriteCommandShortcut or _plugin_menuentrysethotkey has a lower priority than x64dbg internal setGlobalShortcut, if a new hotkey registered by the plugin has already been setGlobalShortcut is already used, the plugin registration hotkey will be invalid.


mrexodia:
Thanks for the report. I agree the current behavior is bugged, but your fix is not building and needs some more work. I converted it to https://github.com/x64dbg/x64dbg/issues/3023

PR #2911, opened 2022-07-15
Add horizontal scrollbar to CPUInfoBox

### LINKED ISSUE #2698: When the message is too long, it will cause some parts of the content can not be displayed
Version:snapshot_2021-07-01_23-17
System:win 10 10.0.19042
When the message is too long, it can result in parts of the content not being displayed, as shown below：
Copy content normally

eltimen:
closes #2698
Screenshots:
изображение

изображение

Implemented via adding optional mode of `StdTable`: auto-resizing columns in proportion to the maximum text length in the cells.

### DIFF
--- a/src/gui/Src/BasicView/StdTable.cpp
+++ b/src/gui/Src/BasicView/StdTable.cpp
@@ -5,6 +5,11 @@ StdTable::StdTable(QWidget* parent) : AbstractStdTable(parent)
 {
 }
 
+void StdTable::enableColumnsAutoResize(bool enabled)
+{
+    mIsColumnsAutoResizeAllowed = enabled;
+}
+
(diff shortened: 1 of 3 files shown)


eltimen:
Ok. It's just that I recently ran into some inconvenience when trying to view a long string. I used the "Follow in dump" menu item to do this. 
Probably the best solution is to override `wheelEvent` or `keyEvent` to add the ability to scroll through the mouse wheel and arrow buttons.

mrexodia:
Yeah, totally agree it’s inconvenient but I don’t think the use case is common enough to take the extra space. Thanks a lot for your contribution though, most people don’t take the time ❤️

I think overriding the mouseWheel event should be doable, will take a look into that.

PR #2927, opened 2022-08-31
Update filter text in the launcher dialog

mobile46:
The debugger exe file (x32dbg.exe, x64dbg.exe) can be renamed but we can't select it from the launcher dialog. This commit updates the filter text.


mrexodia:
Thanks!

Issue #1635, opened 2017-07-02
Error starting process (CreateProcess, ERROR_NOT_SUPPORTED)


n0way0ut:
x64dbg version: build version 21 jun 2017

windows 7 sp1 x64

![launcher exe](https://user-images.githubusercontent.com/5345101/27772590-e75853aa-5f65-11e7-8ecd-a43d77a4b911.jpg)

AraHaan:
wait wot I use Windows 7 sp1 ultimate here and normally CreateProcess should work just fine. I might have to take a look at it later.

mrexodia:
It's related to the architecture of your operating system. Windows 32 bit works fine

n0way0ut:
it works fine also if I open it with x64dbg instead of x32dbg.

AraHaan:
it seems to work fine on my 64 bit install of windows 7 ultimate.

stonedreamforest:
May be the reason for the plugin

wk-952:
@n0way0ut This looks like a valid .Net app, use something like [dnSpy](https://github.com/0xd4d/dnSpy) to debug it.
![untitled](https://user-images.githubusercontent.com/11916365/27869680-f570abda-61a0-11e7-97a0-e5e63526c964.png)

AraHaan:
ILSpy works too for it as well (well if you want to decompile it back to .net code from IL) @wk-952.

blaquee:
Im working on a fix for this. Should be done in a day or so after some testing. I'll need someone to test on 32bit OS if I dont install a VM by then

PR #2929, opened 2022-09-03
Reducing disk I/O operations by reusing local var value

greenozon:


mrexodia:
That makes sense, thanks! Originally this was a fix for https://github.com/x64dbg/x64dbg/issues/1635, but I guess I made a copy-paste error.

greenozon:
Great, thanks!
1 more question: should the lines ## 89 & 90 be moved inside the above subroutine GetPeArch()?
https://github.com/x64dbg/x64dbg/blob/1ec3031aca1b525b43ecb4fc7345bc36fd7b3153/src/dbg/commands/cmd-debug-control.cpp#L89

mrexodia:
I don't think so, that's a check specific to x64dbg to figure out what architecture the AnyCPU binary will run with.

mrexodia:
It seems that this change has caused a regression, see https://stackoverflow.com/a/7914778/1806760

greenozon:
Yeah, might be.. it means that  we may still get back to refactoring idea to keep all the arch related operations inside just 1 API
and not outside of it like..  - 
if(arch == PeArch::DotnetAnyCpu)

Otherwise if someone else would like to call GetPeArch()  they have to add afterwards if()... logic?..

PR #2935, opened 2022-09-17
Fix undecorated label display when it comes from exported symbols.

### LINKED ISSUE #2726: The labels shown in disassembly view are not undecorated when symbol is unloaded
Currently the disassembly view and graph view always show mangled symbol name even if UndecorateSymbolNames=true

owerosu:
Hello,

This patch fix a regression where undecorated label wasn't shown in assembly view when it comes from exported symbols.
Fix #2726.

Regards,
Owerosu


mrexodia:
Thanks!

PR #2939, opened 2022-09-24
Fix crash in function 'ProcessFileSections'

gmh5225:
Fix a crash caused by some virus modifications to the pe header.

### DIFF
--- a/src/dbg/memory.cpp
+++ b/src/dbg/memory.cpp
@@ -124,20 +124,27 @@ static void ProcessFileSections(std::vector<MEMPAGE> & pageVector)
 
         // Retrieve module info
         std::vector<MODSECTIONINFO> sections;
-        duint sectionAlignment = 0;
+        duint sectionAlignment = PAGE_SIZE;
         duint modSize = 0;
         duint sizeOfImage = 0;
         {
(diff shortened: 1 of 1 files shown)


mrexodia:
Could you share the executable on malshare?

gmh5225:
Sorry. I can't provide this sample right now due to some limitations. The main behaviour of this example is to make modInfo->headers=0 and modify the pe header.

mrexodia:
Can you send me just the header bytes or a summary of the PE (sections, optional header) with stuff redacted? I need to be able to reproduce this behavior locally and add it to my testbed…

mrexodia:
Actually I noticed you modified an old version that had a bug where the module name was used instead of the address to find the module. Could you try the latest version and see if it’s still bugged?

gmh5225:
> Actually I noticed you modified an old version that had a bug where the module name was used instead of the address to find the module. Could you try the latest version and see if it’s still bugged?

I'm really, really sorry. I forgot to upstream, no problem with the new version. I think I should close this PR. 0.0

mrexodia:
Would still appreciate some information on that malware. It was clearly triggering something weird…

Issue #2885, opened 2022-06-11
Folder/file path of the target to debug in the Window title?


ScyllaHide:
### Feature type

Quality of life

### Describe the feature

When debugging different versions of a program, it could be you lose track of which version you looking at. Showing the file path in the window title like IDA does could quickly help there.

mrexodia:
There is not enough room in the Window title in my opinion.

ScyllaHide:
on 1080p display:
IDA
https://postimg.cc/YvR3DVx7][img]https://i.postimg.cc/YvR3DVx7/ida.png
x64dbg
https://postimg.cc/9DCbpVZ1][img]https://i.postimg.cc/9DCbpVZ1/x64.png

not sure if the file path could be limited to only show the last 100 chars?! This would immensly help.

mrexodia:
Yeah I guess it could be an option to show the full path.

PR #2946, opened 2022-10-05
Shows filepath of current file in window title

ErnestoAMujica:
Addresses issue #2885. Could be a good idea to include an option to turn this on/off in the Preferences -> GUI settings menu, but I'm not sure how to do that.


mrexodia:
Just wanted to comment that I didn’t forget about this. It just needs some testing, also not sure about the 100 character limit…

mrexodia:
Still didn't forget 😅

PR #2952, opened 2022-10-11
add flags to csv export trace table

brandonros:


brandonros:
@mrexodia thoughts?

brandonros:
image

brandonros:
```
Index,Address,Bytes,Disassembly,Registers,Memory,Flags,Comments
0,00000001448A76A4,68 3A6BB31D,push 1DB36B3A,rax: 1448A76A4-> 1448A76A4 rbx: 7FFE0385-> 7FFE0385 rcx: 140000000-> 140000000 rdx: 1-> 1 rsp: 14F4A8-> 14F4A0 rbp: 0-> 0 rsi: 1-> 1 rdi: 7FFE0384-> 7FFE0384 r8: 0-> 0 r9: 0-> 0 r10: 7FFE0384-> 7FFE0384 r11: 246-> 246 r12: 1448A76A4-> 1448A76A4 r13: 0-> 0 r14: 140000000-> 140000000 r15: 0-> 0,000000000014F4A0: 170000-> 1DB36B3A,eflags: 244-> 246,
1,00000001448A76A9,E8 C2C96D00,call dts9_patcherv.144F84070,,000000000014F498: 7FFCE19547B1-> 1448A76AE,eflags: 246-> 246,

```

other than that, this actually works locally and looks good I believe?

brandonros:
image

mrexodia:
Not really sure why you need the flags if they are already in the register view...

brandonros:
they are not included in CSV output

should I get rid of flags column and just add eflags to registers column?

mrexodia:
Hm this is a good question. I think for most use cases you don’t really want the flags in there because it gets very verbose (although I totally agree it’s more complete with flags). What’s your use case for this? Perhaps you can achieve what you need with the binary format better…

brandonros:
working on reversing VMProtect3 packer with an instruction emulator and having to hand-diff compare it to what it does on a real processor via x64dbg CSV output

if we scrap the flags column can you add an option that says "should include flags output in trace" somewhere to circumvent the verbosity?

mrexodia:
Take a look at https://github.com/mrexodia/dumpulator/blob/main/tests/x64dbg-tracedump.py I use this script to diff x64dbg traces with dumpulator traces.

PR #2959, opened 2022-10-16
Use an icon to represent user/system party to reduce column width

torusrxxx:
Call stack, memory map and symbols view now get a party column


mrexodia:
Thanks!

PR #2966, opened 2022-11-01
Upgrade actions/checkout package to v3

austinvazquez:
Upgrade actions/checkout package from v2 to v3.

https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/

Signed-off-by: Austin Vazquez <macedonv@amazon.com>


mrexodia:
Thanks!

Issue #2838, opened 2022-02-20
Relative Paths for x96dbg.ini


MrMendelli:
### Feature type

Quality of life

### Describe the feature

I would like to be able to use relative paths in the launcher settings file to point to `x32dbg.exe` and `x64dbg.exe` respectively. I've already tried editing the file manually and assuming my implementation _(`.\`/`\`/`/`)_ isn't wrong, the launcher seemingly refuses the use of relative paths.

mrexodia:
This should be _relatively_ (heh) easy to implement. The relevant code is in: https://github.com/x64dbg/x64dbg/blob/development/src/launcher/x64dbg_launcher.cpp#L454

SwapnilKumbhar:
Is there any ongoing work on this issue? I would be happy to contribute if not!

mrexodia:
Yeah you’re welcome to! Be sure to check out the guide on how to build x64dbg. It’s a bit involved unfortunately… https://github.com/x64dbg/x64dbg/wiki/Compiling-the-whole-project

SwapnilKumbhar:
Thanks for the link. I'll give it a shot!

PR #2977, opened 2022-11-19
Add relative path support to launcher

not6:
#2838


mrexodia:
The reason there is an absolute path there is because the launcher’s current directory will be next to the program you debug, not next to the launcher itself.

I have to check more thoroughly, but you need to manually resolve the path to be next to the launcher executable instead of getting the full path.

not6:
@mrexodia This version seems to work fine, I don't know why the check fails though.
Also sorry if I do anything dumb, I haven't actually used GitHub or git overall that much before..

mrexodia:
Thanks!

MrMendelli:
I appreciate the efforts made to look into this, thank you.

PR #2982, opened 2022-11-26
Use common undecorate flags

justanotheranonymoususer:
To make the output more similar to other tools such as WinDbg.

For example:

x64dbg:
`public: class _bstr_t & __ptr64 __cdecl _bstr_t::operator=(class _bstr_t const & __ptr64) __ptr64`

WinDbg:
`public: class _bstr_t & __cdecl _bstr_t::operator=(class _bstr_t const &)`

I got the value `0x20800` by looking at the assembly of `get_undecoratedName` in msdia140, which calls `get_undecoratedNameEx` with these flags.


mrexodia:
Looks good, thanks!

PR #2987, opened 2022-12-03
Fix and improve "Show PID/TID in HEX" feature

not6:
There's still a lot of improvements that could be made, where the PID/TID is always hexadecimal, but this was the worst one IMO.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -473,6 +473,7 @@ static void DebugUpdateTitle(duint disasm_addr, bool analyzeThreadSwitch)
         _snprintf_s(modtext, _TRUNCATE, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Module: %s - ")), modname);
     char threadswitch[256] = "";
     DWORD currentThreadId = ThreadGetId(hActiveThread);
+    bool PIDTIDinhex = settingboolget("Gui", "PidTidInHex");
     if(analyzeThreadSwitch)
     {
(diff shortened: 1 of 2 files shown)


src/dbg/debugger.cpp
```diff
@@ -482,7 +483,10 @@ static void DebugUpdateTitle(duint disasm_addr, bool analyzeThreadSwitch)
         {
             char threadName2[MAX_THREAD_NAME_SIZE] = "";
             if(!ThreadGetName(PrevThreadId, threadName2) || threadName2[0] == 0)
-                sprintf_s(threadName2, "%X", PrevThreadId);
```
mrexodia, line 485: There is a function `formatpidtid` that does this

src/dbg/debugger.cpp
```diff
@@ -492,7 +496,7 @@ static void DebugUpdateTitle(duint disasm_addr, bool analyzeThreadSwitch)
     if(ThreadGetName(currentThreadId, threadName) && *threadName)
         strcat_s(threadName, " ");
     char PIDnumber[64], TIDnumber[64];
-    if(settingboolget("Gui", "PidInHex"))
```
mrexodia, line 495: Same, it can be switched to the function `formatpidtid`

mrexodia (changes requested):

not6:
@mrexodia Should be good now

mrexodia (approved):

mrexodia:
Awesome, thanks a lot!

PR #2995, opened 2022-12-10
Added context menu entries (Inc, dec, zero) to registers

Bytabyte:
Nice to have :)

### DIFF
--- a/src/gui/Src/Gui/CPURegistersView.cpp
+++ b/src/gui/Src/Gui/CPURegistersView.cpp
@@ -16,11 +16,11 @@ CPURegistersView::CPURegistersView(CPUWidget* parent) : RegistersView(parent), m
     // precreate ContextMenu Actions
     wCM_Modify = new QAction(DIcon("register_edit"), tr("Modify value"), this);
     wCM_Modify->setShortcut(QKeySequence(Qt::Key_Enter));
-    wCM_Increment = new QAction(DIcon("register_edit"), tr("Increment value"), this);
+    wCM_Increment = new QAction(DIcon("register_inc"), tr("Increment value"), this);
     wCM_Increment->setShortcut(QKeySequence(Qt::Key_Plus));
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/CPURegistersView.cpp
```diff
@@ -16,6 +16,12 @@ CPURegistersView::CPURegistersView(CPUWidget* parent) : RegistersView(parent), m
     // precreate ContextMenu Actions
     wCM_Modify = new QAction(DIcon("register_edit"), tr("Modify value"), this);
     wCM_Modify->setShortcut(QKeySequence(Qt::Key_Enter));
+    wCM_Increment = new QAction(DIcon("register_edit"), tr("Increment value"), this);
```
mrexodia, line 19: There's the icons `register_inc, register_dec, register_zero`

mrexodia (changes requested):

mrexodia:
@torusrxxx what do you think of restoring these actions?

src/gui/Src/Gui/CPURegistersView.cpp
```diff
@@ -542,6 +581,13 @@ void CPURegistersView::displayCustomContextMenuSlot(QPoint pos)
             wMenu.addAction(wCM_Modify);
         }
 
+        if(mINCREMENTDECREMET.contains(mSelected))
```
mrexodia, line 584: Also I think these options should be all the way at the bottom, since they are used less frequently.

Bytabyte: IMHO, I think these options should be under "Modify value" because they do similar job which is changing registers value.

Bytabyte: You'd be surprised how often these options are used.

mrexodia:
Thanks!

PR #3004, opened 2023-01-11
unexpected variable override

bb107:
Reference type bpColor will override the value of mBreakpointBackgroundColor


mrexodia:
Thanks a lot!

PR #3017, opened 2023-02-16
Get thread name from GetThreadDescription

### LINKED ISSUE #3016: Get default thread name with GetThreadDescription (WinAPI)
### Feature type

Quality of life

### Describe the feature

Starting with Windows 10, apps can set thread names on handles with `SetThreadDescription`. Then later this name can be retrieved by [`GetThreadDescription`](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getthreaddescription).

It would tremendously help if x64dbg could dynamically get address of the latter function if it's available, and populate the thread names in Threads tabs using it.

More and more apps are setting up descriptive debug names for their threads, so this would aid a lot instead of trying to manually figure out which thread is for what every time. Some other tools support this and it's very useful.

Currently only the main thread is named "Main Thread" by default:
https://github.com/x64dbg/x64dbg/blob/09bc3aed4f8267126d144e379c4fc14e4954113d/src/dbg/thread.cpp#L40-L44

I'd try making a PR myself, but I fear trying to get the codebase to compile and understand would take me a lot of time, while someone who already has the environment set up and knowledge of codebase could probably take a few minutes to implement this...

p0358:
Closes #3016 

It gets the name in ThreadCreate and ThreadGetList, both sources work on their own separately if you comment out the other one.

The reason I added thread name updating also to ThreadGetList is that technically the app can update its thread description at any time (which docs mention), which could mean initial name could get outdated, even though such situation is rather rare. But given the threads are refreshed every 1 second, it should be fine.

For code style I mimicked how _GetThreadId is accessed in the same file.

Tested:
image

There's one oddity I noticed though: the thread name won't be shown in Handles view instead of the thread number, unless you explicitly set a thread name yourself in GUI, even if to the same value. Is that expected behavior?

### DIFF
--- a/src/dbg/thread.cpp
+++ b/src/dbg/thread.cpp
@@ -37,9 +37,19 @@ void ThreadCreate(CREATE_THREAD_DEBUG_INFO* CreateThread)
     // Duplicate the debug thread handle -> thread handle
     DuplicateHandle(GetCurrentProcess(), CreateThread->hThread, GetCurrentProcess(), &curInfo.Handle, 0, FALSE, DUPLICATE_SAME_ACCESS);
 
+    typedef HRESULT(WINAPI * GETTHREADDESCRIPTION)(HANDLE hThread, PWSTR * ppszThreadDescription);
+    static GETTHREADDESCRIPTION _GetThreadDescription = (GETTHREADDESCRIPTION)GetProcAddress(GetModuleHandleW(L"kernel32.dll"), "GetThreadDescription");
+    PWSTR threadDescription = nullptr;
+
     // The first thread (#0) is always the main program thread
     if(curInfo.ThreadNumber <= 0)
         strcpy_s(curInfo.threadName, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Main Thread")));
+    else if(_GetThreadDescription && SUCCEEDED(_GetThreadDescription(curInfo.Handle, &threadDescription)) && threadDescription)
+    {
+        if(threadDescription[0])
+            strncpy_s(curInfo.threadName, StringUtils::Escape(StringUtils::Utf16ToUtf8(threadDescription)).c_str(), _TRUNCATE);
+        LocalFree(threadDescription);
+    }
     else
         curInfo.threadName[0] = 0;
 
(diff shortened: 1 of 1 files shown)


p0358:
```
hooks\AStyleHelper.exe check
Nonconforming files:
src/dbg/thread.cpp
Command exited with code 1
```

Not sure why the CI says this, on my local copy it outputs nothing from check/silent (but %errorlevel% is 9009???)

p0358:
nvm now it should be fine xD (Visual Studio lied to me that there were no changes in file)

mrexodia:
I'll get back to this, probably during the weekend.

src/dbg/thread.cpp
```diff
@@ -37,9 +37,19 @@ void ThreadCreate(CREATE_THREAD_DEBUG_INFO* CreateThread)
...
+    static GETTHREADDESCRIPTION _GetThreadDescription = (GETTHREADDESCRIPTION)GetProcAddress(GetModuleHandleW(L"kernel32.dll"), "GetThreadDescription");
+    PWSTR threadDescription;
+
     // The first thread (#0) is always the main program thread
     if(curInfo.ThreadNumber <= 0)
         strcpy_s(curInfo.threadName, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Main Thread")));
+    else if(_GetThreadDescription && SUCCEEDED(_GetThreadDescription(curInfo.Handle, &threadDescription)))
+    {
+        if(threadDescription[0])
+            strcpy_s(curInfo.threadName, StringUtils::Utf16ToUtf8(threadDescription).c_str());
```
mrexodia, line 50: This could overflow and crash x64dbg. Should probably be `strcpy_s(curInfo.threadName, StringUtils::Utf16ToUtf8(threadDescription).c_str(), _TRUNCATE);`

src/dbg/thread.cpp
```diff
@@ -127,6 +137,17 @@ void ThreadGetList(THREADLIST* List)
...
         memcpy(&List->list[index].BasicInfo, &itr.second, sizeof(THREADINFO));
 
+        typedef HRESULT(WINAPI * GETTHREADDESCRIPTION)(HANDLE hThread, PWSTR * ppszThreadDescription);
+        static GETTHREADDESCRIPTION _GetThreadDescription = (GETTHREADDESCRIPTION)GetProcAddress(GetModuleHandleW(L"kernel32.dll"), "GetThreadDescription");
+        PWSTR threadDescription;
+        if(_GetThreadDescription && SUCCEEDED(_GetThreadDescription(threadHandle, &threadDescription)))
+        {
+            // Thread name may have changed
+            if(threadDescription[0])
+                strcpy_s(List->list[index].BasicInfo.threadName, StringUtils::Utf16ToUtf8(threadDescription).c_str());
```
mrexodia, line 147: Ditto

p0358:
I fixed your suggestions and also, upon further thought, made the string be escaped like in the other place where thread name from exception is set. Also made it check if `threadDescription` isn't nullptr before dereferencing it -- it technically has no right to be if function succeeded (and Microsoft's sample code confirms that), unless someone tampered with kernel32, but better safe than sorry (and avoid crash here), right?

PR #3018, opened 2023-02-16
Don't break on SetThreadName exception

p0358:
My arguments for this change:
- this gets spammed a lot in some apps during startup or runtime, forcing you to keep pressing Go or disable first-chance exception breaking altogether (it doesn't look like there's another way to just filter out that one)
- it's supposed to be just an information for the debugger itself rather than a real exception
- it shouldn't be missed if needed since the operation is logged in the debug log

In case of the game I'm debugging this changed from having to press Go 30 times on startup and regularly during runtime, to literally nothing (because all these random exceptions I was getting were apparently just setting thread name on the debugger). 

So I think it would be a nice QoL change. I *think* most people shouldn't really need an exception like this, if someone wants to watch a thread being created, there are better ways to do it than wait on this exception by default...

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -2022,6 +2022,7 @@ static void cbException(EXCEPTION_DEBUG_INFO* ExceptionData)
                 String ThreadNameEscaped = StringUtils::Escape(ThreadName());
                 dprintf(QT_TRANSLATE_NOOP("DBG", "SetThreadName(%X, \"%s\")\n"), nameInfo.dwThreadID, ThreadNameEscaped.c_str());
                 ThreadSetName(nameInfo.dwThreadID, ThreadNameEscaped.c_str());
+                return;
             }
         }
     }
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -2020,9 +2020,10 @@ static void cbException(EXCEPTION_DEBUG_INFO* ExceptionData)
             if(MemRead((duint)nameInfo.szName, ThreadName(), MAX_THREAD_NAME_SIZE - 1))
             {
                 String ThreadNameEscaped = StringUtils::Escape(ThreadName());
-                dprintf(QT_TRANSLATE_NOOP("DBG", "SetThreadName(%X, \"%s\")\n"), nameInfo.dwThreadID, ThreadNameEscaped.c_str());
+                dprintf(QT_TRANSLATE_NOOP("DBG", "SetThreadName exception on %p (%X, \"%s\")\n"), addr, nameInfo.dwThreadID, ThreadNameEscaped.c_str());
                 ThreadSetName(nameInfo.dwThreadID, ThreadNameEscaped.c_str());
-                return;
+                if(ThreadNameEscaped.length() > 0 && !settingboolget("Events", "ThreadNameSet"))
+                    return;
             }
         }
     }
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -2022,7 +2022,7 @@ static void cbException(EXCEPTION_DEBUG_INFO* ExceptionData)
                 String ThreadNameEscaped = StringUtils::Escape(ThreadName());
                 dprintf(QT_TRANSLATE_NOOP("DBG", "SetThreadName exception on %p (%X, \"%s\")\n"), addr, nameInfo.dwThreadID, ThreadNameEscaped.c_str());
                 ThreadSetName(nameInfo.dwThreadID, ThreadNameEscaped.c_str());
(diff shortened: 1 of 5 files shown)


mrexodia:
You should be able to hide this exception by ignoring it (either globally or per-process) in the settings. Doing things this way would not allow a user to break on the exception anymore if they want to, so I don’t want to merge it.

p0358:
But as far as I see in the settings, I can only ignore particular base exception type like `0x406D1388` (`MS_VC_EXCEPTION`). That would be quite a bummer, because this exception is also used for various other things, like all uncaught `throw` in C++ on Windows. But to be setting thread name it needs to have specific `dwType` and `dwFlags` parameter, and also valid `dwThreadID` specified on the structure in `ExceptionData->ExceptionRecord.ExceptionInformation`. So ignoring the whole exception code altogether isn't the best option.

I think the only way to do it with settings, considering the checks, could be to add it to this page on settings (as something like `Thread Name Set`):
image

But I'm not sure if it's worth it. Do you think the above would be a better solution, should I try coding it?

mrexodia:
Perhaps for now it’s enough to create a boolean for it. The problem with silently continuing on an exception like this is that technically someone could craft an exception that gets hidden from x64dbg per default. Potentially some extra logic could be introduced to make sure the SetThreadName function is actually throwing the exception, but I’m not sure how feasible that is…

p0358:
> The problem with silently continuing on an exception like this is that technically someone could craft an exception that gets hidden from x64dbg per default.

That's a fair point, though it's not entirely hidden since thread name change is still getting logged, and the `return;` would get triggered only if all conditions including thread ID being valid pass. Potentially a check of >0 name length could be added before return is done. I think the relative risk of doing that by default should be low (though maybe the option to disable that for the sake of allowing people to still catch that exception if they really want would be rather good).

> Potentially some extra logic could be introduced to make sure the SetThreadName function is actually throwing the exception, but I’m not sure how feasible that is…

By checking stack trace? I think it's a bad idea though, because I've seen multiple bigger codebases just craft and throw the exception manually to set the thread name instead of actually calling SetThreadName (like WTF from Chromium/WebKit or Source Engine's tier0)

mrexodia:
I totally agree that for normal programs it is best to hide this exception from the user. Unfortunately we have to assume a malicious program per default.

Likely the best way forward is to add a specific option to the exceptions tab and change the logging to indicate that a thread name change exception is thrown.

p0358:
Okay, waiting on CI to test the UI thing. Not sure if my logging change is the best, but at least it indicates that `SetThreadName` is logged as a result of exception and logs the address by default, which might be sufficient (one should hopefully then figure out to look into settings if they want to break on it). And ofc it will still break either way if the exception parameters aren't valid for name setting (as could be in a malicious program).

p0358:
Seems to work.

With option enabled:
> Thread 1584 created, Entry: tforevive.[00007FF875E9C890](x64dbg://localhost/address64#00007FF875E9C890), Parameter: [000001AF04356000](x64dbg://localhost/address64#000001AF04356000)
> SetThreadName exception on [00007FF94368CD29](x64dbg://localhost/address64#00007FF94368CD29) (630, "[TFORevive] OptimizedTFOFilesystemStartup_Thread")
> EXCEPTION_DEBUG_INFO:
>            dwFirstChance: 1
>            ExceptionCode: 406D1388 (MS_VC_EXCEPTION)
>           ExceptionFlags: 00000000
>         ExceptionAddress: kernelbase.[00007FF94368CD29](x64dbg://localhost/address64#00007FF94368CD29)
>         NumberParameters: 3
> ExceptionInformation[00]: [0000000000001000](x64dbg://localhost/address64#0000000000001000)
> ExceptionInformation[01]: [000001AF04333200](x64dbg://localhost/address64#000001AF04333200)
> ExceptionInformation[02]: [0000000000000630](x64dbg://localhost/address64#0000000000000630)
> First chance exception on [00007FF94368CD29](x64dbg://localhost/address64#00007FF94368CD29) (406D1388, MS_VC_EXCEPTION)!

Without:
> Thread 8040 created, Entry: <tforevive.unsigned long __cdecl OptimizedTFOFilesystemStartup_Thread(void *)> ([00007FF875E9C890](x64dbg://localhost/address64#00007FF875E9C890)), Parameter: [000001AF043560E0](x64dbg://localhost/address64#000001AF043560E0)
> SetThreadName exception on [00007FF94368CD29](x64dbg://localhost/address64#00007FF94368CD29) (1F68, "[TFORevive] OptimizedTFOFilesystemStartup_Thread")

Doing force-push now to fix the whitespace issue (thanks Visual Studio) and to move Debug Strings into the gap (as the horizontal options were synced so to keep that)
image

mrexodia:
Meant to comment here. This definitely needs some more reviewing than #3017

PR #3028, opened 2023-02-28
New Option to Show Path of Debuggee in Window Title

### LINKED ISSUE #2885: Folder/file path of the target to debug in the Window title?

0xNB:
# Issue

Closes #2885
Closes #2946

>**Feature type**
>
>Quality of life
>
>**Describe the feature**
>
>When debugging different versions of a program, it could be you lose track of which version you looking at. Showing the file path in the window title like IDA does could quickly help there.

# Solution

Under the GUI settings we find the new option `Set Long Window Path`. If this is enabled, instead of the normal window title (showing current thread and other parameters) the absolute path to the debugge is shown instead.

image

When we attach to a target we can now choose between the two views shown below, by toggling the option:

image

image

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -1202,10 +1202,15 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
 
         // check if we need to change the main window title
         bool bNewWindowLongPath = settingboolget("Gui", "WindowLongPath");
-        if(bWindowLongPath != bNewWindowLongPath)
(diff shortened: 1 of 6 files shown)


mrexodia:
Thanks!

XiaLuoHun:
This option does not always take effect. When this option is enabled, it will not take effect the next time it is started in debug mode

mrexodia:
Thanks I messed this up when refactoring. Will be fixed soon!

ScyllaHide:
So it was finally added! awsome!

PR #3044, opened 2023-03-11
Misc Maintenance

torusrxxx:
* fixed "vardel x" said x is deleted, but x is not deleted actually.
* cleaned up XMM formatting. I wanted to allow printing XMM in hex but it seems to require more work later.

### DIFF
--- a/src/dbg/variable.cpp
+++ b/src/dbg/variable.cpp
@@ -297,9 +297,9 @@ bool varset(const char* Name, const char* Value, bool ReadOnly)
 \brief Deletes a variable.
 \param Name The name of the variable to delete. Cannot be null.
 \param DelSystem true to allow deleting system variables.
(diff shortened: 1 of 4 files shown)


src/dbg/variable.cpp
```diff
@@ -319,19 +319,21 @@ bool vardel(const char* Name, bool DelSystem)
     }
 
     if(!DelSystem && found->second.type != VAR_USER)
-        return false;
+        return -2;
     found = variables.begin();
-    String NameString(Name);
+    String NameString(name_);
```
torusrxxx, line 324: Maybe NameString is supposed to be the real name of a variable, not alias?

mrexodia:
I'll do my best to review this week. I totally forgot any of this code, so I would need to do some tests to see what's going on...

mrexodia:
Thanks!

PR #3077, opened 2023-04-29
Add breakpoint background color in call stack view

torusrxxx:
It is possible to press F2 to add a breakpoint directly in call stack view, but the visual indicator is missing.
Added a default color for hardware breakpoints as well.


mrexodia:
Thanks!

PR #3079, opened 2023-04-29
Add the DbgGetSymbolInfoAt API

m417z:
Mainly to overcome the restrictive `MAX_LABEL_SIZE` limit, and to be a able to get heavily templated symbols.

### DIFF
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -335,6 +335,7 @@ typedef enum
     DBG_MENU_PREPARE,               // param1=int hMenu,                 param2=unused
     DBG_GET_SYMBOL_INFO,            // param1=void* symbol,              param2=SYMBOLINFO* info
     DBG_GET_DEBUG_ENGINE,           // param1=unused,                    param2-unused
+    DBG_GET_SYMBOL_INFO_AT,         // param1=duint addr,                param2=SYMBOLINFO* info
 } DBGMSG;
 
 typedef enum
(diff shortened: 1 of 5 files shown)


src/bridge/bridgemain.h
```diff
@@ -334,6 +334,7 @@ typedef enum
     DBG_ANALYZE_FUNCTION,           // param1=BridgeCFGraphList* graph,  param2=duint entry
     DBG_MENU_PREPARE,               // param1=int hMenu,                 param2=unused
     DBG_GET_SYMBOL_INFO,            // param1=void* symbol,              param2=SYMBOLINFO* info
+    DBG_GET_SYMBOL_INFO_AT,         // param1=duint addr,                param2=SYMBOLINFO* info
```
mrexodia, line 337: New values can only be appended to the end, otherwise this would break things...

src/dbg/_exports.cpp
```diff
@@ -1653,6 +1653,21 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
     }
     break;
 
+    case DBG_GET_SYMBOL_INFO_AT:
+    {
+        auto modbase = ModBaseFromAddr((duint)param1);
```
mrexodia, line 1658: This check is already done in `SymbolFromAddressExact`, so it can be removed.

mrexodia (changes requested):

src/dbg/_exports.cpp
m417z: But the query is done again, I thought it wouldn't hurt. `ModInfoFromAddr` says it DOES _NOT_ USE LOCKS.

```cpp
MODINFO* ModInfoFromAddr(duint Address)
{
    //
    // NOTE: THIS DOES _NOT_ USE LOCKS
    //
    auto found = modinfo.find(Range(Address, Address));
    ///...
```

mrexodia:
Thanks! I've been looking at this for a use case I have as well. I added the `startRva` and `endRva` parameters to `enumSymbols`, but then I saw the terrible API and gave up from there...

src/dbg/_exports.cpp
mrexodia: The extra check doesn’t hurt, but it’s pointless and a (harmless) race condition.

The ModInfoFromAddr can only be used while holding a lock on the modules (hence the comment). This is done with the SHARED_ACQUIRE(?) macro and this ensures the module won’t be unloaded while symbols are being queried.

m417z: 👍 Removed the check.

mrexodia: Hm I just saw the issue, but I guess this will work fine.

mrexodia:
Thanks!

PR #3083, opened 2023-05-01
Set proper default style for example text in appearance dialog

torusrxxx:
Fine tuned the style of example text when using the transparent background color.
Qt cannot open text file with Chinese characters in file name??? Add error messages (releases from 2021 cannot open it either)


mrexodia:
Thanks as always!

PR #3085, opened 2023-05-03
Fixed tab order when user closing CPU tab and restarting the debugger

### LINKED ISSUE #3035: Restore previous tab location when closing a detached tab
### Operating System

Windows

### x64dbg Version

The last

### Describe the issue

Tabs don't return to their positions after "Detach". It is like a bug.

### Steps to reproduce

1. To move a tab.
2. To detach it.
3. To close it.

The tab should return to its last position. Instead of this, it becomes the last tab. And you have to move it again (to the position you want to see it). If you often need to detach some tab, it will be always the last. Please fix.

requizm:
Fix #3035


mrexodia:
This issue was already fixed from what I can tell.

PR #3091, opened 2023-05-07
Add the DbgXrefAddMulti API

m417z:
For some binaries, the XFG Marker plugin adds a large amount or xrefs (more than 100,000), and it takes several seconds. I added the DbgXrefAddMulti API to make adding a large amount of xrefs faster.

I also added other small fixes that I saw while browsing the code.


src/dbg/_exports.cpp
```diff
@@ -1670,6 +1670,8 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
 
     case DBG_GET_SYMBOL_INFO_AT:
     {
+        SHARED_ACQUIRE(LockModules);
```
m417z, line 1673: That's probably what I should have done in my previous PR.

src/gui/Src/Gui/ZehSymbolTable.cpp
```diff
@@ -4,24 +4,35 @@
...
+        memset(&info, 0, sizeof(info));
+        return &info;
+    }
+
+    SYMBOLINFO* get() { return &info; }
+    const SYMBOLINFO* get() const { return &info; }
+
     SYMBOLINFO* operator->() { return &info; }
+    const SYMBOLINFO* operator->() const { return &info; }
 };
```
m417z, line 36: This change is motivated by a mistake I made with the existing version of the wrapper - calling `GetSymbol(&info)` twice in a row may lead to a memory leak. I also think that overloading `&` is confusing, and common classes such as STL smart pointers provide a `get` method instead, which is also what I used.

m417z:
With this change, it runs 33% of the time. I hoped for a larger improvement, but it's still significant.

Before:

> Found and marked 4144 XFG entries, added 299538 xrefs, added 7221 comments
> Done in 5578 ms

After:

> Added 599076 xrefs in 1766 ms
> Found and marked 4144 XFG entries, added 599076 xrefs, added 7221 comments
> Done in 1844 ms

mrexodia:
The underlying data structure for this feature are quite bad. Could you share a binary?

m417z:
The target of XFG Marker? I'm using my system's explorer.exe for testing, version 10.0.22621.1485. Can also be downloaded [here](https://msdl.microsoft.com/download/symbols/explorer.exe/170D67714be000/explorer.exe).

m417z:
Here's some WIP code using DbgXrefAddMulti with time measurements if you want to play with it:
https://github.com/m417z/x64dbg-xfg-marker/tree/test-mutl-label

m417z:
I tried another approach which looks much better :)

> Added 599076 xrefs in 187 ms
> Found and marked 4144 XFG entries, added 599076 xrefs, added 7221 comments
> Done in 266 ms

I pushed as soon as it compiled (and surprisingly didn't crash), so please review carefully.

torusrxxx:
People familiar with DirectX know many APIs that may be used to process more than one element accept an array in order to maximize gaming performance.

m417z:
I used build 1.0.1580 for the recent couple of days and experienced no issues. Let me know if anything is missing for the PR to be merged.

mrexodia:
Yeah there’s nothing wrong with the PR. I just need to find free time to
sit down and properly review things.

mrexodia:
Looks good! I pushed some small changes to `XrefAddMulti` to your branch, please review to make sure I didn't totally break it.

src/dbg/xrefs.cpp
```diff
@@ -40,7 +40,7 @@ struct XrefSerializer : AddrInfoSerializer<XREFSINFO>
             record.addr = duint(json_hex_value(json_object_get(reference, "addr")));
             record.type = XREFTYPE(json_hex_value(json_object_get(reference, "type")));
             value.type = max(record.type, value.type);
-            value.references.insert({ record.addr, record });
+            value.references.emplace(record.addr, record);
```
m417z, line 43: It's a matter of preference, but I prefer `insert` here and in the other places. There's some good reasoning here:
https://quuxplusone.github.io/blog/2021/03/03/push-back-emplace-back/

Other than that, looks good.

mrexodia: Personally I tend to avoid following the melon man's advice. He makes a lot of 'technically correct' arguments and then actively suggests a _default_ that shouldn't apply unless you're a "library author" who has to consider every detail. A good example would be the [`for (auto&& elt : range)`](https://quuxplusone.github.io/blog/2018/12/15/autorefref-always-works/) post, which has likely ruined business-logic code for no good reason... `</rant>`

The linked post specifically talks about `vector`, where you can use `push_back` and `emplace_back` with the _same type_ and it then uses a (very questionable) benchmark to justify why you should always prefer `push_back`. The key difference here is that you pass the `w` directly. The correct way to use `emplace_back` would be: `widgets.emplace_back(foo, bar, baz);` (since constructs the object in-place), but he kind of glosses over this.

In our specific case I switched to `emplace` because it's simply more readable and I don't really see how the post about `vector` applies. It might if I did `references.emplace({ record.addr, record });` but this is not a valid use of `emplace` for map-like types...

m417z: 👍 Let's merge it, I think both are fine, I personally slightly prefer `insert` here, e.g. IDE type suggestions are usually better, but it's not that important.

mrexodia:
Awesome, thanks!

PR #3098, opened 2023-05-18
Introduced dump to trace

### LINKED ISSUE #2378: Stack and memory dump in trace browser
Add call stack and memory dump in trace browser, based on information available in trace file. This requires the trace file be scanned from current selected instruction up to the first one.

(I'm working on this, let me create an issue first, currently there is no dump or stack window in trace view, references and log view will be used first)

torusrxxx:
This fixes #2378
There's obviously a lot more to do. This pull request is here only for those who desperately want to dump from a trace or want to become alpha testers.
Right click on an instruction in the trace, and click the "Information" menu to get full dump of the trace up to the selected instruction into the log view.


torusrxxx:
The trace dump index is working and tested on a 136MB huge trace.
Speed is lightening. Search constant takes 10 seconds and memory reference search takes 46ms. When memory reference search is used for the first time it takes 17s including reading the trace file and building the index.
Memory usage is moderate. It saves a record in trace for every byte, so memory usage could be improved by saving a record for every dword instead. This is more complex due to special treatment needed for unaligned memory access. Currently it uses about 567MB for the 136MB test trace, may be acceptable.
Next step will be the GUI part.

mrexodia:
Would you like this to be merged into `development` or will you continue development on this branch? Currently the format checks are failing, so merging is blocked...

torusrxxx:
I think those new files are already formatted, no idea why appveyor doesn't like that. I'll pause development for a while, but I don't think this feature is completed. The aim of this pull request is to bring in dump and stack windows, not to merely accelerate memory reference search. Of course I support merging it if you think merging this would allow other people to use this feature and contribute to it. Possibly disable the debug feature of dumping the trace memory in log view when clicking on "information" menu, it should be available in the dump tab instead.

torusrxxx:
Dump widget has been added to trace, now it looks almost like the CPU view. There are still tons of usability problems and random crashes.

mrexodia:
It looks like this is conflicting with the changes in #3192 a bit. I will try to properly refactor the `HexDump` to take an `AbstractMemoryPage` with pure virtual `read/write` functions, since this will be useful for other projects as well.

mrexodia:
The following files are causing the formatting checks error btw:

```
src/gui/Src/Tracer/TraceDump.cpp
src/gui/Src/Tracer/TraceDump.h
src/gui/Src/Tracer/TraceFileDump.cpp
src/gui/Src/Tracer/TraceFileDump.h
```

Might be CRLF vs LF?

src/gui/Src/Tracer/TraceFileDump.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
```
captain-hoity-toity, line 1: 1. **Include Guards**: The `#pragma once` directive is not standard C++. It is better to use include guards for better portability.

```cpp
#ifndef TRACEFILEDUMP_H
#define TRACEFILEDUMP_H

// code

#endif // TRACEFILEDUMP_H
```

2. **Class Organization**: It would be better to separate `TraceFileDump` and `TraceFileDumpMemoryPage` into two different header files for better modularity and readability.

3. **Naming Conventions**: Class members should follow a consistent naming convention. For example, `maxIndex` should be `m_maxIndex` or `_maxIndex`.

4. **Access Modifiers**: `TraceFileDump::memAreas` is public and can be modified from outside the class. It would be better to make it private and provide a getter method if it's needed outside the class.

5. **Commenting**: The code lacks comments which makes it hard to understand the purpose of some methods and variables. It would be better to add comments to explain what each method and variable does.

6. **Error Handling**: There is no error handling in the `addMemAccess` method. It would be better to add some error checking to ensure that the method behaves correctly.

7. **Const Correctness**: Methods that don't modify the state of the object should be marked as `const`. For example, `getMaxIndex()` should be a `const` method.

8. **Thread Safety**: `TraceFileDumpMemoryPage::setSelectedIndex` and `TraceFileDumpMemoryPage::getSelectedIndex` are not thread-safe. It would be better to use a `QMutexLocker` to ensure thread safety.

9. **Unused Variables**: In the `TraceFileDump::DumpRecord` struct, `isWrite` and `isExecute` are declared but not used anywhere. It would be better to remove unused variables.

mrexodia: The `#pragma once` is preferable actually. There is absolutely zero reason to use include guards, ever.

Otherwise I guess fair points, but please comment at the appropriate place instead of the top of the file…

torusrxxx:
Maybe refactor `HexDump` first, and then we continue from that?

mrexodia:
The formatting of `TraceManager.cpp` is being reverted to `LF` for some reason. Perhaps you didn't set `git config --global core.crlf false`, or are you using the web editor/github.dev?

torusrxxx:
Yeah, the compiling guide doesn't mention it. The files with CR/LF encoding are uploaded on GitHub.com and next time it will be OK.

torusrxxx:
Dump tab is mostly usable now. Next focus should be getting trace tab and stack working and fixing crash.

torusrxxx:
DisableTraceDump setting has been added, it shouldn't break things horribly now.

mrexodia:
Heya, hope you're doing well! Do you have a list of things left to do for this? It would be nice to eventually merge all the PRs, but I don't know how stable this is...

torusrxxx:
I was busy as usual at the end of the year. I have finished using it back in October. I didn't test this since then. The stack window is missing. However it would require a bit of work which unfortunately could not be finished in 2023, and I think it's fine, you can already use the dump like a stack window. Then it would be in maintenance mode.

mrexodia:
🥳 will give it a try ASAP!

torusrxxx:
Xrefs dialog is added so you can quickly find out what accessed this memory location. Extremely easy and fast.

thug-shaker:
```math
```

torusrxxx:
This was waiting for review for a very long time. What do you think about this?

mrexodia:
Yeah I've been busy recently, just got around to doing some x64dbg maintenance today. The changes look good to me, but I think the disabling functionality could be better.

For me the ideal user experience would be:
- Disable the trace dump per default
- Whenever the user triggers an action that requires it, build the dump
- Allow enabling per default

This way no resources are consumed when it's not needed. But if you don't have time to implement this I will take a look at it later and we can just merge it as-is.

torusrxxx:
Ok, I'm going to improve that soon.

torusrxxx:
I'm adding support for tab detaching in new tabbed trace view. Currently there's crashing with certain user action sequences.

torusrxxx:
Now the dump is disabled, and a "load dump" button appears there. Clicking on the button will load the dump view.
Searching will also load the dump view automatically. This makes future searches faster.
And #3390 is fixed. Now the user can go to a virtual address. If multiple executions of the same virtual address are found, the xrefs dialog is displayed. The search uses the dump index to find all the indices instantly.

mrexodia:
So I've been testing the new functionality on a 700mb trace and unfortunately the default experience had some issues:

- Loading the trace puts x64dbg in `Not responding` (not a new issue, but wanted to mention it)
- Pressing the xref context menu goes into `Not responding` without issuing a warning and uses ~7GB of RAM
- After pressing the `Load dump` button the window does not restore the shape automatically
- There are no context menus to follow in dump, which means you need to manually Ctrl+G and type the address
- Closing the recording hangs for a very long time while freeing memory

I did some minor cleanups in the code (removed asserts, fixed warnings) and I added a warning message when loading the dump on big traces (just put an arbitrary ~0.2 GiB RAM usage):

image

Everything looks fine though, so let's merge this as-is and then improve things, I think the xref feature is quite useful!

PR #3102, opened 2023-05-24
Fix trace export and search being extremely slow on large trace files

shocoman:
Here, I made a GIF of the exporting process (the columns are current row, time taken to write 4096 rows, ram usage).

qtcreator_pITPqXLMAj

You can see how it suddenly slows down after the counter reaches something around 1 million rows. It quickly eats up all "cachable" 2048 pages and then just spends most of the time trying to find the oldest one to evict. With searching the situation is the same except the ram usage goes up to 1200MB.


mrexodia:
Thanks!

PR #3108, opened 2023-06-01
Add actions to copy DWord/QWord and Comments in the Stack view

shocoman:
It would be nice to be able to copy the second (data) column from the stack view.
I found several ways to do it which are not very intuitive or simple (copy the entire row with Ctrl-C and then somehow extract the needed column text, open the Modify dialog and copy the value from there, choose the "Copy data" tab of the Edit dialog and select "C-Style DWord" or something, copy raw bytes with Shift-C and then somehow reverse the endianness and remove spaces).
Besides, almost all other tables allow you to copy each column contents separately.
The added actions also can copy several rows of the selected column for no particular reason (the disassembly table behaves similarly).


mrexodia:
Thanks, seems to work fine!

PR #3118, opened 2023-06-23
Reduce number of cached trace file pages to save memory

torusrxxx:
I have decided to reduce number of cached trace file pages for x32dbg, over 1GB memory usage on x32dbg is unacceptable, especially considering #3098 will consume large amount of memory as well, and that is not paged.


mrexodia:
Thanks!

PR #3125, opened 2023-07-05
Symbol loading improvements

Mattiwatti:
This PR upstreams two small PDB handling-related changes I made while trying to get a MinGW-created PDB file to load (ultimately unsuccessfully, but at least this now gives an error message with the failure reason rather than deleting the PDB file).

For most users the most important change is probably the fact that "download symbols for all modules" should now be a lot faster if some or all of the symbol files already exist on disk. These files will still be validated, but not downloaded.


mrexodia:
Thanks!

Issue #3094, opened 2023-05-15
virtualmod command - status and intentions


ynwarcs:
### Operating System

Windows 10, Windows 11

### x64dbg Version

May 12 2023

### Describe the issue

Recently, someone asked about detecting memory-loaded modules via x64dbg in [this stackexchange post](https://reverseengineering.stackexchange.com/questions/31861/load-pdb-for-memorymodule-loaded-dll-in-x64dbg). I wanted to suggest [`virtualmod`](https://help.x64dbg.com/en/latest/commands/analysis/virtualmod.html) but after some quick testing, I realized that the command is broken right now.

After some experimentation, I managed to get it working for the use case mentioned in the post without much trouble. The issues that needed sorting out are:
1. `loadedSize` of the `MODINFO` structure is set to zero, which immediately fails the analysis of the module due to bounds checks in `ImageNtHeaders`.
2. The remote process memory belonging to the virtual module is mapped into the debugger memory via a buffer owned by a stack `Memory<unsigned char*>` variable. After the analysis is finished, the variable is destroyed and the module is left pointing to free'd memory, which later crashes the debugger.
3. All of the section analysis functions assume that the headers and sections data comes from a file rather than memory. This is reflected in two patterns:
- `RtlImageDirectoryEntryToData` is always called with `MappedAsImage = false`, which should be set to `true` when passing in-memory module data.
- All of the RVA section fields are rebased to file offsets. When processing in-memory data, they should be taken at face value.

The fixes are contained in this commit: https://github.com/ynwarcs/x64dbg/commit/aa0a5cd82cdb1998af2f9f020fefd86db368f2b4. After applying these fixes, the command functions as expected. As a test, I used the [MemoryModule](https://github.com/fancycode/MemoryModule) project. Unfortunately, an additional fix was needed on their side as well (see https://github.com/ynwarcs/MemoryModule/commit/10250ad4b7ee464579dfb13605427bad1626cbd9), but this is not x64dbg related.

However, before making a PR, I would like to know:
1. What is the common use case for `virtualmod`? There isn't much documentation available online, so I'm not sure what the true intentions behind the command are.
2. Are there any particular solutions to in-memory DLL loading that should work under `virtualmod`? MemoryModule was already tested, the only other publicly available solution I'm aware of is [Reflective DLL Injection](https://github.com/stephenfewer/ReflectiveDLLInjection).
3. Is the command expected to work with natively loaded modules which already appear on the list of modules?

### Steps to reproduce

[...]

PR #3131, opened 2023-07-13
Support for MemoryModule recognition and debugging via virtualmod.

mrexodia:
- Fixed some bugs in virtualmod: loadedSize was never set + the mapped data would be destroyed prematurely, causing a crash after initially parsing the module.
- Use raw RVA instead of rebasing when reading module information from memory.
- Set MappedAsImage = true when calling RtlImageDirectoryEntryToData on in-memory sections, as expected.
- This works out of the box for some basic stuff (e.g. symbols, imports, exports are loaded properly) but for full functionality a fix is needed in MemoryModule (see https://github.com/ynwarcs/MemoryModule/commit/10250ad4b7ee464579dfb13605427bad1626cbd9). TL;DR: VirtualSize field in the section header is not properly set by MemoryModule, leading to sections not being parsed properly during page sorting, leading to an error spam and some other annoying behaviour.


mrexodia:
Closes https://github.com/x64dbg/x64dbg/issues/3094

Issue #2961, opened 2022-10-16
hope to have command to save log, not only GUI 


RachelWealth:
### Feature type

Other

### Describe the feature

As command can help for third-party programme transfer

utkonos:
Yes, this would be a very helpful feature! Log control commands:

`logsave [arg1]`
this would save the log to a text file with a timestamp filename like the GUI does. If the optional `arg1` is provided, that string is used as the filename or path to save to.

`logredir [arg1]`
this would enable log redirection to a timestamp filename like the above. if the optional `arg1` is provided, that string would be used as the filename or path for the redirected log data.

`logredirstop`
stops log redirection

ElvisBlue:
Well this cound be implement by writing a plugin. x64dbg provide `_plugin_registercommand` API to add your custom command.

mrexodia:
I could swear there is some kind of API for this, but I don’t remember… why
do you need this functionality? It might be easier to have a command that
appends to a custom log file…

utkonos:
> why do you need this functionality?

The use case here is in a script. At the end of the script, it would be very useful to be able to run `logsave` right before `ret`. Another location would be in error handling also just before a `ret` and after a `msg`.

The two nice to haves: `logredir` and `logredirstop` would have a similar purpose: in scripts. One could start and stop logging during certain parts of the script where logging to a file would be useful such as during a trace or when gathering network indicators or other other specific parts of a script.

> It might be easier to have a command that appends to a custom log file…

Yes, having `logsave` and/or the other two is just an idea for how to solve the use case. I would be happy with any feature that could solve the use case.

torusrxxx:
The log redirection and saving doesn't have an external API yet, mostly because it is handled by GUI, not DBG.

utkonos:
I asked ChatGPT to write this plugin. Does this look correct?

```
#include <cstdio>
#include <cstring>

#include "x64dbg.h"

// Plugin information
const char* plugin_name = "Log to File";
const char* plugin_version = "1.0.0";

// Console command to enable logging to a file
void cmd_logtofile(int argc, char* argv[])
{
    // Check if the correct number of arguments was provided
    if (argc != 2)
    {
        dputs("Usage: logtofile [filename]");
        return;
    }

    // Get the filename from the command arguments
    const char* filename = argv[1];

    // Open the file for writing
    FILE* file = fopen(filename, "w");
    if (!file)
    {
        dprintf("Failed to open file %s for writing\n", filename);
[...]

PR #3133, opened 2023-07-15
Add Commands: LogSave, LogRedirect, LogRedirectStop

Bluefissure:
Commands are designed as proposed in https://github.com/x64dbg/x64dbg/issues/2961#issuecomment-1292878777

The only different is for `logredir [arg1]`: ~~if arg1 is not provided, it'll fail just as providing the default value in the log redirect dialog, to be compatible with the dialog behaviour.~~ arg1 is now mandatory.

Let me know if anything needs to be changed (such as the command name), this is my first time contributing.

### DIFF
--- a/src/gui/Src/Bridge/Bridge.cpp
+++ b/src/gui/Src/Bridge/Bridge.cpp
@@ -135,10 +135,7 @@ void* Bridge::processMessage(GUIMSG type, void* param1, void* param2)
         break;
 
     case GUI_REDIRECT_LOG:
-        if(!param1)
-            emit redirectLogToFile(QString());
-        else
-            emit redirectLogToFile(QString((const char*)param1));
+        emit redirectLogToFile(QString((const char*)param1));
         break;
 
     case GUI_STOP_REDIRECT_LOG:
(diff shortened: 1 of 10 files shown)


src/gui/Src/Gui/LogView.cpp
```diff
@@ -375,30 +379,56 @@ void LogView::clearLogSlot()
     this->clear();
 }
 
+void LogView::redirectLogSlotStop()
```
mrexodia, line 382: The `Slot` should be put at the end of the function name to avoid confusion (same for the other functions).

Bluefissure: I renamed to `stopRedirectLogSlot`.

src/gui/Src/Bridge/Bridge.cpp
```diff
@@ -127,6 +127,24 @@ void* Bridge::processMessage(GUIMSG type, void* param1, void* param2)
...
+    case GUI_SAVE_LOG:
+        if(!param1)
+            emit saveLog();
+        else
+            emit saveLogToFile(QString((const char*)param1));
+        break;
+
+    case GUI_REDIRECT_LOG:
+        if(!param1)
+            emit redirectLogToFile(QString());
```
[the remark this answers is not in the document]
Bluefissure: I removed the null check for `param1` since it's now mandatory.

mrexodia (approved):

mrexodia:
Thanks, I made some major changes and merged it. The only potential issue I saw is that the commands don't necessarily did their job after returning, but since all the logging is synchronized via the GUI thread this shouldn't make a difference in practice...

PR #3135, opened 2023-07-15
Use same link as found in CONTRIBUTING.md

SpraylnlPray:
Hi there,
I'm trying to get into contributing to open source projects and as I'm interested in reverse engineering I thought I might give it a go with this project.
When I was reading through `README.md` and clicked on the "good first issues" link it showed an empty list, which left me a bit puzzled. I guess the reason is that this hyperlink uses the "Easy" label to filter:
image

Luckily I found a link that works while reading through `CONTRIBUTING.md` and thought I might start off with this simple PR. Now the same link is used in `README.md` and it should apply the correct filter to anyone who clicks it.

I haven't compiled x64dbg yet as described in the list, I hope it is still fine to already open this PR. I'll give the rest of the things in the document a go as well and tried to follow the steps for creating this PR, please let me know if something is off and needs to be done differently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ You can also [compile](https://github.com/x64dbg/x64dbg/wiki/Compiling-the-whole
 
 ## Contributing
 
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks! I updated the link, but likely it’s cached. Will investigate.

PR #3147, opened 2023-07-24
add isdebuggerfocused() to determine if x64dbg is focused

torusrxxx:
The new expression function returns 1 if x64dbg is focused. It can be used in a conditional breakpoint so that you're able to break when you're playing the game but not when you're looking at CPU.
It doesn't return 1 when a plugin window is focused. Currently there's also no way to break when the debuggee is focused after 1 seconds so that it doesn't break just when it is focused.
I also evaluated approach of a system variable that changes when x64dbg is focused, but I'm concerned that event filter may impact GUI responsiveness too much. (some old games with DirectInput lock up the entire system with low-level keyboard & mouse hooks that you either don't put a breakpoint there, or disable these hooks and can't play it)


mrexodia:
Looks good to me!

PR #3149, opened 2023-07-25
Add Highlighting mode to the Graph view

shocoman:
graph_highlight_4

It should be somewhere close to 100% copy of the highlighting functionality from the disassembly view. Implemented by adding another field to the "Text" structure. Now each "Text" object also carries Zydis tokens for the instruction it represents.
I refactored this structure a bit to couple those two fields (`lines` and `lineTokens`) together. Also, basic block headers, RVAs, comments, etc are not highlightable.

A few things that I may or may not have had to do:
- Removed the raw pointer field "highlight_token" (replaced with ZydisTokenizer::SingleToken).
- Removed the structures "Token" and "HighlightToken", since they are not presently used.
- Made a full copy (or I think so) of the disassembly's highlighting behaviour even though it's weird sometimes. The highlighting mode (the red border thing) behaves differently depending on whether "Permanent highlighting mode" is enabled. Like when this "permanent" mode is on, I can copy a highlighted token just by clicking outside on an empty space with RMB, but when I do it with this option off, it just stops highlighting without any context menus. Maybe some of this behaviour is actually intentional or useful, so I added it as is.

<details><summary>GIF from CPU disasm (the first half: without the permanent highlighting, the second: with it):</summary>
<p>

permanent_highlighting_mode

</p>
</details>

### DIFF
--- a/src/gui/Src/Gui/DisassemblerGraphView.cpp
+++ b/src/gui/Src/Gui/DisassemblerGraphView.cpp
@@ -85,10 +85,7 @@ DisassemblerGraphView::DisassemblerGraphView(QWidget* parent)
     //colorsUpdatedSlot(); <-- already called somewhere
 }
 
-DisassemblerGraphView::~DisassemblerGraphView()
-{
-    delete this->highlight_token;
-}
+DisassemblerGraphView::~DisassemblerGraphView() {}
 
 void DisassemblerGraphView::resetGraph()
 {
@@ -96,7 +93,8 @@ void DisassemblerGraphView::resetGraph()
     this->ready = false;
     this->viewportReady = false;
     this->desired_pos = nullptr;
-    this->highlight_token = nullptr;
+    this->mHighlightToken = ZydisTokenizer::SingleToken();
+    this->mHighlightingModeEnabled = false;
     this->cur_instr = 0;
     this->scroll_base_x = 0;
     this->scroll_base_y = 0;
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/DisassemblerGraphView.cpp
```diff
@@ -410,7 +408,7 @@ void DisassemblerGraphView::paintNormal(QPainter & p, QRect & viewportRect, int
 
             for(Instr & instr : block.block.instrs)
             {
-                for(auto & line : instr.text.lines)
+                for(auto line : instr.text.lines)
```
mrexodia, line 411: Is the copy here intentional?

mrexodia:
Looks good in principle, thanks for taking the time to contribute! I need some more time to try and review this properly though. Meanwhile left some comments.

The context menu behavior in permanent highlighting mode is probably not desirable. The idea was to allow copying tokens directly in highlighting mode (and potentially later extend this for token-specific options like display base and enum etc), but this feature is probably never used…

src/gui/Src/Gui/DisassemblerGraphView.cpp
shocoman: Yes, it was intentional, otherwise it would permanently change the original text. But now when I think about, maybe it wasn't a good idea to do it in a painting loop function. 
The best I can come up with is to save highlighting tokens right before painting, and restore their original color afterwards. Something like this:

```c++
std::vector<std::tuple<int, QColor, QColor>> prevColors;
// highlight selected tokens
for (int i = 0; i < line.size(); i++)
{
    auto & rt = line[i];
    if(rt.text == mHighlightToken.text)
    {
        prevColors.push_back(std::make_tuple(i, rt.textColor, rt.textBackground));
        rt.textColor = mInstructionHighlightColor;
        rt.textBackground = mInstructionHighlightBackgroundColor;
    }
}
// paint
RichTextPainter::paintRichText(...);
// restore old colors
for (const auto & col: prevColors) {
    auto & rt = line[std::get<0>(col)];
    rt.textColor = std::get<1>(col);
    rt.textBackground = std::get<2>(col);
}
```

Except this isn't pretty, and I am not sure there is no better way. Anyway, just tell me what to do and I'll fix it

mrexodia: Yeah, in the disassembly this isn't an issue because the tokens are thrown away very quickly. With the graph view they are (or at least that was the intention) only fetched once. I think the 'best' solution would be to duplicate the rich text and then when the user performs a highlighting action:

- Restore the original rich text
- Apply highlighting

That way there is no copying/hackery going on during the painting.

shocoman: If by duplication you meant simply adding another field to Text, then this should be it

mrexodia:
Let's see what people think, thanks for addressing my concerns!

Issue #2234, opened 2019-10-26
Lots of accessibility issues with X64DBG


akash07k:
Dear developers and team,
First of all, I thank you all for this great tool.
It has really a great potential.
But the issue that the accessibility is majorly and badly lacking in X64DBG.
For example, there are many controls which are not having any labels such as edit ffields for typing the strings, lists containing the strings etc.
The primary issue I've faced that when we search for some strings, the list is not accessible at all and we are unable to change any string or JMP/JNE calls.
Please help.
since the application is using QT5 Widgets, please refer this link:
https://doc.qt.io/qt-5/accessible.html

akash07k:
Please refer these links in order to know that what is accessibility and why is it important:
https://developer.mozilla.org/en-US/docs/Learn/Accessibility/What_is_accessibility
https://www.w3.org/WAI/fundamentals/accessibility-intro/
https://accessibility.blog.gov.uk/2016/05/16/what-we-mean-when-we-talk-about-accessibility-2/
https://www.usability.gov/what-and-why/accessibility.html

mrexodia:
Please document specific examples you think are lacking. I’m usually not using the mouse too much and Qt should take care of most things automatically. Due to the nature of x64dbg it will never be compatible with screen readers, but the tab order and &menu item should allow you to use x64dbg.

akash07k:
> Please document specific examples you think are lacking. I’m usually not using the mouse too much and Qt should take care of most things automatically. Due to the nature of x64dbg it will never be compatible with screen readers, but the tab order and &menu item should allow you to use x64dbg.

As I said before, the list of the strings is completel  inaccessible with the keyboard and to the screen readers.

akash07k:
> Please document specific examples you think are lacking. I’m usually not using the mouse too much and Qt should take care of most things automatically. Due to the nature of x64dbg it will never be compatible with screen readers, but the tab order and &menu item should allow you to use x64dbg.

Also, the edit fields for entering the command, searching the strings are not having any labels.
We are not even able to change the JMP, JNE etc calls. that part is also completely inaccessible

akash07k:
> Please document specific examples you think are lacking. I’m usually not using the mouse too much and Qt should take care of most things automatically. Due to the nature of x64dbg it will never be compatible with screen readers, but the tab order and &menu item should allow you to use x64dbg.

The data/UI under each tab such as CPU, GRAPH, Call Stack, Sources etc is also not accessible at all.

[...]

PR #3154, opened 2023-07-29
Add accessible names

torusrxxx:
It should be able to announce the names for all views in screen reader now (#2234).


mrexodia:
Thanks!

PR #3155, opened 2023-07-29
Fix a DbgSetEncodeType crash if an out-of-date db file is used

### LINKED ISSUE #3093: DbgSetEncodeType may crash if an out-of-date db file is used
### Operating System

Windows 11 version 22H2 (OS Build 22621.1702)

### x64dbg Version

May 12 2023

### Describe the issue

The crash happens here:
https://github.com/x64dbg/x64dbg/blob/17584fcbd6a8e4bc5ed7a605488a08fa5b5b4054/src/dbg/encodemap.cpp#L244

I suspect that incorrect sizes for `EncodeMap` are restored from the db in case it was created with a different version of the executable.

### Steps to reproduce

1. Make sure you have no db files for `explorer.exe`
2. Download [explorer.exe 10.0.22000.120](https://msdl.microsoft.com/download/symbols/explorer.exe/E846E7494c4000/explorer.exe) and rename it to `explorer.exe`
3. Load it, press F9 until you reach the entry point, and mark some code as data with: End, Q, Enter
4. Download [explorer.exe 10.0.22621.1635](https://msdl.microsoft.com/download/symbols/explorer.exe/A4159E294cc000/explorer.exe) and rename it to `explorer.exe`
5. Load it, press F9 until you reach the entry point, and mark some code as data with: End, Q, Enter
6. Observe a crash

### Attachments

_No response_

m417z:
Fixes #3093

P.S. While debugging, I got the "Trying to get last error of a thread that doesn't exist!" assert and had to comment it out.

### DIFF
--- a/src/dbg/encodemap.cpp
+++ b/src/dbg/encodemap.cpp
@@ -82,7 +82,31 @@ struct EncodeMap : AddrInfoHashMap<LockEncodeMaps, ENCODEMAP, EncodeMapSerialize
 
 static EncodeMap encmaps;
 
-static bool EncodeMapGetorCreate(duint addr, ENCODEMAP & map, bool* created = nullptr)
+static bool EncodeMapValidateModuleInfo(duint key, ENCODEMAP & map, duint segsize)
+{
+    // The map size might be smaller if it was loaded from cache for another
+    // version of the module. Adjust the size in this case.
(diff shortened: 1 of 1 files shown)


mrexodia:
Look good, thanks!

PR #3156, opened 2023-07-30
Adding gzip decompression to downlibs

P1tt1cus:
Hello, this is my first time contributing. 

I was building my own symbol cache server and I noticed that x64dbg did not support/accept gzip encoding. I thought it would be good to take some time to add support for this as it dramatically improves performance when downloading PDB's from symbol servers that support this. 

I also increased the chunk size slightly to improve performance by trying to reduce the number of iterations it takes to write to the PDB.


mrexodia:
Awesome, thanks!

As a sidenote: will you publish the cache server? Seems like a useful thing to have

P1tt1cus:
Thanks for merging. Sure, I will let you know once I have posted it.

mrexodia:
For future reference (took me a while to find this because GH search sucks): https://github.com/P1tt1cus/FastSymApi

PR #3161, opened 2023-08-03
Quote image paths in the Dark theme stylesheet

shocoman:
When I place the program in a directory whose path contains an uneven number of opening and closing parentheses (like "C:/exa(mple" or "C:/e(x)am(ple"), images referenced in the css file do not load properly (and the whole file gets messed up during parsing). I guess the right way would be to escape the paths with backslashes at runtime, but then the url has to be parsed properly, and parsing is hard, who knows what might happen, etc. On the other hand, simple quoting is safe, and because `(` and `)` (as well as `\`) are not valid path characters in Windows, you can't escape from a quoted string.

Pasted image 20230803152408

### DIFF
--- a/src/gui/Src/Gui/MainWindow.cpp
+++ b/src/gui/Src/Gui/MainWindow.cpp
@@ -655,10 +655,11 @@ void MainWindow::loadSelectedTheme(bool reloadOnlyStyleCss)
     {
         auto style = QTextStream(&cssFile).readAll();
         cssFile.close();
-        style = style.replace("url(./", QString("url(%1/../themes/%2/").arg(applicationDirPath, selectedTheme));
-        style = style.replace("url(\"./", QString("url(\"%1/../themes/%2/").arg(applicationDirPath, selectedTheme));
-        style = style.replace("url('./", QString("url('%1/../themes/%2/").arg(applicationDirPath, selectedTheme));
-        style = style.replace("$RELPATH", QString("%1/../themes/%2").arg(applicationDirPath, selectedTheme));
+
+        style = style.replace("url(./", QString("url(approot:/themes/%1/").arg(selectedTheme));
+        style = style.replace("url(\"./", QString("url(\"approot:/themes/%1/").arg(selectedTheme));
+        style = style.replace("url('./", QString("url('approot:/themes/%1/").arg(selectedTheme));
+        style = style.replace("$RELPATH", QString("approot:/themes/%1").arg(selectedTheme));
         qApp->setStyleSheet(style);
     }
 
--- a/src/gui/Src/main.cpp
+++ b/src/gui/Src/main.cpp
@@ -8,6 +8,7 @@
 #include <QTextStream>
 #include <QLibraryInfo>
 #include <QDebug>
+#include <QDir>
 #include "MiscUtil.h"
 
 MyApplication::MyApplication(int & argc, char** argv)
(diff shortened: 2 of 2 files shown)


shocoman:
No, [the code](https://github.com/x64dbg/x64dbg/blob/e7995ffd5319ac0bbc26afd68f0cb07293a2e45f/src/gui/Src/Gui/MainWindow.cpp#L659C13-L659C13) should already cover this case

mrexodia:
My bad, you’re right. However, isn’t it still better to do the quoting in that code? Users are likely to not realize this is a potential issue and they won’t necessarily escape their URLs…

shocoman:
Yes, that is definitely better. By escaping, do you mean escaping the entire URL or just the "absolute path" prefix? In the latter case, we already have the string, so it's easy, and in the former case, we have to extract the path directly from the stylesheet, which can be problematic (depending on how complex Qt's CSS specification is, i.e. which characters are allowed in the path, which are only allowed if inside quotes, etc.). Plus, if users edit a theme's CSS stylesheets, they should probably already know its rules of escaping, etc

mrexodia:
I guess it would be ideal if only escaping the `$RELPATH` is possible, otherwise it would indeed require some extra work to extract the URL properly. I doubt the CSS specification is very complex though and in the worst case it's always possible to fall back to the current behavior (which is unlikely to break under most circumstances I'd say).

shocoman:
I don't think they conform to any real CSS specification. As usual, even the easiest case turns out to be not that easy.
Particularly, there are two reasons why certain paths may not be allowed:
1) They are invalidated by the parser (this one covers the parenthesis case described above; the relevant code is [here](https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qcssparser.cpp#n2928) which is called from [here](https://code.qt.io/cgit/qt/qtbase.git/tree/src/gui/text/qcssparser.cpp#n2834))
2) They are invalidated by the tokenizer (when it doesn't have a token for the provided string; the rules of Qt's tokenizer are [here](https://github.com/qt/qtbase/blob/a9982a3b637e482c52a1ca97f9e235eb302048c3/util/lexgen/css3-simplified.lexgen) )
The second clause invalidates such unpopular characters as "¢", "£", or "©". In other words, you cannot have any code point below 256 in your path, save for the usual Ascii letters and digits. The exception is if they are proper tokens. For example, you can't have "@" but can "@ABC", because the latter forms the token "ATKEYWORD_SYM"

So the solution, as terrible as it is, is to just escape everything but letters and digits. Somehow Qt allows this even inside quoted strings. The update covers "'" which is valid in Windows paths, but could be used to escape from a quoted string.

mrexodia:
If everything works now that’s great, but your comment inspired a potential other solution: https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getshortpathnamew

This will generate a DOS-style path, even for complex unicode paths and could be used as a workaround to force the path to only contain ANSI chars… I used this in the x64dbgpy plugin, because python2 didn’t support unicode paths at all.

Ill do my best to review this in the upcoming days, but I’m traveling so it might take a bit.

shocoman:
> I used this in the x64dbgpy plugin, because python2 didn’t support unicode paths at all.
By what if these short paths are disabled? They were for me.

Another update. This time using a path search prefix. So instead of writing something like "%absolute_dir_path%/my_image.png", we can register an application-wide prefix such as "approot", and then use it in all paths including CSS ones: "approot:/my_image.png". That way paths are still absolute, but Qt will do the substitution automatically after CSS parsing. Currently, I don't see why it should break

mrexodia:
Yeah if that works that’d be great! I remember trying to find something like that, but either couldn’t find it or encountered some issue. It’s definitely the cleanest solution!

mrexodia:
Cheers!

PR #3162, opened 2023-08-05
Add tips to some common actions

torusrxxx:
* Status bar tips are added to some actions supported by CommonActions
* Allow go to while not debugging in trace view


mrexodia:
Thanks 🙂

PR #3165, opened 2023-08-08
Add secondary sorting to Memory Map

### LINKED ISSUE #3010: Being restless in Memory Map tab when sorting
### Operating System

Windows 10

### x64dbg Version

Jan 25 2023

### Describe the issue

Pause debugged process, go to Memory Map tab and sort it by Protection and see result !

thx.

### Steps to reproduce

1. 
2. 
3. 

### Attachments

_No response_

HayFieldLaps:
This fixes #3010 

Fixed unstable sorting in Memory Map when sorting by Type, Protection or Initial (Protection). To resolve this, a secondary sort is now performed on Address (only in ascending order) for rows with equal values. This change affects MemoryMapView::sortRows and results in QString::compare being called twice (also in StdTable::SortBy::AsText), introducing additional computational complexity. Further optimization may be needed.

### DIFF
--- a/src/gui/Src/Gui/MemoryMapView.cpp
+++ b/src/gui/Src/Gui/MemoryMapView.cpp
@@ -195,6 +195,37 @@ void MemoryMapView::setupContextMenu()
     connect(Config(), SIGNAL(shortcutsUpdated()), this, SLOT(refreshShortcutsSlot()));
 }
 
+void MemoryMapView::sortRows(int column, bool ascending)
+{
+    auto sortFn = mColumnSortFunctions.at(column);
+    auto addressSortFn = mColumnSortFunctions.at(0);
+    std::vector<size_t> index;
+    index.resize(mData.size());
+    size_t i;
+    for(i = 0; i < mData.size(); i++)
+    {
+        index[i] = i;
+    }
+    std::stable_sort(index.begin(), index.end(), [column, ascending, this, &sortFn, &addressSortFn](const size_t & a, const size_t & b)
+    {
+        if(column > 0 && QString::compare(mData.at(a).at(column).text, mData.at(b).at(column).text) == 0)
+        {
+            auto less = addressSortFn(mData.at(a).at(0).text, mData.at(b).at(0).text);
+            return less;
+        }
+
+        auto less = sortFn(mData.at(a).at(column).text, mData.at(b).at(column).text);
+        return ascending ? less : !less;
+    });
+    auto copy1 = mData;
+    auto copy2 = mIcon;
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/MemoryMapView.cpp
```diff
@@ -195,6 +195,37 @@ void MemoryMapView::setupContextMenu()
     connect(Config(), SIGNAL(shortcutsUpdated()), this, SLOT(refreshShortcutsSlot()));
 }
 
+void MemoryMapView::sortRows(int column, bool ascending)
```
mrexodia, line 198:
```cpp
void StdTable::sortRows(int column, bool ascending)
{
    auto sortFn = mColumnSortFunctions.at(column);
    auto addressSortFn = mColumnSortFunctions.at(0);
    std::stable_sort(mData.begin(), mData.end(), [column, ascending, &sortFn](const std::vector<CellData> & a, const std::vector<CellData> & b)
    {
        auto less = sortFn(a.at(column).text, b.at(column).text);
        if(column > 0 && !less)
            less = addressSortFn(a.at(0).text, b.at(0).text));
        return ascending ? less : !less;
    });
}
```

Would something like this work too?

HayFieldLaps: No, I had tried something similar to this. The issue arises from:

[https://github.com/x64dbg/x64dbg/blob/b7347f4506e117548b5d32b23a68cd6a5eabba34/src/gui/Src/BasicView/StdTable.cpp#L18](url)
`bool StdTable::SortBy::AsText(const QString & a, const QString & b)
{
    auto i = QString::compare(a, b);
    if(i < 0)
        return true;
    if(i > 0)
        return false;
    return duint(&a) < duint(&b);
}`

Equality collisions are solved (somewhat non-sensibly) by a pointer comparison that will have different results each time the table is sorted. I made a brief attempt to solve the solution here, but the changes I tried always seemed to create issues in other widgets that also use these sort functions.

I very well could be overlooking something simple here, x64dbg's source is quite large and I've only recently begun familiarizing myself with it.

mrexodia: Oh that pointer comparison is 100% wrong, likely it was an artifact from before stable_sort was introduced.

Feel free to change that and let me know any issues that arise.

mrexodia: Oh lol, that was indeed an easier fix than I thought 😅

mrexodia:
Thanks!

Issue #1596, opened 2017-05-17
Ctrl+G "Go to expression" in a dump window, applies to the wrong dump


MulleDK19:
x64dbg v25.

Pressing Ctrl+G in the dump brings up the "Go to expression" dialog. However, it goes to the specified address in the active dump tab, and not the dump that was active (detached).

**To reproduce:**

1. Right click Dump 2 and click "Detach".
2. Click in the detached Dump 2 window to make sure it's active, then press Ctrl+G (or right click the detached dump window > Go to > Expression) and enter an address.
3. Press OK.
4. The address is navigated to in the active Dump tab, instead of the Dump in the separate window.

**Workarounds**

- Close the detached window, go to in the tab, then detach again.
- To avoid having to detach: Right click in the detached window, and click "Sync with expression", enter the address, and press OK, then right click the detached window, and click "Sync with expression", clear the text box and press OK.

mrexodia:
duplicate of #1540

MulleDK19:
@mrexodia Not a dup, and still broken. You can't go to an address in a dump windows unless they're attached. Going to an expression in a detached dump will go to the expression in the active attached dump. If all dumps are detached, the entire application crashes.

Cirn09:
This bug still exists!!! And it is really easy to reproduce~
![REPRODUCE](https://user-images.githubusercontent.com/25722111/144381983-71c8e5b4-e0e6-4dc7-aba8-535dc3fcbc95.gif)

mrexodia:
Feel free to fix it and send a pull request :)

G3ph4z:
Couldn't reproduce it in the latest version, so probably not an issue anymore.

mrexodia:
Yeah I also tried it and looks like this is fixed, thanks!

PR #3172, opened 2023-08-11
Set dump index from ACTIVEVIEW

### LINKED ISSUE #3113: The memory window does not correctly locate the address in the correct window.
### Operating System

Windows 11 Insider Preview 230609-1655

### x64dbg Version

Jun 15 2023

### Describe the issue

When I locate the memory in the detached window, it is not located in the detached memory window, but is located in the [Memory 1] window by default

### Steps to reproduce

1. Detach the memory window 2 or 3,4,5
2. locate the memory(in the current window) in the detached window and you will see, it just located in the memory 1 window.

### Attachments

_No response_

HayFieldLaps:
Fixes #3113 and #1596.

Fixes an issue with the "Follow DWORD/QWORD in Current Dump" commmand, as well as the "Go to..." expression box when a dump view is detached. It's not the most elegant but it's not terrible either.

If nativeTitle is ever changed [here](https://github.com/x64dbg/x64dbg/blob/6a970e657dc2279c9d2f584dc56cea14a2ba3e2f/src/gui/Src/Gui/CPUMultiDump.cpp#L30) this fix will break. The current user set dump view name does not affect this, nor does any language changes.

### DIFF
--- a/src/dbg/commands/cmd-gui.cpp
+++ b/src/dbg/commands/cmd-gui.cpp
@@ -36,8 +36,6 @@ bool cbDebugDump(int argc, char* argv[])
         dprintf(QT_TRANSLATE_NOOP("DBG", "Invalid address \"%s\"!\n"), argv[1]);
         return false;
     }
-    ACTIVEVIEW activeView;
-    GuiGetActiveView(&activeView);
     if(argc > 2)
     {
         duint index = 0;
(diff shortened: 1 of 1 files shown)


src/dbg/commands/cmd-gui.cpp
```diff
@@ -46,6 +48,16 @@ bool cbDebugDump(int argc, char* argv[])
         }
         GuiDumpAtN(addr, int(index));
     }
+    else if(strcmp(activeView.title, "Dump 1") == 0)
```
mrexodia, line 51: This would probably be better to do as a regex to support additional dumps, but otherwise looks good!

mrexodia (changes requested):

torusrxxx:
nativeTitle will never be changed.

src/dbg/commands/cmd-gui.cpp
```diff
@@ -47,7 +48,20 @@ bool cbDebugDump(int argc, char* argv[])
...
     }
     else
-        GuiDumpAt(addr);
+    {
+        ACTIVEVIEW activeView;
+        GuiGetActiveView(&activeView);
+        std::string titleString = activeView.title;
+        std::regex dumpPattern("Dump ([1-9][0-9]*)");
+        std::smatch matchResult;
+        if(std::regex_match(titleString, matchResult, dumpPattern))
```
ZehMatt, line 57: Does this really need regex? sscanf should do the trick with `Dump %d`

mrexodia: Yeah I’ll fix this once I’m back home.

HayFieldLaps:
Switched the code to use sscanf_s, this is much nicer than using regex. I took your request too literally haha, probably should have thought of this myself.

mrexodia:
Thanks!

justanotheranonymoususer:
Unfortunately I can still repro https://github.com/x64dbg/x64dbg/issues/1596 in version November 7 2023 of the debugger

PR #3174, opened 2023-08-14
In Graph mode, also apply highlighting to different sizes of the same register

shocoman:
Last time I forgot that different sub-registers (like rax, eax, ax, al) should be highlighted as well.

Unfortunately, it gets a bit awkward because we have to compare Zydis tokens and then apply changes to rich text. And these rich texts and tokens are not synced (rich text additionally has RVA and comments). Maybe the cleanest solution would be to store RVA and comments separately, and then, when needed, regenerate rich text from the token array and prepend/append the rest. But that would cause a lot of unnecessary object creation on every highlight change.
I've removed the duplicate rich text since the colors can be extracted directly from the tokens. At least the way they are used now.


mrexodia:
Cheers!

PR #3175, opened 2023-08-15
Add isdebuggeefocused, and optimized isdebuggerfocused

torusrxxx:
isdebuggeefocused() is added, in my opinion more useful.
It seems the best way to do it is just GetForegroundWindow().


mrexodia:
Thanks!

PR #3176, opened 2023-08-15
Added mnemonic brief column and keeps mnemonic brief configuration

torusrxxx:
The hidden by default mnemonic brief column helps keep comments column clean while showing mnemonic brief. It also remembers mnemonic brief settings.


mrexodia:
I don’t like this idea. This column is something I toggle quickly if I forget the instruction, it’s not something to keep enabled and adding it next to the comments makes it invisible unless you mess around too much…

mrexodia:
And I think I misunderstood, it looks like you can manually move the mnemonic brief to the column if you unhide it.

torusrxxx:
Newbies need to permanently enable mnemonic brief until they get used to it. The mnemonic brief is moved to the new column, so that the comments column stay clean and you can also make mnemonic brief column narrow to see only a few words, compared to always show mnemonic brief in full currently. There's no change to the default GUI.

mrexodia:
Will do some merging when I'm back home tomorrow!

mrexodia:
👍🏻

PR #3180, opened 2023-08-21
Show a tooltip with a sorted list of flags in the Register View

shocoman:
Pasted image 20230821142338

Sometimes programs work with eflags directly (like checking bit 8 to see if the trap flag is set).
In such moments it's nice to have a list of all the flags with their respective bits to quickly understand which flag it is or is not.
Currently, even though the bit numbers are shown in tooltips, the registers themselves aren't sorted in any way, so it takes some time to find the right flag (if you don't remember them exactly).

### DIFF
--- a/src/gui/Src/Gui/RegistersView.cpp
+++ b/src/gui/Src/Gui/RegistersView.cpp
@@ -1302,6 +1302,35 @@ QString RegistersView::helpRegister(REGISTER_NAME reg)
     switch(reg)
     {
     //We don't display help messages for general purpose registers as they are very well-known.
+    case EFLAGS:
+    {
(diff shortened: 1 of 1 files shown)


mrexodia:
I definitely like the idea, but I think it would be important to also include the (hex) mask for each flag.

shocoman:
Pasted image 20230823171131

Now it looks like this. The table in the tooltip is styled with html, since the tooltip's font is not monospaced to make one properly in text. I went with the 'h' suffix instead of '0x' for the mask to avoid leading zeroes, because they make it a bit harder to distinguish

mrexodia:
Looks awesome, thanks!

PR #3183, opened 2023-08-22
CTRL + C functionality in AbstractStdTable/SearchListView

HayFieldLaps:
Related issue: #2438 

-CTRL + C functions properly in SearchListView
-CTRL + A when search box is empty selects all rows in table (if allowed) 
-SHIFT + HOME/END when search box is empty selects rows from table (if allowed)
-ESC clears the filter rather than just the Search edit line


mrexodia:
Thanks for your contribution!

PR #3185, opened 2023-08-23
Using reserve and unordered_map for optimize inserts

GermanAizek:
I am sure, that not all fixes will be needed, because there are cases when count elements does not exceed 3 for avg user x64dbg.

### DIFF
--- a/src/dbg/analysis/recursiveanalysis.cpp
+++ b/src/dbg/analysis/recursiveanalysis.cpp
@@ -149,6 +149,7 @@ void RecursiveAnalysis::analyzeFunction(duint entryPoint)
     CFGraph graph(entryPoint);
     UintSet visited;
     std::queue<duint> queue;
+    visited.reserve(queue.size());
     queue.push(graph.entryPoint);
     while(!queue.empty())
     {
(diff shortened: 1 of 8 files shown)


src/dbg/analysis/recursiveanalysis.cpp
```diff
@@ -39,7 +40,8 @@ void RecursiveAnalysis::SetMarkers()
         // This algorithm orders basic blocks and then iterates, growing the chunk downwards
         // Function ranges are collected in another ordered map for loop insertion
         const auto & function = functionItr.second;
-        std::map<Range, const BridgeCFNode*, RangeCompare> blockRanges, functionRanges;
```
mrexodia, line 42: This has to be an `std::map` because of the way the `Range` works. It allows a lookup of a value _inside_ the range vs the exact range.

mrexodia (changes requested):

mrexodia:
To fix the CI you can run `format.bat` before you commit.

src/dbg/analysis/recursiveanalysis.cpp
mrexodia: thanks!

mrexodia:
Thanks for the contribution!

PR #3199, opened 2023-08-29
Pass Escape key event if search text is empty

HayFieldLaps:
Was blocking Esc from closing Attach window.

Bug I introduced in #3183.


mrexodia:
Thanks for following up on that!

PR #3201, opened 2023-08-30
removed extra dot when saving trace dump

### LINKED ISSUE #3181: QFileDialog filter contains an extra period when saving a trace dump
Pasted image 20230821180447

As a result, the dialog won't show other dumps by default. I think the problem is [here](https://github.com/x64dbg/x64dbg/blob/a85f17756791a63ed281f15a1c312510b0339d3a/src/gui/Src/Tracer/TraceBrowser.cpp#L90)

f09f9095:
resolves #3181


mrexodia:
Thanks, appreciated!

PR #3203, opened 2023-09-02
Add UI for mnemonichelp

torusrxxx:


mrexodia:
Isn’t this already in the context menu for the disassembly?

torusrxxx:
This one prompts for user input so that you can search for an instruction that isn't in disassembly.

PR #3206, opened 2023-09-04
Fixed cycling through filtered references without clearing filter.

HayFieldLaps:
Allow "Actions -> Go to Next/Previous Reference" to cycle through a filtered reference list.  Was previously clearing the filter on the first "Next/Previous" action.


mrexodia:
Cheers!

Issue #2735, opened 2021-09-19
GUI request


Systemhlp:
### Feature type

Debugging

### Describe the feature

Hello 
I think , it is better there is a window frame at disassembly portion.
Look below image.
![1](https://user-images.githubusercontent.com/87657997/133929556-46a659b5-5e12-4cab-9534-7d0259555273.png)
Thank you and good day !

torusrxxx:
What do you actually want to do if the feature is implemented? I think if you want to inspect the code you can double click to go there.

ZehMatt:
I think the author is asking to have a border/frame around the preview to better separate it.

torusrxxx:
The default theme already has the border. Just dark theme has the issue.

mrexodia:
Yeah, I ran into this issue while working on the dark theme with JustMagic.
Unfortunately I couldn’t find a workaround for it so I just left it like
this 🤷🏻‍♂️

PR #3208, opened 2023-09-07
Dark theme: fix several visibility issues

shocoman:
There are several places where widgets (mostly text) are hard to see due to their dark color.

**Disassembly pop up border**

before/after:
image

On default theme, the disassembly popup window has proper border. 
For some reason (probably intentionally), this standard 'frame' border completely disappears after trying to set the 'border' attribute in the stylesheet (the current dark theme stylesheet does it for all QWidgets).
Another problem is that this widget is custom drawn, so even though we can set the border via CSS, the custom loop will draw on top of it.
The problem is solved by additionally drawing a thin frame around the popup. Then this frame can be styled normally or made transparent via CSS.

Accidentally fixes issue #2735

---

**Hardware breakpoint over CIP color**

image

The "before" case is completely unreadable. 
I changed the breakpoint background color to black (it was transparent). Looks weird, but adds a little bit of contrast

---

**Context menu border color**

image

Aside from that tiny streak of shadow on the bottom-right side, the context menus essentially have no visible boundaries and, as a result, blend with background completely. Somehow it didn't bother me until recently.
Separators are also 'fixed'. 
Overall, it looks very similar to the standard Windows 10's dark mode

---

**Button outline**

image

As usual, the borders are barely visible. Especially when compared to those checkboxes nearby.

An unwanted side effect is that the strangeness of these custom table headers is now much more noticeable. For some reason, they are drawn as PushButtons.

image

---

**Table/List text selection color**

image

Black color doesn't play well with dark background

---

**Font tab in the Appearance dialog**

image

--- 

**System breakpoint scripts**

image

And a small brightness increase for disabled QLineWidgets and text


ZehMatt:
Those changes look great

ZehMatt (approved):

PR #3209, opened 2023-09-07
Fix typo in bookmarkclear command

HayFieldLaps:


mrexodia:
Thanks, I guess nobody ran that function before 😅

PR #3212, opened 2023-09-12
Replaced pushes and inserts to emplace for compact code

GermanAizek:


src/dbg/commands/cmd-searching.cpp
```diff
@@ -1021,7 +1021,7 @@ bool cbInstrGUIDFind(int argc, char* argv[])
                 //very likely a GUID
```
captain-hoity-toity, line 1021: To ensure code reusability, this logic might fit well in a base class or mixin.

mrexodia:
Thanks for the refactor, really appreciate it!

Issue #3210, opened 2023-09-08
"File does not exist" when trying to relaunch file from network drive


TheOnAndOnlyZenomat:
### Operating System

Windows 10, Build 19045.3324

### x64dbg Version

Sep 2 2023

### Describe the issue

When trying to "Restart" an exe loaded from a network drive it fails with "File does not exist". This also occurs, when opening the file dialog, which opens in the last used folder (the network drive), and directly clicking on the exe.
To get it to reload I have to reselect the network drive and traverse the entire file hierarchy to the exe.

There seems to be an issue from 2015 (https://github.com/x64dbg/x64dbg/issues/147) with the same issue, but states, that it should already be fixed.

If I can provide more logs or other assistance, please let me know, I am happy to do some debugging.

### Steps to reproduce

1. Load an exe from a network drive
2. click the reload button / open the file dialog and select the exe
3. "File does not exist" in the bottom left corner

PR #3217, opened 2023-09-20
commandparser handle UNC paths

HayFieldLaps:
Solves #3210 .

I'm not sure if this is actually an optimal fix. It had seemed that the issue was being caused by escaped double quotes:

`init \\desktop-name\Network\test.exe`

This worked fine.

`init "\\desktop-name\Network\test.exe"`

Would result in the first backslash being parsed out.

Despite that, the double quotes parsing code looks fine. Maybe I'm missing something here though.

### DIFF
--- a/src/dbg/commandparser.cpp
+++ b/src/dbg/commandparser.cpp
@@ -78,9 +78,6 @@ Command::Command(const String & command)
             case '{':
                 dataAppend(ch);
                 break;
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks dad 😂

image

mrexodia:
I changed the implementation by the way, but the result should be the same.

PR #3219, opened 2023-09-21
Add support for plugins in sub directories

ZehMatt:
Some plugins may contain multiple other libraries so the plugin folder becomes quite messy over time. This adds support to load plugins from a sub directory, ex.: `x64/plugins/myplugin/myplugin.dp64`.

The directory name MUST match the plugin name in order to work. I also refactored the way how pluginload and pluginunload works, one does not have to use `pluginload myplugin/myplugin.dp64` for it to work, providing just the plugin name is now the preferred way, plugins will also load when `pluginload myplugin.dp64` is used even when its from a sub directory. I personally think that the requirement of having the same name is sensible, having that not a requirement would add way more complexity and the command pluginload would have to always specify the relative path.

This PR also fixes x64dbg being able to load the same plugin multiple times due to `isLoaded` never being copied to the plugin list. Another thing this PR fixes is pluginunloadall not clearing the actual plugin list and only unloading the dll, probably not an actual issue as this happens only when x64dbg closes, however now one could add a command to unload them all and it would work.

In case the requirement for the folder name is too strict I'm open for suggestions, this also means only one plugin can be loaded from the directory, this is just a means to better organize plugins with lots of dependencies. All existing plugins will also not be affected by this PR.

### DIFF
--- a/src/dbg/plugin_loader.cpp
+++ b/src/dbg/plugin_loader.cpp
@@ -27,7 +27,7 @@ static std::wstring gPluginDirectory;
 /**
 \brief The current plugin handle.
 */
-static int gCurPluginHandle = 0;
+static int gCurPluginHandle = 1;
 
 /**
 \brief List of plugin callbacks.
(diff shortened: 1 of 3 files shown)


mrexodia:
Thanks!

PR #3224, opened 2023-09-22
Add some cross platform example projects

mrexodia:
Still a work in progress, just some proof of concept code to show that (some of) the `BasicView` components work on Mac/Linux (and with Qt6)

- [x] Replace vendored dependencies with `fetch-content` to not bloat the repository unnecessarily
- [x] Set up GitHub Actions to make sure things stay compatible


torusrxxx:
What are your plans on this? I will be on holiday soon, do you think I should quickly make dump in trace ready to merge so that you can start modifying the dump and disassembly there to use your new headers?

mrexodia:
Yeah would be good if we could merge the new trace features (at least hidden behind a setting if it’s not fully ready yet).

I’m not in a big hurry with this though, just working on it from time to time. I wanted to open an issue/discussion about this with more details, but the plan is to slowly refactor the core UI components to be completely independent of x64dbg, so they can be reused for other projects. Additionally I want to merge back the improvements to the graph view that Cutter made, but that will probably be after this.

thug-shaker:
```math
```

PR #3226, opened 2023-09-22
Hide Keep Size QCheckBox from all but Binary Edit

HayFieldLaps:
Removes the "Keep Size" checkbox from the following dialogs. It's nonsensical having them here nor were they doing anything.

- CPUDisassembly->Find Pattern
- CPUDump->Find Pattern
- CPUStack->Find Pattern


mrexodia:
Thanks, sorry for the delay!

PR #3239, opened 2023-10-11
Update HexEditDialog.ui

stdust:
HexEditDialog's tab orders have been assigned to code for less frequently used codepages and to keep size, etc so unnecessary tab orders should be deleted

### DIFF
--- a/src/gui/Src/Gui/HexEditDialog.ui
+++ b/src/gui/Src/Gui/HexEditDialog.ui
@@ -56,7 +56,7 @@
               <string/>
              </property>
              <property name="pixmap">
-              <pixmap resource="../../resource.qrc">:/Default/icons/fatal-error.png</pixmap>
+              <pixmap>:/Default/icons/fatal-error.png</pixmap>
              </property>
             </widget>
            </item>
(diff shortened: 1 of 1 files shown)


stdust:
I tested the compiled file and I see an error. 
This is how the tab order moves     " Asc -> unicode -> last code page -> hex "
The tab order works correctly until last code page   but "Last code page -> hex" requires 2 tab to go
I'm not familiar with the qt program, so I can't solve the problem I need help from anyone

torusrxxx:
Are you sure there are no tab stops even for OK and Cancel buttons?
I'm a bit oldschool. The tab order can be optimized, but I think tab stops are there to help people using a software without mouse.

mrexodia:
Yeah I agree, all the controls should be accessible using the keyboard

stdust:
OK and Cancel buttons do not need tab ordering 
There are two hotkeys for Ok (Enter and Alt+O) and Cancel (Esc Alt+C).
Pressing tab multiple times to go to the OK button and pressing enter is the same as pressing enter immediately.

stdust:
> Yeah I agree, all the controls should be accessible using the keyboard

I agree 100% with your comment 

It would be nice to have a tab oder on the CPU window to access all controls without a mouse. 
We need a tab order that cycles through Disassembly view, register, stack, and dump

mrexodia:
Thanks, we'll see how it's received.

PR #3240, opened 2023-10-12
Find Pattern searches now begin at start of current region

HayFieldLaps:
It didn't make sense for searches to begin from the current selection, nor was this obvious from the UI, maybe this was a bug.
The start/end of the search ranges are dependant on the users Memory Map view (Section or Region view), perhaps this should be made clear to the user somehow.

Disassembler: Search for -> Current Region -> Pattern (Ctrl + B)
Previous range: Selection -> End of Region
New range: Start of Region -> End of Region
Removed "Entire Block" checkbox

Dump: Find Pattern... (Ctrl + B)
Previous range: Selection -> End of Region
New range: Start of Region -> End of Region
Removed "Entire Block" checkbox

Stack: Find Pattern... (Ctrl + B)
Previous range: Selection -> End of Region
New range: Start of Region -> End of Region
Removed "Entire Block" checkbox

Memory Map: Find Pattern... (Ctrl + B)
Search through all selected regions (if more than one is selected)

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -1217,7 +1217,6 @@ void CPUDisassembly::findCallsSlot()
 void CPUDisassembly::findPatternSlot()
 {
     HexEditDialog hexEdit(this);
-    hexEdit.showEntireBlock(true);
     hexEdit.isDataCopiable(false);
     hexEdit.mHexEdit->setOverwriteMode(false);
     hexEdit.setWindowTitle(tr("Find Pattern..."));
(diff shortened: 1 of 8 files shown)


mrexodia:
Generally this is how search works in text documents/hex editors, so I don't think removing the behavior altogether is a good idea. How about renaming the (confusingly named) 'Entire block' checkbox to 'Start from selection' and keeping the current behavior per default (eg invert the logic everywhere)? That preference should be remembered in the INI, so the user can choose what mode they prefer for individual views.

If you check something like HxD/010 Editor the search is also from the current selection, so it's a bit tricky. I do agree that the 'Current Region' naming is confusing, but I think that's fine if the checkbox is named 'Start from selection'

HayFieldLaps:
I didn't realize that this behavior was an industry standard. Just renaming the option is a fine solution. I've made all the changes you've recommended.

I left the 'Entire Block' checkbox as it does make sense in the context of memory map's 'Find Pattern...' dialogue.

mrexodia:
Thank you!

PR #3250, opened 2023-10-25
Fixed a bug in the Edit Breakpoint dialog that might erase double quotes

### LINKED ISSUE #3248: String in the expression loses the double quotes
### Operating System

Windows 10

### x64dbg Version

Oct 5 2023

### Describe the issue

When you edit the breakpoint again, the string in the expression loses the double quotes.

### Steps to reproduce

step1: Edit a breakpoint for the first time, just `"D:"`
step2: Correct display of `"D:"`
step3: Edit the breakpoint again, but we haven't changed any of the data.
step4: `"D:"` becomes `D:`. The string's double quotes are missing, which invalidates the expression.

### Attachments

step1
step2
step3
step4

shocoman:
Should fix #3248

As I can see, one of the few things that this dialog does is escape double quotes. But it's done in the methods like `on_editLogText_textEdited`, which are obviously called only when something is actually updated. So if the user doesn't touch the text fields, they don't get escaped. The solution is to make this escaping mandatory as soon as the OK button is pressed.

Somewhat unrelated, I also made the window resizable horizontally. If I remember correctly, the maximum size of those text fields is 256 bytes, but the window has room only for ~42 characters by default.

### DIFF
--- a/src/gui/Src/Gui/EditBreakpointDialog.cpp
+++ b/src/gui/Src/Gui/EditBreakpointDialog.cpp
@@ -35,6 +35,8 @@ EditBreakpointDialog::EditBreakpointDialog(QWidget* parent, const BRIDGEBP & bp)
     setWindowIcon(DIcon("breakpoint"));
     loadFromBp();
 
+    connect(this, SIGNAL(accepted()), this, SLOT(acceptedSlot()));
+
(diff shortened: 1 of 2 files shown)


mrexodia:
Cheers!

PR #3251, opened 2023-10-26
Make error messages clearer if x96dbg fails to open a file

lhmouse:
When x96dbg failed to open a file, it displayed a message box with the path as the message and a reason string as the caption. This looks weird, as the caption conveys useful information but it can usually be truncated.

We should display the reason string as the message, and the path as the caption instead.


mrexodia:
Makes sense, thanks!

Issue #3163, opened 2023-08-05
Long symbol name can make goto dialog wider than screen width


torusrxxx:
Long symbol name can make goto dialog wider than screen width because the symbol name is displayed in the label. Then it is impossible to click OK and Cancel buttons.

HimaTeju:
Can I get a screenshot or video of where this issue is happening, more details??

mrexodia:
A program like this should reproduce the issue:

```cpp
[...]

PR #3260, opened 2023-10-29
#3163: Long symbol name

foralost:
Had a crash course regarding Qt GUI...

The window now does not want to increase in size (somehow?, maybe by reducing the label width it got force shortened?), I have added max values to prevent expanding too much if it would happen.

Added a functionality to split a long symbol name with break lines based on the width of the editExpression.

Solves #3163 .


mrexodia:
Thanks!

Issue #3232, opened 2023-10-03
Name a thread from call stack view


torusrxxx:
Call stack view shows an overview of what each thread is doing. However users can't name a thread from there. The thread names are also not shown in call stack view.
Sometimes there are too many threads created by the system. The user should be able to mark these threads as system threads and then hide them.

torusrxxx:
![3232](https://github.com/x64dbg/x64dbg/assets/15761310/aa50f43c-77ed-4055-9334-b152659aa5cb)

torusrxxx:
I think a better first step would be adding "follow in threads view" here.

foralost:
So:
1. Adding a thread name on horizontal line near the number of the thread (if defined).
2. Adding an option to the right mouse button menu while in Call Stack view:
   - Follow in Thread View
   - Rename Thread
 

right?

PR #3262, opened 2023-10-30
#3232: Providing Rename Thread and etc. for CallStack View

foralost:
Adding `Follow in Threads` and `Rename Thread` option to context menu.
image

Solves #3232

### DIFF
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -1152,6 +1152,7 @@ typedef enum
     GUI_GRAPH,
     GUI_MEMMAP,
     GUI_SYMMOD,
+    GUI_THREADS,
 } GUISELECTIONTYPE;
 
 #define GUI_MAX_LINE_SIZE 65536
@@ -1281,6 +1282,7 @@ typedef enum
     GUI_SAVE_LOG,                   // param1=const char* file name,param2=unused
     GUI_REDIRECT_LOG,               // param1=const char* file name,param2=unused
     GUI_STOP_REDIRECT_LOG,          // param1=unused,               param2=unused
+    GUI_SHOW_THREADS,               // param1=unused,               param2=unused
 } GUIMSG;
 
 //GUI Typedefs
@@ -1423,6 +1425,7 @@ BRIDGE_IMPEXP void GuiMenuSetName(int hMenu, const char* name);
 BRIDGE_IMPEXP void GuiMenuSetEntryName(int hEntry, const char* name);
 BRIDGE_IMPEXP void GuiMenuSetEntryHotkey(int hEntry, const char* hack);
 BRIDGE_IMPEXP void GuiShowCpu();
+BRIDGE_IMPEXP void GuiShowThreads();
 BRIDGE_IMPEXP void GuiAddQWidgetTab(void* qWidget);
 BRIDGE_IMPEXP void GuiShowQWidgetTab(void* qWidget);
 BRIDGE_IMPEXP void GuiCloseQWidgetTab(void* qWidget);
--- a/src/dbg/commands/cmd-gui.cpp
+++ b/src/dbg/commands/cmd-gui.cpp
@@ -8,6 +8,60 @@
 #include "value.h"
 #include "variable.h"
 
+bool cbShowThreadId(int argc, char* argv[])
+{
+    duint threadID = 0;
+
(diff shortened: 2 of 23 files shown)


foralost:
I would need help regarding switching to `Threads` view, because I have found that `Disassembly` window is quite supported, but that one is not.

foralost:
And also the CI says that my new-added files are nonconforming, but locally it says otherwise:
image

mrexodia:
Most likely the reason is your CRLF settings. To fix run: `git config --global core.crlf false`

foralost:
@mrexodia  nah, it did not work :< , git diff shows nothing with enabled whitespaces... im using CRLF.
(I have reset branch and committed again, maybe setting everything from ground up will do the trick).

I will try to investigate the CPP solution and study the disasm command on the weekend to find out how the Dissasembly window is called. Probably there will be a new command to switch to the Threads window, but will see.

mrexodia:
The core.autocrlf option automatically converts the whitespaces in the commit. The files still stay CRLF on your local checkout. I’ll take a look tomorrow, but this is the only reason I can think of…

mrexodia:
Just confirmed, the newly-created files `ThreadGUIUtils.cpp` and `ThreadGUIUtils.h` use LF line endings.

mrexodia:
I think in general it is not necessary to create new files for this functionality. Simply copying the code in both places will work just fine.

mrexodia:
Ah I just remembered there is a `SimpleInputBox` function in `MiscUtil.h` that does pretty much exactly what you want.

foralost:
@mrexodia those god damn newlines... 

Ok, then on weekend I will continue progress, thanks!

foralost:
Added the `showthreadid` command, which switches to the `Threads` view (with also implementing new messages for switching to this view). Also used in the new context menu.

Used `MiscUtils.h` as suggested for both locations.

If everythings works ok, I will fix alphabetical order in function names etc.  Plz test.

src/bridge/bridgemain.h
```diff
@@ -1149,6 +1149,7 @@ typedef enum
     GUI_DISASSEMBLY,
     GUI_DUMP,
     GUI_STACK,
+    GUI_THREADS,
```
mrexodia, line 1152: You cannot insert enums for binary compatibility. Please append it to the end

src/bridge/bridgemain.h
```diff
@@ -1281,6 +1282,8 @@ typedef enum
     GUI_SAVE_LOG,                   // param1=const char* file name,param2=unused
     GUI_REDIRECT_LOG,               // param1=const char* file name,param2=unused
     GUI_STOP_REDIRECT_LOG,          // param1=unused,               param2=unused
+    GUI_SELECT_THREAD_BY_ID,        // param1=ThreadID
```
mrexodia, line 1285: I think the existing “GuiSelectionSet” message can be used for this

src/dbg/commands/cmd-gui.cpp
```diff
@@ -8,6 +8,36 @@
 #include "value.h"
 #include "variable.h"
 
+bool cbShowThreadId(int argc, char* argv[])
```
mrexodia, line 11: This should be merged with the existing command: https://github.com/x64dbg/x64dbg/blob/development/src/dbg/commands/cmd-thread-control.cpp#L49

mrexodia (changes requested):
In general I think the selection api messages can be reused, otherwise looks mostly fine. Thanks!

src/bridge/bridgemain.h
foralost: Appended to the end.

mrexodia:
Thanks for fixing the suggestions, wasn't home this weekend but will try to merge it asap

foralost:
The thread ID should be provided in HEX or in DEC? Cause I saw in  #3271 that there [ToHex is used](https://github.com/x64dbg/x64dbg/pull/3271/files#diff-8d202accfd12cde659d9fd3c54531ec206ff0672e1e9973bdfa39a5d71a92603R267).

mrexodia:
There is a GUI option to choose, but I think we’ll be removing it. I think there is a helper function in StringUtils to format it automatically though…

mrexodia:
Thank you!

PR #3271, opened 2023-11-13
Auto switch thread when following callstack

torusrxxx:
#3144


mrexodia:
Thanks!

torusrxxx:
Seems this is broken because it always jumps to EIP instead of staying at the callstack selected by the user after following

PR #3272, opened 2023-11-14
Fixed get count threads for multi-cpu system with NUMA architecture

GermanAizek:
Fixed very old problem on any Windows NT and modern Windows Server 😆 

https://developercommunity.visualstudio.com/t/hardware-concurrency-returns-an-incorrect-result/350854

https://stackoverflow.com/questions/31209256/reliable-way-to-programmatically-get-the-number-of-hardware-threads-on-windows

Why this commit is useful not only for server configurations, now a very cheap PC configuration is building from Xeon E54xx, X34xx, E3-xxxx, E5-xxxx, E7-xxx, any Silver, any Gold, any Platinum series, cheapest on LGA 2011v3 socket two-socket board with NUMA support is cheap on Alibaba, Baidu or Aliexpress and Amazon.

Examples:
https://www.alibaba.com/product-detail/DDR4-x99-dual-cpu-Lga2011-V3_1600443686429.html
https://www.amazon.com/Desktop-Motherboard-Gigabit-LGA2011-SATA3-0/dp/B0CC6KSTXC

### DIFF
--- a/src/dbg/_global.cpp
+++ b/src/dbg/_global.cpp
@@ -399,49 +399,49 @@ void WaitForMultipleThreadsTermination(const HANDLE* hThread, int count, DWORD t
 // This implementation supports both conventional single-cpu PC configurations
 // and multi-cpu system on NUMA (Non-uniform_memory_access) architecture
 // Original code from here: https://developercommunity.visualstudio.com/t/hardware-concurrency-returns-an-incorrect-result/350854
-// Modified by GermanAizek
-duint GetThreadCount() noexcept
+duint GetThreadCount()
 {
+    duint threadCount = std::thread::hardware_concurrency();
+
+    typedef BOOL(*WINAPI GetLogicalProcessorInformationEx_t)(
+        LOGICAL_PROCESSOR_RELATIONSHIP,
+        PSYSTEM_LOGICAL_PROCESSOR_INFORMATION_EX,
+        PDWORD
+    );
+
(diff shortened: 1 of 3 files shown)


GermanAizek:
@mrexodia I think this is a very useful commit since x64dbg only works on Windows platform, I accidentally noticed that my server does not use all core four processors on motherboard, but only one processor worked on 100%.

GermanAizek:
Ok, I'll fix it now and write detail comment for function.

GermanAizek:
@mrexodia fix it https://github.com/x64dbg/x64dbg/pull/3272/commits/380d58345ae6c4bf2e985122b336318a2348ed43

GermanAizek:
@mrexodia fix it https://github.com/x64dbg/x64dbg/pull/3272/commits/eeab4c47ed07b8483509545ee972c27f382311ef

GermanAizek:
when I build in msvc, I don't get error about noexcept keyword, in CI it exists.

mrexodia:
Yeah it's because x64dbg uses Visual Studio 2013 and targets Windows XP. The code is extremely 'modern c++' and written a bit convoluted, I'll take a look at it.

mrexodia:
Thanks for your contribution!

GermanAizek:
@mrexodia Sorry, I started fixing similar issue in Clang, MinGW and other Windows compilers. Problem with getting count threads is that it turns out just getting count threads up to Windows 11 is not enough, but for Win11 and newer it works. Win32 developers in LLVM-project team told me that it is necessary for all cores to configure cpu_affinity (if core > 64), perhaps I will supplement this function by checking Windows version and calling cpu_affinity function, if I can make bug fix in compilers, then I will change macro to detecting compiler version.

PR #3275, opened 2023-11-21
fix No such signal in graph view

torusrxxx:


mrexodia:
Nice catch, thanks! I wonder why I didn't see this in the log...

PR #3277, opened 2023-11-24
Optimize select/deselect all in Page Rights UI

ZehMatt:
It took ages for the dialog to show up when the section is large. The select all implementation was also very inefficient, now its all pretty much instant.

~~The change in MemGetPageRights is maybe not the most optimal thing, the change will cause it to prefer the cache if the debugger is paused which should be safe in like 99% unless something modifies it externally, in most cases the memory pages are updated when the user modifies anything from within x64dbg.~~

Can't use the cache as when bListAllPages is disabled its not allowed to split which would lead to having the wrong information in the page list, ideally the UI memory map builds its own view and the actual memory map reflects the actual pages.


mrexodia:
Cheers!

PR #3278, opened 2023-11-25
#3051: Follow a thread from handles window

### LINKED ISSUE #3051: Switch to the thread of a window from the handles view

foralost:
Added the optionality to Follow a thread from Handles window.
I also have added a separate column for Thread ID to prevent a collision with thread names.

image

Closes #3051 .


src/gui/Src/Gui/HandlesView.cpp
```diff
@@ -42,12 +42,13 @@ HandlesView::HandlesView(QWidget* parent) : QWidget(parent)
     mWindowsTable->addColumnAt(8 + 8 * charWidth, tr("Handle"), true, "", StdTable::SortBy::AsHex);
     mWindowsTable->addColumnAt(8 + 120 * charWidth, tr("Title"), true);
     mWindowsTable->addColumnAt(8 + 40 * charWidth, tr("Class"), true);
-    mWindowsTable->addColumnAt(8 + 8 * charWidth, tr("Thread"), true, "", StdTable::SortBy::AsHex);
+    mWindowsTable->addColumnAt(8 + 16 * charWidth, tr("Thread ID"), true, "", StdTable::SortBy::AsHex);
+    mWindowsTable->addColumnAt(8 + 32 * charWidth, tr("Thread"), true);
```
mrexodia, line 46: Adding a new column would break existing user's UI settings (since it restores the size/order of the wrong column). In this case I don't think it's necessary either because it just clutters the view.

src/gui/Src/Gui/HandlesView.cpp
```diff
@@ -317,11 +321,26 @@ void HandlesView::disableWindowSlot()
...
 
+
 void HandlesView::followInDisasmSlot()
 {
     DbgCmdExec(QString("disasm %1").arg(mWindowsTable->mCurList->getCellContent(mWindowsTable->mCurList->getInitialSelection(), 0)));
 }
 
+void HandlesView::followInThreads()
+{
+    QString threadID = mWindowsTable->mCurList->getCellContent(mWindowsTable->mCurList->getInitialSelection(), 4);
```
mrexodia, line 332: Here you can use `getCellUserdata` to store a `duint` (thread id) for a column. Just have to populate it when the handles are being refreshed.

src/gui/Src/Gui/HandlesView.cpp
```diff
@@ -317,11 +321,26 @@ void HandlesView::disableWindowSlot()
...
+
+    if(Config()->getBool("Gui", "PidTidInHex"))
+    {
+        duint threadIDnum = strtoll(threadID.toUtf8().constData(), NULL, 0x10);
+        threadID = QString::number(threadIDnum);
+    }
+
+    DbgCmdExec(QString("showthreadid " + threadID));
```
mrexodia, line 340: Doing `DbgCmdExec(QString("showthreadid %1").arg(threadID)` is a bit clearer to read. Also the casing of the variable should be `threadId` to keep things consistent

src/gui/Src/Gui/HandlesView.cpp
```diff
@@ -118,6 +119,8 @@ HandlesView::HandlesView(QWidget* parent) : QWidget(parent)
     mActionFollowProc = new QAction(DIcon(ArchValue("processor32", "processor64")), tr("Follow Proc in Disassembler"), this);
     connect(mActionFollowProc, SIGNAL(triggered()), this, SLOT(followInDisasmSlot()));
     mActionFollowProc->setShortcut(Qt::Key_Return);
+    mActionFollowThread = new QAction(DIcon("arrow-threads"), tr("Follow Thread in Threads window"), this);
```
mrexodia, line 122: I think it should be named `Follow in Threads` to stay consistent with the call stack:

image

mrexodia (changes requested):
Looks good mostly, just think it can be simplified. Mostly stylistic changes, but if you don't have time/will to adjust I will probably get to it in the next few days...

src/gui/Src/Gui/HandlesView.cpp
foralost: Where can I find the `getCellUserData`?

Maybe I could do the same approach as CallStackView: {Thread Name} - {Thread Id}

src/gui/Src/Gui/HandlesView.cpp
foralost: Added icons for CallStackView and adjusted the names

src/gui/Src/Gui/HandlesView.cpp
foralost: Name of variable adjusted and used `.args(threadId)`

src/gui/Src/Gui/HandlesView.cpp
foralost: Column removed.

src/gui/Src/Gui/HandlesView.cpp
mrexodia: Didn't forget about this, locally I did the plumbing to add support for `getCellUserdata` to the `SearchListView` but something isn't working right and I need some time to debug it...

foralost: Ok, in this case the `getCellUserdata` works, but the command `showthreadid` works for thread numbers in DEC (as displayed on menu title), not for HEX (as the code is putting in that way).

I will add support for HEX numbers for `showthreadid` command in another MR. 

Is somewhere else the function failing?

foralost:
I will also update the Call Stack View, but after `xxxCellUserdata` will be merged with master branch.

src/gui/Src/Gui/HandlesView.cpp
mrexodia: Hm probably the command should use valfromstring to parse the arguments to numbers right?

foralost: Theres a plain strtoll 😄 😄 😄 , I do not remember why I did not use it, probably I thought that it would be an overtake (we would need only numbers after all). I remember I also had troubles with HEX vs DEC. 

I could replace it with `valfromstring `.

mrexodia: Yeah please do, the default radix is hex that’s why there is this ToHexPtr function to format the numbers for commands

foralost: Done

src/dbg/commands/cmd-gui.cpp
```diff
@@ -10,55 +10,21 @@
...
-    DbgGetThreadList(&threads_head);
-    GuiUpdateThreadView(); // To be sure that we have the freshest stuff
 
-    bool threadFound = false;
-    uint32_t row;
-    for(row = 0; row < threads_head.count; row++)
-    {
-        THREADALLINFO currThread = threads_head.list[row];
-        if(currThread.BasicInfo.ThreadId == threadID)
+        SELECTIONDATA newSelection = { threadId, threadId };
```
mrexodia, line 19: I simplified this by changing from index-based selection to threadId-based selection

mrexodia:
Yolo

foralost:
yoooooolooooooo

czw., 21 gru 2023, 21:33 użytkownik Duncan Ogilvie ***@***.***>
napisał:

PR #3281, opened 2023-11-26
#3238: Escaping { from any command argument

### LINKED ISSUE #3238: "File does not exist" when launching executables from directories containing braces
### Operating System

Windows 11, x64, build 22621.2361, Chinese

### x64dbg Version

Oct 5 2023

### Describe the issue

If the executable file resides in a directory whose name contains a pair of braces (`{}`), x64dbg will refuse to launch it with a "File does not exist" error.

### Steps to reproduce

1. Pick an executable file and move it to a directory named `{}`;
2. Launch the executable from x64dbg, either by "File > Open" or drag-and-drop;
3. x64dbg will refuse to launch it with a message "File does not exist".

### Attachments

_No response_

foralost:
Escaping the `{` from file path for initalization of the debugger by pressing Open File or Drag&Drop or restarting.

Also, it allows for providing `{` as a normal character for any cmd argument that is needed to be escaped (`setthreadname`, `loadlib`, `BreakpointEdit`).

plz test :<

Closes #3238

### DIFF
--- a/src/gui/Src/Gui/MainWindow.cpp
+++ b/src/gui/Src/Gui/MainWindow.cpp
@@ -1248,7 +1248,7 @@ void MainWindow::openFileSlot()
 
 void MainWindow::openRecentFileSlot(QString filename)
 {
-    DbgCmdExec(QString().sprintf("init \"%s\"", filename.toUtf8().constData()));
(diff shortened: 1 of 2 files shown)


mrexodia:
Thanks! For next time, if you write `Closes #xxx` it will automatically close that issue after merging:

image

PR #3283, opened 2023-11-27
Double click to follow wndproc

torusrxxx:
Double click should go to wndproc.

I also think we should start wrapping WIN32 API calls in #ifdef per #3224.


mrexodia:
I don't think it's necessary to put `#ifdef` around Windows API calls just yet, at least not in the more specialized views like the handles view. For stuff like the `StdTable` etc it would be good to try and avoid Windows API directly, but I will create GitHub Actions jobs to make sure stuff is compiling for Linux etc.

PR #3291, opened 2023-12-17
#3280: Selection shifts when MemoryMap is updated

### LINKED ISSUE #3280: Selection shifts when MemoryMap is updated
### Operating System

Windows 10 x64 22H2 (OS Build 19045.3693)

### x64dbg Version

Nov 18 2023

### Describe the issue

When updating the MemoryMap, another element is selected and this is annoying.

### Steps to reproduce

1. Attach to any active process (for example DWM.exe)
2. Go to MemoryMap
3. Try to select something

### Attachments

https://github.com/x64dbg/x64dbg/assets/125130325/77d5ffbc-d559-4637-9a6f-c5cdfcaa4e7d

foralost:
Restoring the selection after the MMap is updated by holding the no. of row which was selected earlier.

https://github.com/x64dbg/x64dbg/assets/28221162/13551567-2e9b-45c5-bd9f-a71b3c78d072

Closes #3280 .

### DIFF
--- a/src/gui/Src/Gui/MemoryMapView.cpp
+++ b/src/gui/Src/Gui/MemoryMapView.cpp
@@ -41,8 +41,15 @@ MemoryMapView::MemoryMapView(StdTable* parent)
     connect(Bridge::getBridge(), SIGNAL(disassembleAt(duint, duint)), this, SLOT(disassembleAtSlot(duint, duint)));
     connect(Bridge::getBridge(), SIGNAL(focusMemmap()), this, SLOT(setFocus()));
     connect(this, SIGNAL(contextMenuSignal(QPoint)), this, SLOT(contextMenuSlot(QPoint)));
+    connect(this, SIGNAL(sortHappenedSignal()), this, SLOT(sortHappenedSlot()));
+    connect(this, SIGNAL(selectionChanged(duint)), this, SLOT(selectionChangedSlot(duint)));
 
     setupContextMenu();
+}
+
+void MemoryMapView::sortHappenedSlot()
+{
+    selectAddressRange(mSelectedAddressRange);
 }
 
 void MemoryMapView::setupContextMenu()
@@ -194,6 +201,11 @@ void MemoryMapView::setupContextMenu()
 
     refreshShortcutsSlot();
     connect(Config(), SIGNAL(shortcutsUpdated()), this, SLOT(refreshShortcutsSlot()));
+}
+
+void MemoryMapView::selectionChangedSlot(duint index)
+{
+    mSelectedAddressRange = getAddressRangeFromSelection(getSelection());
 }
 
 void MemoryMapView::refreshShortcutsSlot()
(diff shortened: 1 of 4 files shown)


mrexodia:
Hm, I think it would be better to select based on the “closest” address since most likely the number of rows updates in an application with high memory pressure… I’ll take a look when I’m back home!

foralost:
Hm, Im thinking about some delta like oldAdresses - currAddress, but im wondering what if there are multiple entries which are 'equally' distant?

... or maybe something more sophisticated

mrexodia:
How about storing the currently-selected base+size and selecting the region that overlaps (or is just before) the previous range?

src/gui/Src/Gui/MemoryMapView.cpp
```diff
@@ -444,6 +446,11 @@ void MemoryMapView::refreshMapSlot()
         setCellContent(i, ColAddress, ToPtrString((duint)mbi.BaseAddress));
         setCellUserdata(i, ColAddress, (duint)mbi.BaseAddress);
 
+        if((duint)mbi.BaseAddress == currSelectedAddressRow)
+        {
+            oldAddressFound = true;
```
mrexodia, line 451: I don't think it's necessary to delay setting the selection and storing the state. You can directly call `setSingleSelection` here to simplify everything. Also the variable name is no longer accurate, it should be something like `prevSelectedBase`

mrexodia (changes requested):

src/gui/Src/Gui/MemoryMapView.cpp
foralost: the `reloadData` function below caused some problems with `setSingleSelection` (it was jumping around like crazy)

foralost:
> How about storing the currently-selected base+size and selecting the region that overlaps (or is just before) the previous range?

Ok, soooooo we would be interested in selecting only the region based on selected address?
Like:
User clicks address -> We find the base address & size -> We store them -> We select the whole region (base address + sections and stuff)

*** the memory updates *** 
aaand tbh what now?

We look at the old base address -> We fetch its new base address & size -> We find the intersection (common range/addresses) -> We select now the intersected range?

mrexodia:
It could be something like this:

```py
void updateMemoryMap():
  oldStart = getCellUserdata(firstSelectedRow, ColAddress)
  oldEnd = getCellUserdata(lastSelectedRow, ColAddress) + getCellUserdata(lastSelectedRow, ColSize)

  firstSelectedRow = 0, lastSelectedRow = 0
  for(const region : memoryMap)
    if region.base < oldStart:
      firstSelectedRow++
    if region.base < oldEnd:
      lastSelectedRow++
    setCellContent(...)

  setSingleSelection(selectedRow)
```

Since regions (lines) in the memory map never overlap each other there will be a few possibilities:
- We just find a region where region.base == oldStart
- The region was split, we will select the first region that overlaps with the old selection
- The region was freed, we will select the first region _before_ the oldselection

The problem starts when multiple regions are selected, the algorithm for selecting the first region will work fine in this case but the part with `lastSelectedRow` is just made up and need experimentation...

foralost:
Ok, will try to work on it this weekend ;)

thug-shaker:
```math
```

foralost:
Ok so as I have understood, instead of selection of specific table lines (addresses), we should select always think about a region/span of addresses (including sizes).

image
This selection would indicate a range of addresses 000002BB03090000 - 00007DF433EA0000 + its size (both inclusive).

Whatever pop ups in this range of newly updated memoryview, should be also selected (cause its within range).

When something drops out select range (inside of it) we just dont select it in the newly outputted memory view (its line would be also gone).

When the beginning/ending line is dropped, then we should select a line which address + size overlaps with the old start/end, in the UI being it probably an upper/below line.

There is no possibility to select multiple regions in x64dbg, so I think its ok.

mrexodia:
Yeah sounds good! Unfortunately you can select multiple regions though, so that might make the code a bit tricky...

image

The principle should be the same though, I would just get the start+size of the first selected range and then expand the selection to the same amount of ranges down as before the update. If the start/size does not perfectly match I think it's fine to just reset the selection, since this should be a rare occurrence...

foralost:
I would try with having a start range as `7FFEA03D0000` to `7FFEA03EEE000` (this one would come from 7FFEA03EED000 + size).
Addresses are sorted so there is always a somehow contiguous address span.

I will try to implement it this way and see how it behaves ;]

mrexodia:
> I would try with having a start range as `7FFEA03D0000` to `7FFEA03EEE000` (this one would come from `7FFEA03EED000` + size). Addresses are sorted so there is always a somehow contiguous address span.
> 
> With multiple regions I was thinking about having selected 2 contiguous spans, which they do not overlap each other.
> 
> I will try to implement it this way and see how it behaves ;]

Yeah but the problem is that the sort can be changed, so if you get the first selected range and then expand down with N items you should (more or less) always get the same selection, without needing a special case for different sort orders.

foralost:
Oh, ok, thanks for alerting me, then I will try to be more cautious ;]

foralost:
Ok, I have managed to write code which keeps selected addresses while having `Address` column sorted, when the order changes, the selection is being updated automatically.
If something is going to be added, the selection should keep track of it, if deleted, the entry should be discarded etc.

When switching sorting from Address to any other column, the start address will be tracked `once` and selected in the other view, and then there will be no sticking to the addresses, since order of those lines in terms of `Address` column is nonexistent.

The problem is that when I switch to Protection for example, some lines are being added / deleted after sorting thus sometimes mangling the selection, which is weird.
For now I cannot find the place in source code where it is happening, will return to it in next weekends.

foralost:
Solved all of the problems and simplified the code, now everything works well regardless if the addresses are sorted or not.
Here is how it works, but I would kindly appreciate other tests as well.


https://github.com/user-attachments/assets/53809a95-e006-4f0f-a868-0e1758fb06f2

mrexodia:
I tried to simplify the code a bit, to keep things readable. I would be possible to specialize it if the address column is being sorted. I think this is enough for most use cases though and it keeps the code understandable.

PR #3292, opened 2023-12-17
#3134: Avoiding having space in MnemonicNop instruction

### LINKED ISSUE #3134: Highlight mode sometimes works incorrectly with NO-OP instructions
### Gif:
incorrect_highlight

### Problem 1:
Some tokens (spaces, commas, etc) of NOP instructions can be highlighted even though they shouldn't be, according to how this feature works with similar-looking non-NOP instructions.
**Cause:** When an instruction has no side effects, all its tokens will be assigned to the type `MnemonicNop` [here](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L358) (to color them all uniformly, I guess). Consequently, the function [ZydisTokenizer::IsHighlightableToken](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L301C29-L301C29) won't be able to filter them out based on the token type alone.
**Solution (the least terrible I've found):** Manually check for certain strings based on the token text in [ZydisTokenizer::IsHighlightableToken](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L301C29-L301C29). Something like this:
```c++
if(token.type == TokenType::MnemonicNop)
    for(QChar ch : " ,{}:+-*")
        if(token.text == QString(ch))
            return false; // do not highlight
```

### Problem 2:
It's only half-related problem but a small one, and it's also shown in the gif above (the first "nop    " case). The instructions that have no operands contain trailing spaces. Aside from the problem 1 above, I think it's only noticeable in exported disassembly (e.g. as a CSV file from the trace view, etc)
**Question:** Why would a function named "tokenizeMnemonic" insert spaces for the operands that may not even exist?
**Solution (probably):** Add a new function that would do the space padding (something like `addMnemonicPadding`). Replace all calls ([all](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L477C33-L477C33) [two](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L177) of them) to [ZydisTokenizer::tokenizeMnemonic(TokenType,QString)](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L452) with an explicit `addToken(type, mnemonic);` and call this `addMnemonicPadding` only if we're sure there are no operands ([here](https://github.com/x64dbg/x64dbg/blob/e796bdce2d3a7f997a61aec007b0237728deefc1/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L125) particularly).

foralost:
`mInst.tokens.push_back(SingleToken(mIsNop ? TokenType::MnemonicNop : type, text, value));` 

This above line would make the additional space from `addToken(TokenType::Space, " ");` to be also `TokenType::MnemonicNop`.
 (called from `bool ZydisTokenizer::tokenizeMnemonic(TokenType type, const QString & mnemonic)` [here](https://github.com/x64dbg/x64dbg/blob/009bff18c93da2f268fa52f0ceee2eea756fdb38/src/gui/Src/Disassembler/ZydisTokenizer.cpp#L501C4-L501C37) in the same file, making it highlightable later in the debugger.

https://github.com/x64dbg/x64dbg/assets/28221162/13dfda4e-7689-4e27-9310-e1f69ed31e38

https://github.com/x64dbg/x64dbg/assets/28221162/aafe217b-4a8d-4f2a-a061-74ca3eff07fd

https://github.com/x64dbg/x64dbg/assets/28221162/16f965ab-e2c7-405b-b67b-2efa11bb6ec1

Closes #3134 .

### DIFF
--- a/src/gui/Src/Disassembler/ZydisTokenizer.cpp
+++ b/src/gui/Src/Disassembler/ZydisTokenizer.cpp
@@ -351,21 +351,31 @@ bool ZydisTokenizer::TokenEquals(const SingleToken* a, const SingleToken* b, boo
     return tokenTextPoolEquals(a->text, b->text);
 }
 
+
+static bool tokenIsSpace(ZydisTokenizer::TokenType type)
+{
+    return type == ZydisTokenizer::TokenType::Space || type == ZydisTokenizer::TokenType::ArgumentSpace || type == ZydisTokenizer::TokenType::MemoryOperatorSpace;
+}
+
 void ZydisTokenizer::addToken(TokenType type, QString text, const TokenValue & value)
 {
-    switch(type)
+    bool isItSpaceType = tokenIsSpace(type);
+    if(!isItSpaceType)
     {
-    case TokenType::Space:
-    case TokenType::ArgumentSpace:
-    case TokenType::MemoryOperatorSpace:
-        break;
-    default:
         text = text.trimmed();
-        break;
     }
+
     if(mUppercase && !value.size)
         text = text.toUpper();
-    mInst.tokens.push_back(SingleToken(mIsNop ? TokenType::MnemonicNop : type, text, value));
+
+    if(isItSpaceType)
+    {
+        mInst.tokens.push_back(SingleToken(type, text, value));
+    }
+    else
+    {
+        mInst.tokens.push_back(SingleToken(mIsNop ? TokenType::MnemonicNop : type, text, value));
+    }
 }
 
 void ZydisTokenizer::addToken(TokenType type, const QString & text)


foralost:
I would like to know how to test for Problem 2 from #3134 . How to export this stuff to CSV file?

foralost:
Colors fixed 😊

src/gui/Src/Disassembler/ZydisTokenizer.h
```diff
@@ -172,6 +172,7 @@ class ZydisTokenizer
 
 
 private:
+    bool isSpaceTokenType(TokenType type);
```
mrexodia, line 175: This should probably just be a `static tokenIsSpace(TokenType type)` in `ZydisTokenizer.cpp`, since it doesn't access the members at all

mrexodia (changes requested):

src/gui/Src/Disassembler/ZydisTokenizer.h
foralost: removed `isSpaceTokenType` form `ZydisTokenizer.h` and left a `static` method `tokenIsSpace` in `ZydisTokenizer.cpp`

mrexodia:
Thanks!

PR #3294, opened 2023-12-27
feat: Add window to loaddll

lynnux:
This patch make loaddll process still running after `LoadLibraryW`, same as ollydbg.

### DIFF
--- a/src/loaddll/loaddll.cpp
+++ b/src/loaddll/loaddll.cpp
@@ -2,22 +2,65 @@
 
 wchar_t szLibraryPath[512];
 
-int main()
-{
-    wchar_t szName[256];
-    wsprintfW(szName, L"Local\\szLibraryName%X", (unsigned int)GetCurrentProcessId());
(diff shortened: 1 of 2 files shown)


mrexodia:
I have reviewed this change, but I decided to not merge it as-is. The general idea is good, but creating a window will only add an unnecessary detection vector in my view. My solution will be to write `Press any key to exit the process` to the console and wait for a key to be pressed, which achieves the same as the empty window.

The `loaddll.exe` is designed to be replaced by the user. I understand there is a desire for a loader that allows you to call exports with arguments and such (#1908), but I honestly don't think it's worth the implementation/maintenance cost. For unpacking you can usually execute `DllMain` and if you need to call a specific API it's easier to use a tool like https://github.com/stevemk14ebr/RETools/tree/master/JITCall to automate it.

mrexodia:
I pushed an alternative approach, let me know if there is anything missing from your use case here.

lynnux:
Adding `Press any key to exit the process` to the console or using `MessageBoxW` are all OK to me, thanks for the commit!

PR #3313, opened 2024-02-17
Sort by status in symbols tab

mobile46:


src/gui/Src/Gui/SymbolView.cpp
```diff
@@ -68,16 +68,23 @@ class ModuleStdTable final : public StdIconTable
...
         case MODSYMLOADING:
-            return tr("Loading");
+            text = tr("Loading");
+            break;
         case MODSYMLOADED:
-            return tr("Loaded");
+            text = tr("Loaded");
+            break;
         }
+        setCellContent(r, c, QString::number(status));
```
mrexodia, line 86: This should be set when the symbols are populated instead.

mrexodia (changes requested):

mrexodia:
What exactly doesn't work about sorting the status? It seems to work fine for me...

mobile46:
No, it's not sorting correctly. It's sorting by the previously selected row. Please see attached gif.

x64dbg_HnYwYVGRkN

mrexodia:
Lol my bad, I was testing the symbols tab 🤦‍♂️

mrexodia:
I'll get back to this btw, need to verify something before merging it.

mrexodia:
Finally had some time to investigate this and as I suspected there is an issue with the implementation. The sorting function always uses the text contents of the table, but in this case the table is 'virtual' and the symbol status isn't present. You add a number that represents the current status, but this is only done when the modules refresh.

If symbols are actually loaded, the displayed status is changed, but the number you set in the cell isn't updated. So this means you will just be sorting incorrectly in a different way.

Thanks for bringing this to my attention though, I will fix it and credit you in the commit.

PR #3314, opened 2024-02-19
Set trace register value as current value

### LINKED ISSUE #3050: Set current value of a register to the value saved in a trace from the register view in trace

sid-38:
Fixes #3050 
From the 'Trace' window, select a register value and set it as the current value. To do so, right click on the trace register value and select 'Set as current value'. This option would be available for any register that can be modified in the CPU registers window.


mrexodia:
Sorry for the late response, thanks!

PR #3333, opened 2024-04-08
Add symbol loading commands to the callstack view context menu.

### LINKED ISSUE #3332: Easily symbolizing call stacks
### Feature type

Quality of life

### Describe the feature

It would be very nice if we could symbolize callstacks efficiently from the call stack view, i.e. load symbols for all (or selected) modules in the call stack of a thread. This is very useful when debugging complex OS binaries, where a call to e.g. `CreateFile` can be preceded by a chain of functions from 10+ different modules. At the moment, the most efficient way to do this is to switch to the "Symbols" view and load symbols for modules one-by-one, or all loaded modules at once, which can take a crap load of time if the program has a giant module like combase.dll.

I implemented a demo on a local branch (https://github.com/ynwarcs/x64dbg/commit/5e4aa6053e7e5362e530fb1c1fea54536201f9a4), if this is deemed useful enough for the general audience, I can make a PR.

ynwarcs:
Fixes #3332.

### DIFF
--- a/src/gui/Src/Gui/CallStackView.cpp
+++ b/src/gui/Src/Gui/CallStackView.cpp
@@ -52,7 +52,7 @@ void CallStackView::setupContextMenu()
     QAction* loadSymbols = makeAction(DIcon("pdb"), tr("Download Symbols for Module"), SLOT(loadSymbolsSlot()));
     mMenuBuilder->addAction(loadSymbols, [this](QMenu*)
     {
-       return isSelectionValid();
+        return isSelectionValid();
     });
     QAction* loadSymbolsSingleThread = makeAction(DIcon("pdb"), tr("Download Symbols for All Modules (This Thread)"), SLOT(loadSymbolsSingleThreadSlot()));
     mMenuBuilder->addAction(loadSymbolsSingleThread, [this](QMenu*)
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/CallStackView.cpp
```diff
@@ -49,6 +49,21 @@ void CallStackView::setupContextMenu()
...
+    mMenuBuilder->addAction(loadSymbols, [this](QMenu*)
+    {
+       return isSelectionValid();
+    });
+    QAction* loadSymbolsSingleThread = makeAction(DIcon("pdb"), tr("Download Symbols for All Modules (This Thread)"), SLOT(loadSymbolsSingleThreadSlot()));
+    mMenuBuilder->addAction(loadSymbolsSingleThread, [this](QMenu*)
+    {
+        return isSelectionValid();
+    });
+    QAction* loadSymbolsAllThreads = makeAction(DIcon("pdb"), tr("Download Symbols for All Modules (All Threads)"), SLOT(loadSymbolsAllThreadsSlot()));
```
ynwarcs, line 62: I don't know how useful this is, but it felt natural to include it along with the command which loads symbols for entries of a single thread. If it doesn't look useful, I'm happy with not including it.

mrexodia: Yeah, I honestly think that downloading them for all threads is unnecessary. Probably the user will just download symbols for all modules in that case.

I'm also thinking to just have a single 'download symbols for this thread', because most likely that's what a user wants regardless. This also mirrors WinDbg's `k` command, which is pretty nice in this regard.

mrexodia (changes requested):

src/gui/Src/Gui/CallStackView.cpp
```diff
@@ -280,17 +270,10 @@ void CallStackView::renameThreadSlot()
...
-    if(DbgGetModuleAt(getCellUserdata(getInitialSelection(), ColFrom), module))
-        DbgCmdExec(QString("symdownload \"%0\"").arg(module));
-}
-
-void CallStackView::loadSymbolsSingleThreadSlot()
+void CallStackView::loadSymbolsForThreadSlot()
 {
     char module[MAX_MODULE_SIZE] = "";
-    duint firstRowForThread = 0;
+    duint firstRowForThread = 1;
```
ynwarcs, line 276: I had wrongly assumed that the user data thread-id field was being set for the thread header row as well, which lead to a bug here. It should now work properly. I could've sworn I tested that earlier but I guess it slipped through.

src/gui/Src/Gui/CallStackView.cpp
ynwarcs: That's fine. I've now removed the single-entry and all-thread commands and left only the single-thread command. I would've sparsely used the single-entry command, but it may not be all that useful in general. If others request it, it can always be re-added.

mrexodia (approved):

mrexodia:
Awesome, thanks a lot!

PR #3342, opened 2024-05-02
Update image memory region processing to support 24H2 changes

### LINKED ISSUE #3338: Memory map corruption on PE with FileAligment 0x1000
### Operating System

Windows 11, Build 24h2 26100.1

### x64dbg Version

snapshot_2024-04-11_18-47.zip

### Describe the issue

Memory map on new windows build on pe's with FileAlignment 0x1000 (almost all modules have it now):
```Address=00007FF8C6720000
Size=00000000000C8000
Party=System 
Page Information=kernel32.dll, ".text", "fothk", ".rdata", ".data", ".pdata", ".didat", ".rsrc", ".reloc"
Allocation Type=IMG
Current Protection=-R--- 
Allocation Protection=ERWC-
 
Address=00007FF6B4AA0000
Size=000000000005B000
Party=System 
Page Information=notepad.exe, ".text", "fothk", ".rdata", ".data", ".pdata", ".didat", ".rsrc", ".reloc" 
Allocation Type=IMG
Current Protection=-R--- 
Allocation Protection=ERWC-
 
Address=00007FF8C7430000
Size=0000000000256000
Party=System 
Page Information=ntdll.dll, ".text", "SCPCFG", "SCPCFGFP", "SCPCFGNP", "SCPCFGES", "RT", "PAGE", "fothk", ".rdata", ".data", ".pdata", ".mrdata", ".00cfg", ".rsrc", ".reloc"
Allocation Type=IMG
Current Protection=-R--- 
Allocation Protection=ERWC-
```

Normally you should see:
```
Address=00007FF641C00000 
Size=0000000000001000 
Party=User
Page Information=x64dbg.exe 
Allocation Type=IMG 
Current Protection=-R---
Allocation Protection=ERWC- 
Address=00007FF641C01000
Size=0000000000012000 
Party=User
Page Information=".text"
Allocation Type=IMG 
Current Protection=ER---
Allocation Protection=ERWC- 
Address=00007FF641C13000
Size=000000000000A000 
...
```

### Steps to reproduce

1. Launch debugger and check memory map

### Attachments

_No response_

ynwarcs:
24H2 adds hotpatching, with hotpatchable modules having an extra page mapped at the end of their image region. We previously assumed that the image region will only contain the header and executable sections, that's no longer the case.

Fixes #3338

### DIFF
--- a/src/dbg/memory.cpp
+++ b/src/dbg/memory.cpp
@@ -258,6 +258,17 @@ static void ProcessFileSections(std::vector<MEMPAGE> & pageVector, std::vector<S
             for(const auto & page : newPages)
                 totalSize += page.mbi.RegionSize;
 
+            if(totalSize < pageSize)
+            {
+                // On Windows 11 24H2+, modules that support hotpatching contain an extra page mapped into their image region
+                MEMPAGE imageExtensionPage = {};
+                imageExtensionPage.mbi.BaseAddress = (PVOID)(pageBase + totalSize);
+                imageExtensionPage.mbi.RegionSize = pageSize - totalSize;
+                sprintf_s(imageExtensionPage.info, "ImageExtension");
+                newPages.push_back(imageExtensionPage);
+                totalSize = pageSize;
+            }
+
             // Replace the single module page with the sections
             // newPages should never be empty, but try to avoid data loss if it is
             if(!newPages.empty() && totalSize == pageSize)
--- a/src/dbg/memory.cpp
+++ b/src/dbg/memory.cpp
@@ -258,9 +258,11 @@ static void ProcessFileSections(std::vector<MEMPAGE> & pageVector, std::vector<S
             for(const auto & page : newPages)
                 totalSize += page.mbi.RegionSize;
 
-            if(totalSize < pageSize)
+            // On Windows 11 24H2+, modules that support hotpatching contain an extra page mapped into their image region
+            // Reference: https://ynwarcs.github.io/Win11-24H2-CFG
+            static auto buildNumber = BridgeGetNtBuildNumber();
+            if(buildNumber >= 26100 && totalSize + 0x1000 == pageSize)
             {
-                // On Windows 11 24H2+, modules that support hotpatching contain an extra page mapped into their image region
                 MEMPAGE imageExtensionPage = {};
                 imageExtensionPage.mbi.BaseAddress = (PVOID)(pageBase + totalSize);
                 imageExtensionPage.mbi.RegionSize = pageSize - totalSize;


ynwarcs:
I spent a few days tracking down where an extra page in some image regions is coming from. That ultimately resulted in this post: https://ynwarcs.github.io/Win11-24H2-CFG. It's pretty long & technical so feel free to skim over it.

Basically, there's now hotpatching on Windows 11 24H2, and most of system modules are marked as hotpatchable, which makes kernel code extend their image region by a single page. This page will store new CFG functions, presumably so that they can be hotpatched painlessly. x64dbg assumed that the image region only contains header + sections, which caused the linked bug. This fix handles that, but I made it a little more generic (i.e. we don't assume there's only going to be a single page, or even that the leftover size in the region is page-aligned), if you want me to make the checks more specific to this page, let me know. At the moment, there's only a single page but this could change in the future I guess.

One more thing to note about this page is that it's seen as committed by the debugger, but is actually not committed by the OS, causing `VirtualProtect` to fail. I tried to tinker around and see if I could get it to work properly, but nothing really worked. This doesn't cause any specific bugs, but stuff like e.g. software breakpoints or writing to the memory on this page doesn't work. Since the debugger doesn't report back any issues about the failure of `VirtualProtect`, it's not obvious why the operation failed.

src/dbg/memory.cpp
```diff
@@ -258,6 +258,17 @@ static void ProcessFileSections(std::vector<MEMPAGE> & pageVector, std::vector<S
             for(const auto & page : newPages)
                 totalSize += page.mbi.RegionSize;
 
+            if(totalSize < pageSize)
+            {
+                // On Windows 11 24H2+, modules that support hotpatching contain an extra page mapped into their image region
+                MEMPAGE imageExtensionPage = {};
+                imageExtensionPage.mbi.BaseAddress = (PVOID)(pageBase + totalSize);
+                imageExtensionPage.mbi.RegionSize = pageSize - totalSize;
+                sprintf_s(imageExtensionPage.info, "ImageExtension");
```
ynwarcs, line 267: "ImageExtension" is the name that shows up in kernel functions which deal with the extra page, hence why I named it this. If the format should be different, let's change it.

mrexodia:
I think the solution should be to check for that flag + build number and insert those pages as fake pages earlier. This way other potential logic errors could be hidden.

mrexodia:
@JustasMasiulis gave me this paste:

```c
#ifndef MemoryImageExtensionInformation
#define MemoryImageExtensionInformation ((MEMORY_INFORMATION_CLASS)0xE)
#endif

struct MEMORY_IMAGE_EXTENSION_INFORMATION
{
    ULONG ExtensionType; // MEMORY_IMAGE_EXTENSION_TYPE
    ULONG Flags;
    ULONG_PTR ExtensionImageBaseRva;
    ULONG_PTR ExtensionSize;
};
```

```cpp
    for (ULONG ExtensionType = 0; ExtensionType < 4; ++ExtensionType)
    {
        MEMORY_IMAGE_EXTENSION_INFORMATION ExtensionInfo{};
        ExtensionInfo.ExtensionType = ExtensionType;
        ExtensionInfo.Flags = 0;
        NTSTATUS Status = ZwQueryVirtualMemory(NtCurrentProcess(), ImageBase, MemoryImageExtensionInformation, &ExtensionInfo, sizeof(ExtensionInfo), nullptr);
        if (!NT_SUCCESS(Status) || (ExtensionInfo.ExtensionSize == 0))
        {
            continue;
        }

        const auto ExtensionAddr = (char*)ImageBase + ExtensionInfo.ExtensionImageBaseRva;
        const auto ExtensionEnd = ExtensionAddr + ExtensionInfo.ExtensionSize;
```

ynwarcs:
Apologies for a late reply as I'd missed the activity here. 

The code above should be fine, but I think we should be using a handle to the debuggee process as the first argument to `ZwQueryVirtualMemory`, as the information we're trying to fetch can be different between the processes. Last time I tested, there was also a bug when passing remote process handles to the function, which meant we'd have to jump through some hoops to actually obtain the information. I'll give it another try tomorrow and report back though.

mrexodia:
I also managed to reproduce the issue in a VM btw. Probably just checking the PE and build number is also enough 🤷‍♂️

ynwarcs:
Apologies once again for the delay, I barely found any time to look into this. I further debugged the conditions necessary for a module to have an extra page, and there are both process-wide conditions (for example, ACG being enabled means no hotpatching = no extra page for modules), and per-module conditions (for example, no extra page if the module doesn't have [DVRT](https://denuvosoftwaresolutions.github.io/DVRT/dvrt.html)). Most conditions rely on undocumented behaviour & flags, such as DVRT or the CodeIntegrity field in the load configuration of the module. As such, I'm not sure it would be a good idea to implement detection based on these conditions, as they could silently change at any point and tracking them all down would require more effort. Unfortunately, there's not a single flag in the PE that we can rely on.

Alternatively, it may be better to handle the extra page situation gracefully, but not commit to a single "explanation" for why a page is there. For example, instead of naming the extra memory "ImageExtension", we can name it "Unknown" to let the user know we're not sure what the memory corresponds to. The way I see it, there are only two possibilities for when we have extra memory at the end of the region - either it's this new page on newer Windows 11 builds, or something non-standard is happening, either by the program or by a driver on the system, in which case we're dealing with some level of obfuscation. In the first case, the extra page is useless for user-mode debugging anyway, so it doesn't matter how we present it to the user as they're not likely to do anything with it. In the second case, we can't do much better than just admitting we don't know what the memory is, anyway.

I'll leave the decision in your hands. I can probably find some time to track down the precise conditions for a module to have an extra page, but I can't promise when that will be.

thug-shaker:
```math
```

mrexodia:
Apologies for the extreme delay. I decided to merge it almost as-is, just hardcoding the limit to 0x1000 bytes and checking the build number. To do this more properly I would need to rewrite the whole algorithm and I cba.

mrexodia:
Thanks a lot for documenting it and providing a fix btw!

PR #3372, opened 2024-06-14
Print implict memory accesses while tracing

NaC-L:
also do horrible hacks while doing that

### DIFF
--- a/src/dbg/TraceRecord.cpp
+++ b/src/dbg/TraceRecord.cpp
@@ -308,13 +308,6 @@ void TraceRecordManager::TraceExecuteRecord(const Zydis & newInstruction)
             newMemoryAddress[--newMemoryArrayCount] = newContext.registers.regcontext.csp - sizeof(duint);
             newMemoryArrayCount++;
         }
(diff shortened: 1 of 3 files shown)


src/dbg/TraceRecord.cpp
```diff
@@ -305,14 +305,14 @@ void TraceRecordManager::TraceExecuteRecord(const Zydis & newInstruction)
           )
         {
             MemRead(newContext.registers.regcontext.csp - sizeof(duint), &newMemory[newMemoryArrayCount], sizeof(duint));
-            newMemoryAddress[newMemoryArrayCount] = newContext.registers.regcontext.csp - sizeof(duint);
+            newMemoryAddress[--newMemoryArrayCount] = newContext.registers.regcontext.csp - sizeof(duint);
```
NaC-L, line 308: While this works, I believe it would be more convenient if zydis was decoding operants as [?sp-size] instead of [?sp].

NaC-L:
Just realized we dont need to do the hack fix for pop.

mrexodia:
Awesome, thanks!

PR #3388, opened 2024-07-08
Enable HiDPI for launcher

torusrxxx:
Windows supports HiDPI for dialog based application. It just needs to be enabled. And we can't use the manifest because it's shared with x32dbg.exe that has its own HiDPI policy.

### DIFF
--- a/src/launcher/x64dbg_launcher.cpp
+++ b/src/launcher/x64dbg_launcher.cpp
@@ -421,8 +421,37 @@ static void deleteZoneData(const std::wstring & rootDir)
     }
 }
 
+typedef BOOL(WINAPI* LPFN_SetProcessDpiAwarenessContext)(int);
+typedef BOOL(WINAPI* LPFN_SetProcessDPIAware)();
+
+static void EnableHiDPI()
+{
+    // Windows 10 Build 1607
+    LPFN_SetProcessDpiAwarenessContext SetProcessDpiAwarenessContext = (LPFN_SetProcessDpiAwarenessContext)GetProcAddress(GetModuleHandle(TEXT("user32.dll")), "SetProcessDpiAwarenessContext");
+    if (SetProcessDpiAwarenessContext != nullptr)
+    {
+        // Windows 10 Build 1703
+        if (SetProcessDpiAwarenessContext(-4)) //DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2
+        {
+            return;
+        }
+        // Windows 10 Build 1607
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

torusrxxx:
It turns out only DPI_AWARENESS_CONTEXT_PER_MONITOR_AWARE_V2 supports dialog based applications automatically, as advertised by Microsoft. I'm going to find an old computer and test it. If it doesn't work then the fallback modes should be removed. There aren't many users using high-definition monitor with an old computer, so the user impact should be minimal. Nevertheless, it reminds me of the control panel to settings migration that is a disastrous Microsoft failure Windows 8 - Windows 10, continues to have an impact on Windows UI design even now!

mrexodia:
Alright, thanks for the update! Just FYI the code signing certificate expired and my vendor is not responding to me, so I introduced a custom signature verification scheme in https://github.com/x64dbg/x64dbg/commit/b07f7bae051b190ae9c2adce070a7c107a61ba77. The idea is to keep all the executable files untouched until I get a new certificate and sign the DLLs with a custom signature instead.

torusrxxx:
What about SignPath: https://about.signpath.io/
Free code signing for open source. One of my another open source project Hyphanet is signing up just now.

mrexodia:
The money isn't an issue, just the validation process is extremely slow and the company I want to buy a new cert from doesn't respond. I talked to signpath in the past, but they require manual releases and AppVeyor for the CI so I skipped it... It will be resolved in time 🙂

torusrxxx:
Hyphanet release manager just mentioned that the validation process for their Windows certificate is going to delay by 6 more weeks. They have released a new version in April, but their SignPath certificate is still not ready, so 3 months later they still can't release a Windows installer.
The launcher turns blurry on Windows 7 with 150% scaling. It's the same as when HiDPI is not supported, so there's no problem.

mrexodia:
I acquired a certificate for open source projects from Certum. I moved away from them because they did not support headless signing and Sectigo still gave PFX files. But according to [this wiki entry](https://github.com/oneclick/rubyinstaller2/wiki/CertumCodeSigning) there is Linux support for the smart card and we can use `osslsigncode`, which means I can set it up on my server at home.

The identity validation completed in less than 24 hours and I generated the CSR for the certificate. I assume in a few days after the weekend they will give me the certificate. In the meantime I just overwrite the executable files from the latest signed release and sign them with the custom ECDSA scheme I implemented in https://github.com/x64dbg/x64dbg/commit/b07f7bae051b190ae9c2adce070a7c107a61ba77

PR #3392, opened 2024-07-13
Save graph as WEBP and TIFF and fix crash

torusrxxx:
It seems there's a crash when saving graph image. Qt Creator outputs the log message:
```
QPaintDevice: Cannot destroy paint device that is being painted
QWidget::repaint: Recursive repaint detected
```
The newly downloaded snapshot also crashes when saving graph image.
Also the image formats WEBP and TIFF are added. The required file format plugins are already shipped, and WEBP image is smaller and lossless, so why not.


torusrxxx:
Just realized the font in the saved image is too small, almost impossible to read the Chinese character inside the image, due to not setting the DPI for the image.

mrexodia:
Looks good, thanks!

PR #3396, opened 2024-07-21
Alert when image saving failed

torusrxxx:
The new release is not yet ready for download, so another update here. Added a messagebox when the graph image failed to save.


mrexodia:
Thanks! Over the coming weeks I will also switch x64dbg to a manual tag-based release model. Was thinking to do `YYYY.MM.DD` as a 'version number' and create proper GitHub releases too, since people have been asking for that.

mrexodia:
I’m not interested in semantic versioning. It had no value for applications where there is no interfaces to break.

PR #3398, opened 2024-07-22
fix crash and free dump in another thread

torusrxxx:
The ability to cancel loading dump also makes x64dbg crashes if the user cancels loading while x64dbg expects it to be loaded. So a return value check is now added.
The trace dump is freed in another thread so the trace can be closed fast.


torusrxxx:
The 64-bit trace index is going to be reduced to 32-bit. It is saved in trace dump a lot, could be consuming too much memory than necessary.

Using a custom coded data structure with custom memory allocator rather than STL is planned for maximum performance. Just replacing the memory allocator could already improve performance because there are too many objects in this std::map, one for each byte accessed.

mrexodia:
Thanks!

PR #3405, opened 2024-08-01
Use 32-bit trace index to save memory

torusrxxx:
Memory usage is reduced from 1.1GB to 999MB for a test trace.


torusrxxx:
[Follow in dump](https://github.com/x64dbg/x64dbg/pull/3405/commits/1179bb6f5be3bb285a152c9d2208d60cc9a14752) action is added to disassembly view and registers view in trace.

mrexodia:
Looks good, thanks!

PR #3406, opened 2024-08-01
Remove high DPI modes incompatible with windows dialog

torusrxxx:
These modes don't support dialog based application, so no need to use them. https://msdn.microsoft.com/library/mt843498(v=vs.85)


mrexodia:
Thanks!

PR #3415, opened 2024-08-07
Find pattern in trace dump

torusrxxx:
Now you can find pattern in trace dump. The memory changes during the trace, and the match can be anywhere at anytime within the trace.
Since there is no unit test, please report incorrect results.


torusrxxx:
Added time statistics for loading dump fully.
Optimized code for loading dump. Dump loading speed is increased by about 20%.
Added a dedicated function to get EIP, rather than copying the registers dump just to get EIP.
Fixed a problem that the first instruction in trace is not loaded in dump view.

PR #3425, opened 2024-08-22
Add option to copy selection bytes only

### LINKED ISSUE #3409: Copy -> Selection (Bytes only)
### Feature type

Quality of life

### Describe the feature

Currently it is possible to copy instruction bytes only with via context menu over "Binary -> Copy". However it would be nice to have also a way to do the same but with keeping the line breaks, so that i don't have to format it output manually.

image

Since are already features to copy the Addresses, assembly instructions, why not also have the possibility to copy instruction bytes line by line :)

There could be a new Context Menu entry under "Copy -> Selection (Bytes only)" or "Binary -> Copy (with Linebreaks)"

The expected result would like like this

```
74 0A 
48 8B C8 
E8 E5 61 00 00 
EB 03 
48 8B C5 
48 89 05 01 60 91 02 
E8 B4 6E 00 00 
E8 2F ED FF FF
```

G3ph4z:
Used the given plugin as a base. 

Closes #3409

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -3,6 +3,7 @@
 #include <QFile>
 #include <QDesktopServices>
 #include <QClipboard>
+#include <QStringList>
 #include "CPUDisassembly.h"
 #include "main.h"
 #include "CPUSideBar.h"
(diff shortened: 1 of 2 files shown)


src/gui/Src/Gui/CPUDisassembly.cpp
```diff
@@ -1520,6 +1522,39 @@ void CPUDisassembly::copySelectionToFileSlot(bool copyBytes)
     }
 }
 
+void CPUDisassembly::copySelectionBytesSlot()
```
mrexodia, line 1525: Thanks for your contribution! Because we're in the GUI code directly there is a slightly different framework for working with the data there. In this case the code can be simplified:

```cpp
QString clipboard = "";
    prepareDataRange(getSelectionStart(), getSelectionEnd(), [&](int i, const Instruction_t & inst)
    {
        if(i)
            clipboard += "\r\n";
        clipboard += "TODO: convert inst.dump to hex";
        return true;
    });
    Bridge::CopyToClipboard(clipboard);
```

mrexodia (changes requested):

src/gui/Src/Gui/CPUDisassembly.cpp
G3ph4z: Oh, thank you for the guidance! I was looking for an easier way to implement this than I did, but I couldn’t figure it out. Of course, I’ll update it soon.

mrexodia (approved):

mrexodia:
Awesome, thanks!

Issue #3142, opened 2023-07-23
Remove PidInHex settings, always show PID/TID in decimal


torusrxxx:
I doubt anyone is enabling that, and by removing it, UI will get more consistent with less code.

mrexodia:
At least one person I know of is using this, but I agree a hexadecimal pid/tid is quite useless in practice...

torusrxxx:
Do you think you want to decline this request?

mrexodia:
I don't think the overhead of keeping this feature isn't very big for us. As long as you use the helper function it's all transparent. Personally I'm not against removing hex necessarily, but don't feel very strong either way.

Iirc it was @justanotheranonymoususer using this feature (could be totally wrong), maybe some input?

justanotheranonymoususer:
I don't remember, but I'm not using it now, so I don't mind

torusrxxx:
Currently call stack always show TID in decimal.

PR #3426, opened 2024-08-24
Removing PidTidInHex, keeping PID/TID in dec

foralost:
This PR removes `PidInHex` settings from Configuration and its uses. All of the TIDs/PIDs are being stored in decimal format.

Issue #3142.


mrexodia:
People are actually using this setting apparently, so it cannot be removed.

foralost:
So what about the associated issue?

mrexodia:
Yeah you're right, let's nuke it.

PR #3427, opened 2024-08-24
Loading current graph in GUI_UPDATE_GRAPH_VIEW

### LINKED ISSUE #3397: Setting a comment or label in graph view does not show in the graph view
When a comment or label is set in graph view, the graph view does not show the new comment or label.
New bookmark and breakpoint have no problems here.
These actions are managed by CommonActions. GuiUpdateAllViews is called by both actions, but it does not reload the graph and therefore does not load the new comment or label.

foralost:
Loading current graph on the `GUI_UPDATE_GRAPH_VIEW` message, allowing `GuiUpdateAllViews` function to initiate a graph reloading (such as after adding a comment).

https://github.com/user-attachments/assets/51857072-4e62-4b26-9642-27d3b7209160

Closes #3397 .

### DIFF
--- a/src/gui/Src/Gui/DisassemblerGraphView.cpp
+++ b/src/gui/Src/Gui/DisassemblerGraphView.cpp
@@ -2004,10 +2004,10 @@ void DisassemblerGraphView::tokenizerConfigUpdatedSlot()
 {
     disasm.UpdateConfig();
     mPermanentHighlightingMode = ConfigBool("Disassembler", "PermanentHighlightingMode");
-    loadCurrentGraph();
+    loadCurrentGraphSlot();
 }
 
-void DisassemblerGraphView::loadCurrentGraph()
+void DisassemblerGraphView::loadCurrentGraphSlot()
 {
     if(ConfigBool("Gui", "GraphZoomMode"))
     {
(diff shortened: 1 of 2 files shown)


src/gui/Src/Bridge/Bridge.cpp
```diff
@@ -1053,6 +1053,9 @@ void* Bridge::processMessage(GUIMSG type, void* param1, void* param2)
 
     case GUI_GET_MAIN_THREAD_ID:
         return (void*)mMainThreadId;
+    case GUI_UPDATE_GRAPH_VIEW:
+        emit loadCurrentGraph();
```
mrexodia, line 1057: This is already handled in `Bridge::doUpdate` (which gets throttled properly). This can be implemented much simpler by calling `loadCurrentGraphSlot()` (that name change is good) directly inside `DisassemblerGraphView::updateGraphSlot()`.

mrexodia, line 1057: This is already handled in `Bridge::doUpdate` (which gets throttled properly). This can be implemented much simpler by calling `loadCurrentGraphSlot()` (that name change is good) directly inside `DisassemblerGraphView::updateGraphSlot()`.

mrexodia (changes requested):

src/gui/Src/Bridge/Bridge.cpp
foralost: thanks! refreshed ;]

torusrxxx:
Loading graph is more expensive, reloading time every time could make it too slow to be usable, needs performance evaluation.

torusrxxx:
Now I think performance shouldn't be a blocker to this bug fix. The user should have not worked with a huge graph in the first place.

mrexodia:
Then we'll merge it!

PR #3438, opened 2024-09-09
Fix "scriptcmd call label" when script is not running

torusrxxx:
Now it is able to use "scriptcmd call label" in breakpoint command.
There may be some hard to tell bugs. For example, x64dbg pops up "script finished" dialog when the last label is executed. Annoying, but it's difficult to tell whether this dialog should pop up.

### DIFF
--- a/src/dbg/simplescript.cpp
+++ b/src/dbg/simplescript.cpp
@@ -16,7 +16,8 @@ enum CMDRESULT
     STATUS_ERROR = false,
     STATUS_CONTINUE = true,
     STATUS_EXIT = 2,
-    STATUS_PAUSE = 3
+    STATUS_PAUSE = 3,
+    STATUS_CONTINUE_BRANCH = 4
 };
 
 static std::vector<LINEMAPENTRY> linemap;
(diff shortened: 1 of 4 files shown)


src/dbg/simplescript.cpp
```diff
@@ -692,6 +694,12 @@ bool scriptcmdexec(const char* command)
     case STATUS_PAUSE:
     case STATUS_CONTINUE:
         break;
+    case STATUS_CONTINUE_BRANCH:
+        if(!bIsRunning)
+        {
+            scriptrun(scriptIp);
```
mrexodia, line 700: I guess we need to pass down that `scriptcmd` triggered this `scriptrun` and not trigger the popup in that case?

torusrxxx:
[Added silentRet argument to scriptrun](https://github.com/x64dbg/x64dbg/pull/3438/commits/1099586326840e5715aff9ce5178d7afa1d1f5e5)
When checking the code in `debugger.cpp`, a possible scenario could occur, that the user uses `scriptcmd call label` to call a script, but that script just modifies `$breakpointcondition` instead of calling `run` itself. To ensure correctness when the script would call `run`, the debugger must have been paused before running the script; but to ensure correctness when the script uses `$breakpointcondition` this must not happen. Currently the code in breakpoint command is executed before debugger breaks, and code called via `scriptcmd call` is executed after the debugger breaks. I added a explicit wait that was previously missing, but it will only waste time if `$breakpointcondition` is false and `scriptcmd` is used.

mrexodia:
Apologies for the delay, new snapshot should be out soon!

PR #3439, opened 2024-09-09
Use LayoutDirectionAuto to fix RTL issues

torusrxxx:
It turns out just changing a setting can fix most issues. Now at least the disassembly is usable. I assume this have no performance cost.


mrexodia:
Thanks!

PR #3475, opened 2024-11-27
Fix wrong BpToBridgeHwSize return value

### LINKED ISSUE #3474: Wrong hardware memory breakpoint size shown in breakpoints view GUI
### Operating System

Windows 10 x64

### x64dbg Version

Oct 18 2024 / latest commit as of right now

### Describe the issue

When you make a byte hardware access breakpoint, the breakpoints view GUI shows word, while the console logs the correct size (byte).
Similarly, if you make a word hardware access breakpoint, the GUI shows dword, etc. On 32-bit, dword hardware breakpoint shows an empty string.

### Steps to reproduce

1. Make a byte hardware access breakpoint
2. Open the breakpoints view
3. Notice that it says word rather than byte as it should

### Attachments

_No response_

not6:
Fixes #3474

### DIFF
--- a/src/dbg/breakpoint.cpp
+++ b/src/dbg/breakpoint.cpp
@@ -782,13 +782,13 @@ static duint BpToBridgeHwSize(const BREAKPOINT & Bp)
         switch(TITANGETSIZE(Bp.titantype))
         {
         case UE_HARDWARE_SIZE_1:
-            return 1;
+            return hw_byte;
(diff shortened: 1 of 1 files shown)


mrexodia:
It's a good idea, thanks! Probably the return type should also be changed to match the enum (and the cast assignment removed).

mrexodia:
Awesome, thanks a lot!

PR #3478, opened 2024-11-29
Update heap enumeration in Windows 11 24H2

devjeonghwan:
From Windows 11 24H2, no all heap information exists in the PEB (only the first heap is stored).
So we must enumerate the entire heap along the process heap descriptor.

You can see https://cafe.naver.com/megayuchi/683 for more details


devjeonghwan:
If you want to get rid of the hard-coded Windows version-dependent structure like `_HEAP`, we can consider a solution to remotely call the Windows API `GetProcessHeaps` or `RtlGetProcessHeap` using TitanEngine.

mrexodia:
Thanks!

Issue #3509, opened 2025-01-15
Comments are not displayed in the comments section


justowee:
### Operating System

Windows 10 64bit 22H2

### x64dbg Version

Jna 6 2025

### Describe the issue

The comments will cover the mnemonic brief column, and the comment column will always be empty.

### Steps to reproduce

1. Add label
2. Add comment

### Attachments

![x64dbg](https://github.com/user-attachments/assets/0704972d-133a-4873-9be5-e88dbffc95ee)

mrexodia:
Yeah the view does not automatically refresh.

vijayleo31:
Issue Overview
Operating System: Windows 10 64bit 22H2
x64dbg Version: Jna 6 2025

Problem: The comments do not appear in the designated comment column despite being added.
Steps to Reproduce
Action 1: Add a label to a disassembly line.
Action 2: Add a comment to that same line.
Observation

The comment column is consistently empty even after following the above steps. This could indicate a display or configuration issue within x64dbg.
Possible Causes
Configuration Settings: The display settings may not be set to show comments.
Software Glitch: There may be a bug in this version of x64dbg affecting comment visibility.
Corrupted Installation: The installation of x64dbg could be corrupted, leading to unusual behavior.
Proposed Solutions
Check Display Settings:

Navigate to the settings in x64dbg and look for options related to comments or disassembly visualization.
Ensure that comments are enabled to be shown.
Update x64dbg:

Check the x64dbg website or repository for updates or patches.
Install the latest version to address any known bugs.
Reinstall x64dbg:

Uninstall the current version of x64dbg.
Download and install a fresh copy from the official site to eliminate any corruption issues.
View Debug Logs:

Check for any debug logs within x64dbg that might indicate errors or issues during operation.
This can provide insights into potential failures while adding comments.
Community Support:

Visit the x64dbg community forums or support channels.
Seek assistance from users who may have encountered similar issues.
Compatibility Mode:

Try running x64dbg in compatibility mode for an earlier version of Windows to see if that resolves the issue.
Right-click on the x64dbg executable, select Properties, and find the Compatibility tab.
Conclusion
Following these steps should help identify the root cause of the problem and potentially resolve the issue of comments not displaying as expected. If the problem persists even after trying all these solutions, consider reaching out to x64dbg’s support for further investigation.

PR #3510, opened 2025-01-16
Fix mislabeled comments/mnemonic brief columns

torusrxxx:
#3509 is because the order of ColComment and ColMnemonicBrief is not the same as the order they are added into the table.

### DIFF
--- a/src/gui/Src/BasicView/Disassembly.h
+++ b/src/gui/Src/BasicView/Disassembly.h
@@ -165,8 +165,8 @@ private:
         ColAddress,
         ColBytes,
         ColDisassembly,
-        ColComment,
         ColMnemonicBrief,
+        ColComment,
     };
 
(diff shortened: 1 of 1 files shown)


torusrxxx:
Please wait until I find time to test it.

mrexodia:
A while back I did exactly the opposite 😅 https://github.com/x64dbg/x64dbg/commit/4e5c2bd9c40abf12705949670813d9161fdf05be. My intention was to have the comment column first and then mnemonic brief, but I don't quite remember why this fixed the issue mentioned.

mrexodia:
Finally had time to look into this properly. The original reason for the problem was that a new column was introduced _in the middle_ (not at the end) and there was no new column config version created. To fix it I created `CPUDisassemblyV2` and confirmed that on clean startup everything works as expected (hidden mnemonic brief column).

PR #3516, opened 2025-02-02
Use library for command line parsing

### LINKED ISSUE #3428: Support specifying debugger commands to run on the command line
### Feature type

Quality of life

### Describe the feature

When launching x64dbg.exe or x96dbg.exe, allow to specify commands to execute right away from the command line. Similar to the [`-c` or `-cfr` options of cdb](https://learn.microsoft.com/en-us/windows-hardware/drivers/debugger/cdb-command-line-options) or [`-ex` or `-x` of gdb](https://sourceware.org/gdb/current/onlinedocs/gdb.html/File-Options.html).

`x96dbg.exe -p 1234 -c "bp somelib!somefunc; g"`

This would be useful when starting x64dbg from a script automatically e.g. for [pwntools debugging integration](https://docs.pwntools.com/en/latest/windbg.html).

peace-maker:
This allows to add additional command line flags more easily. ~~It probably requires https://github.com/LLVMParty/args/pull/3 unless I messed up the compilation setup.~~
https://github.com/LLVMParty/args is copied into the tree. You can pass additional arguments to the target program after `--` on the command line.

Add `-c` flag to run a command right away on startup and `-cf` to run multiple commands read from a file on startup.

Closes #3428

### DIFF
--- a/src/dbg/x64dbg.cpp
+++ b/src/dbg/x64dbg.cpp
@@ -5,6 +5,7 @@
  */
 
 #include "_global.h"
+#include "args.h"
 #include "command.h"
 #include "variable.h"
 #include "debugger.h"
@@ -28,6 +29,8 @@
 #include "stringformat.h"
 #include "dbghelp_safe.h"
 #include <shellapi.h>
+#include <shlwapi.h>
+#include <fstream>
 
 static MESSAGE_STACK* gMsgStack = 0;
 static HANDLE hCommandLoopThread = 0;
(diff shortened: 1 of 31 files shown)


peace-maker:
In the current state, the commands specified on the commandline using `-c` and `-cf` are executed asynchronously on startup and thus before the debugger attached to the target or started a new process. `-c "bp prog.exe!main"` prints `The command "bp prog.exe!main" is debug-only`.

The command should be applied once x64dbg is in debug mode to allow setting breakpoints right away when starting the debugger.

src/dbg/x64dbg.cpp
```diff
@@ -675,13 +677,140 @@ static DWORD WINAPI loadDbThread(LPVOID hEvent)
...
+        addPositional("filename", filename, "Filename of program to debug");
+        addPositional("currentdir", currentDir, "Current working directory of new process");
+        addExtra(extraCmdline);
+
+        addString("-p", pid, "Process ID to attach to.");
+        addString("-a", pid, "Alias for -p.");
+        addString("-tid", tid, "Thread Identifier (TID) of the thread to resume after attaching.");
+        addString("-e", event, "Handle to an Event Object to signal on attach");
+
+        // TODO: Allow repeating multiple -c arguments
```
mrexodia, line 711: You can separate commands with a `;` to execute multiple commands in a row (but as you probably saw they are not executed like the script and do not wait for the debugger to be paused again)

src/launcher/x64dbg_launcher.cpp
```diff
@@ -432,6 +435,42 @@ static void EnableHiDPI()
...
+    std::string tid;
+    std::string event;
+    std::string command;
+    std::string commandFile;
+    bool help = false;
+
+    CommandlineArguments() : ArgumentParser("x64dbg")
+    {
+        addPositional("filename", filename, "Filename of program to debug");
+        addExtra(extraCmdline);
```
mrexodia, line 460: This is a bit problematic, because it changes the command line interface of the launcher. This interface is documented and necessary for supporting GFlags and the shell integration, which simply pass `x96dbg.exe my.exe arg1 arg2`. I guess I could add support for extra arguments without a `--` delimiter, but perhaps it's better to leave the launcher alone for now?

peace-maker: Hm, yes, maybe. It'd be easier to configure if you only need one path to the x96dbg.exe instead of two for the x32 and x64 ones. Or are the paths well defined relative to the x96dbg?

mrexodia: Yeah they are always in the same spot relative to x96dbg. Also eventually I plan to completely deprecate x32dbg and implement proper wow64 support. Another idea could be to handle the transition to the right debugger executable directly in x64dbg.exe during command line parser.

mrexodia: Yeah I think the command line interface of the launcher has to stay the same, just the final command line constructed for x64dbg has to be in the new format.

For the commands I think calling `scriptcmdexec` (on the command thread) instead of `DbgCmdExec` should mostly do the trick. Currently the machinery related to programmatically execute scripts is not that good, but doing this should work :tm:

```c++
String command = "scriptcmd ";
command += args.command;
DbgCmdExec(command.c_str());
```

For the script file it should also reuse the existing initialization script functionality, similar logic already exists in `cbInitializationScriptThread`. The question is what if there is already an initialization script, should the command line override it or run before/after?

mrexodia: Another behavior could be to make the command line specify the initialization script, so if you restart the executable it will execute the same script automatically.

src/dbg/x64dbg.cpp
```diff
@@ -675,13 +677,140 @@ static DWORD WINAPI loadDbThread(LPVOID hEvent)
...
+    String tid;
+    String event;
+    String command;
+    String commandFile;
+    bool help = false;
+
+    CommandlineArguments() : ArgumentParser("x64dbg")
+    {
+        addPositional("filename", filename, "Filename of program to debug");
+        addPositional("currentdir", currentDir, "Current working directory of new process");
```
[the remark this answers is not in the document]
peace-maker: I don't understand the BridgeInit part, passing the cwd to the `init` command should work as intended? Are the changes in https://github.com/x64dbg/x64dbg/commit/800f5e6938f2fd70de517c6acdcb58409e72541d enough?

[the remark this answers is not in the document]
peace-maker: Now it uses the parent folder of the x64dbg.exe by default instead of the one my shell is currently in, so this isn't it yet.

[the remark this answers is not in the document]
mrexodia: During `BridgeInit` (which is called before the DBG/GUI components get to execute), I call `SetCurrentDirectoryW` to basically force all file operations to be relative to x64dbg's installation directory. This is necessary because of plugins and other resources have to be loaded from x64dbg's folder and not from where the user launched `x64dbg.exe`.

I think to solve this issue we need to call `GetCurrentDirectoryW` and store that in a variable, which can then be queried by calling a new `const wchar_t* BridgeStartupDirectory()` function during argument handling in `parseArguments`.

mrexodia:
Thanks again for your patience, are you can see it required quite a lot of refactoring to get working 😅

PR #3521, opened 2025-02-14
Show launcher on first launch after installing

whindsaks:
This avoids having to re-start x96dbg on first run (when x96dbg.ini has not been initialized).


mrexodia:
Thanks, sorry for the delay!

Issue #3138, opened 2023-07-16
Launcher should provide the option to uninstall context menu integration


torusrxxx:

foralost:
`context menu integration`

what is that?

torusrxxx:
Yes. To uninstall `Debug with x64dbg` entries, the user have to come to https://github.com/x64dbg/x64dbg/wiki/Frequently-Asked-Questions and download a [regfile](https://raw.githubusercontent.com/x64dbg/x64dbg/development/bin/x64dbg_shell_remove.reg). This should be in the launcher instead, the user shouldn't reach out to web to uninstall it.

mrexodia:
I honestly don’t think it’s very important to improve this. It would require some annoying changes to the launcher UI, which is currently very simple and that’s probably a benefit. I have similar reservations about #3316. It’s trivial to restart x64dbg as admin already

torusrxxx:
Maybe rename the third button in launcher from "install" to "configuration"? Not providing an uninstaller that uninstalls everything puts the software into the same category as many spyware/adware that you must download another removal tool to remove it.

torusrxxx:
Also on #3316, it's easy to restart x64dbg as admin but I'll still prefer a desktop shortcut when I just want to run it elevated every time. 10 seconds of launch/relaunch time and additional menu clicks vs 1 second, every time x64dbg is launched. Plus there's not a easy way to create the admin shortcut.

mrexodia:
Yeah I think renaming it to “configuration” is fine. As for the admin shortcut you can do it yourself very easily…

G3ph4z:
I've started working on this along with #1640, I think it goes together.

G3ph4z:
PR got merged, so this can be closed.

Issue #1640, opened 2017-07-06
More options in x96dbg setup


AlexAltea:
Nothing particularly complex. It would be great if x96dbg's "Setup" offered:
- [ ] Creating start menu entries for x64dbg.
- [ ] Updating PATH with the x64dbg folder.
- [x] Creating desktop icons.
- [x] Assigning database icons.
- [ ] Toggle context menu entries.

Since there would be way too many Yes/No dialogs, maybe using checkboxes would be easier to read.

mrexodia:
Yes, sounds good!

dreamforceinc:
How I can remove entries from contex menu?

mrexodia:
Faq.x64dbg.com

gmastergreatee:
Checkbox also for :-
- [ ] Add or remove context menu entries

G3ph4z:
If it's still applicable,, I'd be happy to try to implement this.

G3ph4z:
PR got merged, so this can be closed.

PR #3532, opened 2025-03-02
New x96dbg configuration dialog

G3ph4z:
It supposed to resolve the following issues:
- #1640 
- #3138 

Launcher:
kép

Configuration dialog:
kép


mrexodia:
Mostly looks good! I don't think it's necessary to rename 'Setup' to configuration. The alignment on the Setup dialog is not matching and things can be compacted quite a bit. I would also default to everything being selected.

G3ph4z:
Thank you! I'll update it accordingly soon.

G3ph4z:
So, updated accordingly and definitely looks better. 

image

mrexodia:
Thanks!

Issue #3136, opened 2023-07-16
Use a new REGDUMP in trace file


torusrxxx:
Current REGDUMP structure contains a lot of redundant information that shouldn't be stored on disk, and it wastes memory as well.
We should also add support for AVX-512 this time, if possible.

mrexodia:
I don’t know if there is an API for the AVX-512 context yet?

torusrxxx:
Checkout the latest Windows SDK, I've seen AVX-512 definitions just in winnt.h. Visual Studio 2022, 10.0.22000.0

torusrxxx:
I tested the below plugin code and confirm it works with AVX-512.
```cpp
#include "pluginmain.h"
#include <sstream>
#include <vector>
#include <iomanip>

#define XSTATE_LEGACY_FLOATING_POINT        (0)
#define XSTATE_LEGACY_SSE                   (1)
#define XSTATE_GSSE                         (2)
#define XSTATE_AVX                          (XSTATE_GSSE)
#define XSTATE_MPX_BNDREGS                  (3)
#define XSTATE_MPX_BNDCSR                   (4)
#define XSTATE_AVX512_KMASK                 (5)
#define XSTATE_AVX512_ZMM_H                 (6)
#define XSTATE_AVX512_ZMM                   (7)
#define XSTATE_IPT                          (8)
#define XSTATE_PASID                        (10)
#define XSTATE_CET_U                        (11)
#define XSTATE_CET_S                        (12)
#define XSTATE_AMX_TILE_CONFIG              (17)
#define XSTATE_AMX_TILE_DATA                (18)
#define XSTATE_LWP                          (62)
#define MAXIMUM_XSTATE_FEATURES             (64)

#undef CONTEXT_XSTATE

#if defined(_M_X64)
#define CONTEXT_XSTATE                      (0x00100040)
#else
#define CONTEXT_XSTATE                      (0x00010040)
#endif

#define XSTATE_MASK_GSSE                    (1ui64 << (XSTATE_GSSE))
#define XSTATE_MASK_AVX                     (XSTATE_MASK_GSSE)
#define XSTATE_MASK_MPX                     ((1ui64 << (XSTATE_MPX_BNDREGS)) | \
                                             (1ui64 << (XSTATE_MPX_BNDCSR)))

#define XSTATE_MASK_AVX512                  ((1ui64 << (XSTATE_AVX512_KMASK)) | \
[...]

Issue #3116, opened 2023-06-19
Create a simplified REGDUMP structure with only GPR


torusrxxx:
Following investigations in #3114, a simplified REGDUMP structure with only GPR could be as useful as full REGDUMP in many situations, but smaller and faster. The contents of this is RAX, RCX, RDX, RBX, RSP, RBP, RSI, RDI, R8, R9, R10, R11, R12, R13, R14, R15, RFLAGS, RIP.

Issue #693, opened 2016-06-01
Cannot set the upper half of YMM registers if the program has not used AVX


torusrxxx:
I encountered such problem when I am debugging my new **Edit Floating-point Register** dialog. It is not my fault, because the traditional "edit FPU register" dialog cannot change it either. And it works immediately after the program executes some AVX instructions.
Here's my test assembly:
```
mov eax,1
loop_begin:
movd xmm1,eax ;load eax into lower dword of xmm1
vbroadcastss ymm1,xmm1 ;spread eax everywhere
add eax,1 ;increment counter
vmovaps ymm3,ymm1 ;test an AVX instruction
mov ecx, (some memory location) 
vmovaps [ecx], ymm3 ;dump the register in memory
jmp loop_begin
```

mrexodia:
Just tested, but it appears to be working fine here.

Mattiwatti:
OK, this bug is killing me....

- I tried this on my regular Windows 7 installation: no repro, as expected from @mrexodia's findings, I could modify both the high and low parts of ymm0-15 fine.
- I then tried it in a Windows 10 RS4 VM, and the bug immediately reproduced. Still no surprises so far, Windows 10 is an OS that's broken beyond repair and this sort of thing is more the rule than the exception.
- ...But after I switched back to Windows 7, I discovered that the bug was **still occurring there**, even though it worked fine before! Rebooting doesn't help, nor does disabling VT-x/VT-d in the BIOS, flashing the entire BIOS from scratch, wiping the CMOS, or even reinstalling Windows.

The CPU is a Core i7 6700K (Skylake), and the hypervisor used is VirtualBox. I've also used the edit FPU dialog many times in the past without issues (though probably only in Windows 7). I guess I will have to buy a new CPU now since you broke it. Thanks a lot guys.

Anyway, the relevant call stack (caller at top) seems to be:
- `x64dbg.dll!setfpuvalue(string, value)`
  - `TitanEngine.dll!SetContextDataEx(hActiveThread, IndexOfRegister, NewRegisterValue)`
    - `TitanEngine.dll!_GetFullContextDataEx(hActiveThread, titcontext, avx)`
      - `if (avx) TitanEngine.dll!GetAVXContext(hActiveThread, titcontext)`
        - `GetThreadContext`
    - (value of to be updated register is copied...)
    - `TitanEngine.dll!_SetFullContextDataEx(hActiveThread, titcontext, AVX_PRIORITY)`
[...]

PR #3536, opened 2025-03-11
Add support for AVX512

torusrxxx:
This pull request adds support for AVX-512.
Now, only commands to read & write AVX-512 registers. UI will come next.


torusrxxx:
@mrexodia I need your suggestions on API changes and compatibility. Currently managing AVX-512 register contexts uses all new APIs in TitanEngine, x32dbg, and x32bridge. There will also be breaking changes to trace files (#3136).

When Intel “AVX-10” comes out with new registers R16-R31, YMM16-YMM31 and XMM16-XMM31 breaking changes will again be required. Therefore it seems better to create different REGDUMP structures for different compatibility and needs (#3116, #3136 and AVX-10) and use a new API interface like ```bool DbgGetRegDumpEx(REGDUMP* regdump, size_t size);```

mrexodia:
Yeah I agree with doing `DbgGetRegDumpEx`. Honestly I thought we were already kinda doing that. The `_dbg_xxx` APIs do not need to be backwards compatible, so we can just introduce a new size parameter there perhaps?

Also are you good with moving to Visual Studio 2022 and Qt6? I think for a while we will keep support for Qt 5.6 and VS2013 for a while, but eventually move to modern standards. Doing two snapshots will break all Qt plugins, but I think it's really time. The CMake changes should make this pretty transparent, we just have to keep writing C++11 for now if we want XP compatibility.

torusrxxx:
What do you think of API interface in TitanEngine (and you will need to copy that code into GleeBug)?

I'm already using Visual Studio 2022, just not Qt 6 yet. I don't want to have a Qt account.

mrexodia:
> What do you think of API interface in TitanEngine (and you will need to copy that code into GleeBug)?

Yeah it looks good, nothing controversial in there I think! I will fix the CI and merge it as well as port the changes to GleeBug.

> I'm already using Visual Studio 2022, just not Qt 6 yet. I don't want to have a Qt account.

I will make sure to publish the extracted Qt installations as 7z archives on GitHub and SourceForge. Any other place I should publish it to make them easy to download?

> There will also be breaking changes to trace files

I think while we're at it, there was some issue with an 8-bit index being used for memory sizes and this was causing some overflow with instructions like FXSAVE: https://pvs-studio.com/en/blog/posts/cpp/1146/#ID906A1478C3. We have a mechanism to version the trace files afaik, so that should also be a safe addition.

This weekend I will try to move x64dbg to a more modern stack, let's pray Microsoft won't decide to break ABI compatibility with their next Visual Studio release and then we should be set for the foreseeable future.

torusrxxx:
CI won’t work until changes in TitanEngine are merged. Currently `GetThreadContext` is called multiple times in TitanEngine, for general purpose registers, for AVX and for AVX-512 separately. In fact it should be called only once, that still needs to be fixed. So I’m also going to add a `DbgGetRegDumpEx`-like API interface in TitanEngine.

mrexodia:
Yeah my plan is to merge the TitanEngine/GleeBug changes and then add them as submodules to the x64dbg repo directory. This way we do not have to keep shipping binaries for updates like this...

torusrxxx:
Note: when testing AVX-512 register modification, remember #693. The program has to use AVX-512 registers once before they can be modified. I’ll see whether it can be fixed in TitanEngine.

torusrxxx:
Now it's possible to view ZMM registers in the RegistersView!
Because there are so many XMM, YMM and ZMM registers, these registers are now combined. If it has a nonzero value in the AVX512 part, then only ZMM registers are shown, otherwise if it has a nonzero value in the AVX part, then only YMM registers are shown, otherwise, just XMM registers are shown.

torusrxxx:
Now it's mostly complete except tracing support, which will be worked on later. I hope we will get it ready soon with a news update!

mrexodia:
I updated the TitanEngine.dll files for all debug engine. At a later date I will start including them in the CMake project.

PR #3544, opened 2025-03-21
Add dump support for half float

torusrxxx:
Many CPUs support F16C, so there should be support for half float in x64dbg as well.


mrexodia:
Looks good, thanks as always!

PR #3545, opened 2025-03-23
forewinds

vijayleo31:
Forwinds Struct: This struct holds the encryption key and provides methods for encryption, decryption, hashing, and file operations. Methods:
new: Initializes the Forwinds struct with a given key. encrypt: Encrypts the provided data using AES-256 encryption. decrypt: Decrypts the provided data using AES-256 decryption. hash: Computes the SHA-256 hash of the


ZehMatt:
What is the point of this? This project uses C/C++ and not Rust, have you perhaps created the PR on the wrong repo?

PR #3562, opened 2025-04-14
Update resources translations

torusrxxx:
Translations in resources files used by launcher is updated.


mrexodia:
Not sure if I pushed this, but there are some issues with UTF-8 .rc files (they do not compile with MSVC2013). I think we should switch to a mechanism similar to the Qt translations and then just read the translated strings from a text file instead of embedding them directly into the executable. This will also allow the launcher to be translated to different languages.

torusrxxx:
The resources file is by far the best option we currently have for the launcher and exe. Anything else will require something made in house or a static linked library, which inflates the code size. We can automatically generate these resources files from a translation source just before compiling for a new release.

PR #3578, opened 2025-05-05
Copy graph image

torusrxxx:
The user can copy the image in addition to save image. However, the copied image may have different resolution with saved image due to HiDPI settings.


mrexodia:
Thanks!

PR #3585, opened 2025-05-26
Draw jump lines under register label; Focus on window in handles view

torusrxxx:
Draw jump lines under register label. The register label in sidebar is now always on top of jump lines.

Focus on window in handles view. The user can call SetForegroundWindow from handles view to focus on a window when the program is running.


mrexodia:
Sorry for the delay, thanks as always!

PR #3588, opened 2025-05-27
Fixed inverted logic

OldGamesCracking:
The following commands were not working or at least returned strange results:

```
findmemall $var, E8, 0x1000
findmemall $var, E8, 123456
```

Where on the other hand these seemed to work (I guess it just seems like since they returned 5000 results [limit]):

```
findmemall $var, E8, -1
```

I'm pretty sure the start-address of the page must be **below** the search-address.


mrexodia:
Thank you!

PR #3597, opened 2025-06-18
Added Type Reload Functionality in Struct View

notpidgey:


src/gui/Src/Gui/StructWidget.cpp
```diff
@@ -109,44 +110,61 @@ void StructWidget::typeAddNode(void* parent, const TYPEDESCRIPTOR* type)
...
     text[ColValue] = ""; // NOTE: filled in later
+
     QTreeWidgetItem* item = nullptr;
     if(parent == nullptr)
     {
-        item = new QTreeWidgetItem(ui->treeWidget, text);
+        if(mInsertIndex != -1)
+        {
+            item = new QTreeWidgetItem(text);
+            ui->treeWidget->insertTopLevelItem(mInsertIndex, item);
```
mrexodia, line 145: Here it should do `mInsertIndex = -1` so the next type won't be inserted in a 'random' position.

src/gui/Src/Gui/StructWidget.cpp
```diff
@@ -467,7 +487,42 @@ void StructWidget::changeAddrSlot()
...
+
+    QTreeWidgetItem* parentItem = selectedItem->parent();
+    mInsertIndex = parentItem ? parentItem->indexOfChild(selectedItem) : ui->treeWidget->indexOfTopLevelItem(selectedItem);
+
+    auto type = selectedItem->data(0, Qt::UserRole).value<TypeDescriptor>();
+
+    delete selectedItem;
+
+    DbgCmdExec(QString("DisplayType %1, %2").arg(type.typeName).arg(ToPtrString(selectedAddr)));
+    doRefresh();
```
mrexodia, line 516: Isn't this a race? The `DisplayType` command might not be executed at this point yet, so `typeUpdateWidget` will do nothing.

src/gui/Src/Gui/StructWidget.cpp
```diff
@@ -467,7 +487,42 @@ void StructWidget::changeAddrSlot()
...
+        else
+            mInsertIndex = -1; // Reset if no selection
+    };
+
+    doRefresh();
+
+    if(!hasSelection || selectedItem->parent() || !DbgIsDebugging())
+        return;
+
+    auto selectedAddr = selectedType.addr + selectedType.offset;
```
mrexodia, line 506: Shouldn't the `selectedType.offset` always be equal to `0` (since only top-level types can be refreshed)?

src/gui/Src/Gui/StructWidget.cpp
mrexodia: If I understand the code correctly the code after the `DbgCmdExec` can be completely removed and everything should still work. The `mInsertIndex` makes sure the new node gets inserted in the right position and `typeUpdateWidget` naturally gets called at the end of the `DisplayType` implementation.

mrexodia (changes requested):
Some minor changes/questions.

notpidgey:
Having update and reload in the same GUI button might not be a great idea. Updating large types such as KUSD on release takes a substantial amount of time even when you might only want values refreshed.

PR #3607, opened 2025-06-26
Fix messages breakpoints: wnd proc or handle is unavailable (wine)

eldarkg:
Messages breakpoints don't work on wine.
When run in wine GetClassLongPtrA/W GCLP_WNDPROC return 0 then Proc is 0.
image


mrexodia:
Thank you!

PR #3622, opened 2025-07-04
Revert "Fixed inverted logic"

### LINKED ISSUE #3615: Find pattern in current module does not have 'start from selection' checkbox
### Operating System

Windows 10 64 bit

### x64dbg Version

2025.06.30

### Describe the issue

since version 2025.06.30 the "Search for Pattern" is not working anymore.

### Steps to reproduce

1. select few instructions and copy the bytes (binary copy)
2. paste the bytes in the pattern search window (search pattern in module). 
3. after searching the result is empty.

### Attachments

Image

mrexodia:
Reverts x64dbg/x64dbg#3588, see closes #3615

### DIFF
--- a/src/dbg/commands/cmd-searching.cpp
+++ b/src/dbg/commands/cmd-searching.cpp
@@ -327,7 +327,7 @@ bool cbInstrFindAllMem(int argc, char* argv[])
             }
         }
 
-        if(page.address <= addr && (find_size == -1 || page.address + page.size <= addr + find_size))
+        if(page.address >= addr && (find_size == -1 || page.address + page.size <= addr + find_size))
             searchPages.push_back(page);
     }
     SHARED_RELEASE();


mrexodia:
@OldGamesCracking it looks like your fix might not have been correct after all 😔

OldGamesCracking:
Maybe I'm missing something here, or things work different than how I think they do, but let's assume the following:

page.address = 1000
page.size = 100
addr = 1001
find_size = 1

So I want to start my search 1 byte into the page. But with the first part `page.address >= addr`, the page would not get pushed into searchPages.

Ok, let's assume `page.address >= addr` was correct and we would have the following situation:

page.address = 1000
page.size = 100
addr = 999
find_size = 10

Then `page.address >= addr` would be true, but `page.address + page.size <= addr + find_size` (1000 + 100 <= 999 + 10) would be false (for a non-global search).

Are we dealing with multiple problems here or am I just stupid? (both may be true)

Let's step back for a moment, this is what I would have written (somewhat explicitly)

```c
if ((page.address + page.size) < addr)
{
  // In either case the page is below the searchrange, exclude it
  continue;
}

if (find_size == -1)
{
  // Unlimited range (only limited by start address)
  // at least (page.address + page.size) <= addr is true
  // add page
}
else
{
  // Limited range

  if ((addr + find_size) < page.address)
  {
    // Page is beyond searchrange, exclude it
    continue;

  }
  else
  {
    // The searchrange extends into the page, add it
  }
}

searchPages.push_back(page);
```

There is a catch though: The searchpattern can now be found within the range [page.address ... addr], so some additional filtering of the results is needed, but the same applies to [addr + find_size ... page.address + page.size]

OldGamesCracking:
I was actually not fully wrong, but also not fully right, I hope #3628 fixes the problem now for both cases.

PR #3632, opened 2025-07-08
Fix autocomment precedence

### LINKED ISSUE #3591: Invalid autocomment text and wrong order of comment vs. labels
### Operating System

Windows 10 x64 build 19045

### x64dbg Version

Mar 15 2025

### Describe the issue

This problem has been there for many builds in a row, from what I know. I decided to write this after I upgraded to the latest release.
1 - contents of registers tries to show text even if there is no text in memory, so it often shows garbage instead of showing nothing or show the correct label.
Image

2 - if I set a label, I expect it to be shown instead of autogenerating garbage, e.g. 

Image

It is expected logically that the user input label will take precedence over autocomments .
3 - there is no way to disable auto-commenting like this
4 - no possibility to tune up auto-commenting so that only strings longer than N characters are commented automatically. 

### Steps to reproduce

1.  Change any command to xor rax, rax for example, now you can see user input labels at the command address 

Image

2.  Change same command to the push rdi, change rdi to the any memory location with data, where data can be parsed to the string like that 

Image
rdi address context 

Image

3. Visual error appears

### Attachments

_No response_

kalhotky:
Fixes #3591.

Before:
_before

After:
_after

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -570,14 +570,10 @@ extern "C" DLL_EXPORT bool _dbg_addrinfoget(duint addr, SEGMENTREG segment, BRID
         }
         else
         {
-            char label[MAX_LABEL_SIZE] = "";
-            if(!getLabel(addr, label, true)) //comment > label > autocomment
-            {
-                String comment;
-                retval = getAutoComment(addr, comment);
-                strcpy_s(addrinfo->comment, "\1");
-                strncat_s(addrinfo->comment, comment.c_str(), _TRUNCATE);
-            }
+            String comment;
+            retval = getAutoComment(addr, comment);
+            strcpy_s(addrinfo->comment, "\1");
+            strncat_s(addrinfo->comment, comment.c_str(), _TRUNCATE);
         }
     }
     PLUG_CB_ADDRINFO info;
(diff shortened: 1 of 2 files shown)


src/dbg/_exports.cpp
```diff
@@ -570,10 +570,14 @@ extern "C" DLL_EXPORT bool _dbg_addrinfoget(duint addr, SEGMENTREG segment, BRID
         }
         else
         {
-            String comment;
-            retval = getAutoComment(addr, comment);
-            strcpy_s(addrinfo->comment, "\1");
-            strncat_s(addrinfo->comment, comment.c_str(), _TRUNCATE);
+            char label[MAX_LABEL_SIZE] = "";
```
mrexodia: This is not the right place to change this behavior. The API requests a comment and you are returning a label which will be highlighted incorrectly by the disassembly. The actual code is spread all over (unfortunately) but I think it's mainly in `Disasembly.cpp:paintContent`

mrexodia (changes requested):
Requires a different approach.

src/dbg/_exports.cpp
kalhotky: Right, next commit should be good.

PR #3638, opened 2025-07-10
Fix incorrect "None" label for disabled trace coverage in shortcuts

### LINKED ISSUE #3635: Trace coverage "Disabled" is wrongly labeled in Hotkeys as "None"
If you go to Tracing 👉 Trace coverage then you'll find "Disable" there.
If you go to Hotkeys 👉 Type in "coverage" then you'll find "Debug -> Trace coverage -> None"

Took me a little to find out that they mean the same thing. Maybe it would be better to keep the both named the same.

adityaxa:
Hi,

This PR removes the misleading "DebugTraceRecordNone" shortcut entry, which incorrectly displayed "None" for the disabled trace coverage state in the shortcuts dialog. 

**Testing:**
*   Verified the shortcuts dialog no longer shows the confusing "None" label.
*   Confirmed that trace coverage functionality remains intact.

Let me know if there are any issues.

Closes #3635


mrexodia:
Thanks for your contribution! Are you on the [discord](https://discord.x64dbg.com)?

PR #3657, opened 2025-07-20
MSVC build under linux wine

eldarkg:
Add support to build under linux wine


cmake/msvc-wine.cmake
```diff
@@ -0,0 +1,13 @@
+set(CMAKE_SYSTEM_NAME Windows)
+
+set(MSVC_BIN_DIR $ENV{MSVC_BIN_DIR})
+set(CMAKE_C_COMPILER ${MSVC_BIN_DIR}/cl)
+set(CMAKE_CXX_COMPILER ${MSVC_BIN_DIR}/cl)
+set(CMAKE_RC_COMPILER ${MSVC_BIN_DIR}/rc)
+
+# FIX LINK : fatal error LNK1158: cannot run 'rc.exe'
```
mrexodia, line 8: Wouldn't this produce wrong executables if there is no manifest?

COMPILE-linux.md
```diff
@@ -0,0 +1,21 @@
+## Prerequisites
```
mrexodia: Please move this file into the `docs` folder to not further pollute the root.

cmake/msvc-wine.cmake
eldarkg: This workaround produce x32bridge.dll and loaddll.exe without RT_MANIFEST containing default UAC fragment:
```xml
<?xml version='1.0' encoding='UTF-8' standalone='yes'?>
<assembly xmlns='urn:schemas-microsoft-com:asm.v1' manifestVersion='1.0'>
  <trustInfo xmlns="urn:schemas-microsoft-com:asm.v3">
    <security>
      <requestedPrivileges>
        <requestedExecutionLevel level='asInvoker' uiAccess='false' />
      </requestedPrivileges>
    </security>
  </trustInfo>
</assembly>
``` 

Info: https://learn.microsoft.com/en-us/cpp/build/reference/manifest-create-side-by-side-assembly-manifest?view=msvc-170

mrexodia: Alrighty, it's Wine-only in the end so if it works for you that's great!

mrexodia:
Thanks a lot! I forgot if I already asked but are you on [Discord](https://discord.x64dbg.com)?

eldarkg:
@mrexodia I was glad to help. I don't use discord

PR #3664, opened 2025-07-24
Intermodular calls: add column for module name

ghadv0:


src/dbg/commands/cmd-searching.cpp
```diff
@@ -812,19 +813,25 @@ static bool cbModCallFind(Zydis* disasm, BASIC_INSTRUCTION_INFO* basicinfo, REFI
     if(foundaddr)
     {
         char addrText[20] = "";
+        char modName[MAX_MODULE_SIZE];
```
mrexodia, line 816: All variables should be initialized, add ` = "";`

mrexodia: All variables should be initialized, add ` = "";`

src/dbg/commands/cmd-searching.cpp
```diff
@@ -754,6 +754,7 @@ static bool cbModCallFind(Zydis* disasm, BASIC_INSTRUCTION_INFO* basicinfo, REFI
     {
         GuiReferenceInitialize(refinfo->name);
         GuiReferenceAddColumn(2 * sizeof(duint), GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Address")));
+        GuiReferenceAddColumn(0 * sizeof(duint), GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "Module")));
```
mrexodia: The value `0` is a magic number used only for the last column to use 'the rest' of the space. Some default that works for 'most' modules would be better.

src/dbg/commands/cmd-searching.cpp
```diff
@@ -812,19 +813,25 @@ static bool cbModCallFind(Zydis* disasm, BASIC_INSTRUCTION_INFO* basicinfo, REFI
     if(foundaddr)
     {
         char addrText[20] = "";
+        char modName[MAX_MODULE_SIZE];
+        if (!ModNameFromAddr(base, modName, sizeof(modName))) {
+            strcpy_s(modName, MAX_MODULE_SIZE, "<no module name");
```
mrexodia: I would leave it empty if there is no module.

mrexodia (changes requested):
Some minor changes, but looks good otherwise!

PR #3666, opened 2025-07-25
Export table to tab-separated values

torusrxxx:
TSV file is less complicated if there may be commas in the values (like disassembly), but no tabs.


mrexodia:
Awesome, thanks!

PR #3671, opened 2025-08-02
Move deleted trace file to trash can (requires Qt 5.15)

torusrxxx:
Now when clicking on "Delete" action in trace tab, the deleted trace file can be undeleted from recycler. This feature requires Qt 5.15 to compile.


mrexodia:
We are using Qt 5.12, not 5.15 so this will not be merged.

torusrxxx:
So you are stuck for not having registered a Qt account? Let's convert this into an issue and wait.

PR #3678, opened 2025-08-07
msvc-wine: use builtin linux cmkr

eldarkg:


mrexodia:
Is this particularly important? The exact cmkr version gets fetched automatically (and optionally cached if you set the CMKR_CACHE environment variable). This also wouldn't work on macos and linux famously has terrible support for executables working across distributions...

eldarkg:
@mrexodia this PR is optional at your discretion. You may close it

mrexodia:
I'd be fine with merging it if there is some way to detect the build is for Wine and then run the `cmkr.exe` through that, but not sure if that's possible...

PR #3679, opened 2025-08-09
Patch file: fix wine usage

eldarkg:
In wine: `Patch file` tries to create `patched_file.exe.*` instead `patched_file.exe`
photo_2025-08-02_16-07-55


mrexodia:
Thanks!

PR #3681, opened 2025-08-10
Goto dialog: add completion for labels

dabeibao:
The lables' completions have lower priority.
The decoration role is also set to "functions" because SymAutoComplete can only return strings.


mrexodia:
Thank you!

PR #3692, opened 2025-08-22
Implement better UX for pattern finding

mrexodia:
image

Vibe coded with Claude 4, but seems to work fine!


torusrxxx:
I think the UX is indeed more clarified. But why use vibe coding for a task when you can complete it in the same time?
Oh, I do want to add AI code comments and documentation, provided that they are verified by our developers.

mrexodia:
I’m just experimenting. A user wrote a message on Discord and I pasted it in Claude Code to see if it works. The result seems decent enough, but I need to find time to review it. Meanwhile the user can try the POC without my investing any time coding 🤷‍♂️

PR #3695, opened 2025-08-26
Added trace include/exclude modules

requizm:
PR is not ready to merge, because I need to improve the UI and increase the code quality. However, first, I want to propose the feature. This was my primary wanted feature from 2 years ago.

image

image

image

For example, if I exclude `ntdll.ll`, ntdll calls going to step over. Other modules step into.


src/gui/Src/Gui/MainWindow.ui
requizm, line 1: I only added 15 lines, did I fucked the tab/space? I couldn't figure it out.

mrexodia:
Is there any use case for this other than tracing in the main process?

requizm:
@mrexodia No actually, however, I thought it would be better if I made it more customizable.

mrexodia:
> I thought it would be better if I made it more customizable.

I don't really think so, it just adds a lot of code complexity and maintenance burden for no reason.  I think there should be a `TraceSetStepFilter none/user/system` command. Then when you run a `TraceIntoXXX` command it will use one of the following step functions:

- StepInto
- StepIntoUser
- StepIntoSystem

Same for the `TraceOverXXX` command. These functions already handle stepping until that module type (confusingly called party) is reached and they will only execute the step callback at that point, which means you do not need extra checks in the trace collection function. It also massively simplifies the UI and API surface, since you can just have a radio box `Module filter: None/User/System` (or whatever) and that will call `TraceSetStepType` before starting the trace.

Really appreciate your contribution, but would have saved some time if you asked beforehand 😅

mrexodia:
And another thought, there is an option to set a custom module type (0 and 1 being reserved as user/system).

image

So it would be possible to (later) extend this to support tracing a custom module party only if the user really needs the functionality (which honestly I doubt has a valid use case). In most cases you might just want to ignore certain modules classified as 'user' because they are in the program directory.

image

But for this you can just change their party to 'system' manually and then set the 'user' step filter.

src/gui/Src/Gui/MainWindow.ui
mrexodia: This is because I did not configure the `.editorconfig` correctly for `.ui` files, but I will fix that up later.

mrexodia:
@requizm could you try #3731, I think it should work for your use cases!

PR #3696, opened 2025-08-26
Remove header size limit when reading trace file

g0th1c54e4:
I’m developing a plugin to enhance the trace functionality of x64dbg (then it can carry more useful information to assist with reverse engineering). To achieve this, I considered appending additional data into the existing JSON blob inside the trace file. However, currently x64dbg enforces a size limit of 16,384 bytes (0x4000) for the JSON blob, which prevents this approach.

I’d like to remove this restriction. Or is there a specific reason why the JSON blob size must be limited when parsing trace files?

### DIFF
--- a/src/gui/Src/Tracer/TraceFileReader.cpp
+++ b/src/gui/Src/Tracer/TraceFileReader.cpp
@@ -406,8 +406,8 @@ void TraceFileParser::readFileHeader(TraceFileReader* that)
         throw std::wstring(L"Unspecified");
     if(header.LowPart != MAKEFOURCC('T', 'R', 'A', 'C'))
         throw std::wstring(L"File type mismatch");
-    if(header.HighPart > 16384)
-        throw std::wstring(L"Header info is too big");
+    //if(header.HighPart > 16384)
+    //    throw std::wstring(L"Header info is too big");
     QByteArray jsonData = that->traceFile.read(header.HighPart);
     if(jsonData.size() != header.HighPart)
         throw std::wstring(L"JSON header is corrupted");
--- a/src/gui/Src/Tracer/TraceFileReader.cpp
+++ b/src/gui/Src/Tracer/TraceFileReader.cpp
@@ -406,8 +406,6 @@ void TraceFileParser::readFileHeader(TraceFileReader* that)
         throw std::wstring(L"Unspecified");
     if(header.LowPart != MAKEFOURCC('T', 'R', 'A', 'C'))
         throw std::wstring(L"File type mismatch");
-    //if(header.HighPart > 16384)
-    //    throw std::wstring(L"Header info is too big");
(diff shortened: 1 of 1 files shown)


g0th1c54e4:
I forgot to clarify the reason why I want to remove the size limit: Trace files with additional appended information currently fail to open in x64dbg(The reason is simply the size restriction imposed on the JSON blob). However, I want x64dbg to still be able to open these files (I did not modify any originally recorded trace data—I only appended extra information). 
For parsing and reading this additional information, I have a separate repository that handles it: https://github.com/g0th1c54e4/x64dbg-trace-parser

mrexodia:
I'm okay with removing the limit, but perhaps @torusrxxx has some insight in why it exists. I suspect it was just added as a sanity check, because the documentation does not mention it: https://help.x64dbg.com/en/latest/developers/tracefile.html

src/gui/Src/Tracer/TraceFileReader.cpp
```diff
@@ -406,8 +406,8 @@ void TraceFileParser::readFileHeader(TraceFileReader* that)
         throw std::wstring(L"Unspecified");
     if(header.LowPart != MAKEFOURCC('T', 'R', 'A', 'C'))
         throw std::wstring(L"File type mismatch");
-    if(header.HighPart > 16384)
-        throw std::wstring(L"Header info is too big");
+    //if(header.HighPart > 16384)
```
mrexodia: Please just remove the code instead of commenting it out.

mrexodia (changes requested):
See other comment.

torusrxxx:
The size check enforced here primarily serves as a sanity check to prevent the following code from attempting to read a lot of data (necessary if the file is corrupted and specifies a size on the order of hundreds of megabytes). The JSON parser used here `QJsonDocument` is not intended to parse a big document efficiently. The size limit 16KB is a reasonable size to store the list of modules, memory maps, thread lists and things like that. It's not intended to be a debug database with many megabytes of data (which should use a better JSON parser). The correct way to add lots of information to a trace file is to define a new chunk type so that x64dbg does not attempt to decode the huge JSON with mostly unsupported information. (If you did that, please share it with us so we don't use the same type number for something else in the future.)

torusrxxx:
Hm, it seems x64dbg does not like any other chunk type (`throw std::wstring(L"Unsupported block type");`), and that should be fixed instead.

mrexodia:
I added support for custom blocks (and extended the JSON to 100kb) @g0th1c54e4, let me know if that works for you.

g0th1c54e4:
> I added support for custom blocks (and extended the JSON to 100kb) @g0th1c54e4, let me know if that works for you.

Thank you! the improvements to custom blocks will definitely make my work much easier.
I’ll also update https://github.com/g0th1c54e4/x64dbg-trace-parser to incorporate this improvement as well.

PR #3701, opened 2025-09-07
Fix ordinal imports display by resolving to function names

### LINKED ISSUE #3608: API symbols loaded by ordinal are not displayed correctly.
### Operating System

Windows 11 23H2

### x64dbg Version

2025-03-15

### Describe the issue

I wrote 3 lines of code to demonstrate the display differences between v2025-03-15 and v2021-07-01.

Important: The program loads function #981 from mfc140.dll via ordinal, not by name. This detail is critical to reproducing the behavior.

The output shown above is from v2021-07-01, and the output below is from v2025-03-15.

Image

### Steps to reproduce

1. just load test2.exe by x64dbg
2. 
3. 

### Attachments

here is the demo exe :
https://drive.google.com/file/d/1eq9imYFA0lDzxcJWqGjUN1SMA37cIjBU/view?usp=sharing

WindowsAPI:
API symbols loaded by ordinal are not displayed correctly in the disassembly. Instead of showing the function name such as `??0CWinApp@@QEAA@PEBD@Z` they display as generic `Ordinal#` placeholders instead. The symbol resolution system wasnt using PDB symbols to resolve ordinal imports to their function names from the target modules export table.

I modified `SymbolFromAddressExact` in `symbolinfo.cpp` to:

1. Detect ordinal imports by checking if an import has `ordinal != -1`
2. Find target module by locating the module that exports the function
3. Resolve the import using the target modules PDB symbols to obtain the un/decorated function name
4. Display resolved name by showing the proper function name instead of `Ordinal#XXX`

I verified with `mfc140.dll` Ordinal#981 now correctly displays:
- Before: `call qword ptr ds:[<Ordinal#981>]`
- After: 
Decorated: `call qword ptr ds:[<??0CWinApp@@QEAA@PEBD@Z>]`
Undecorated: `call qword ptr ds:[<public: __cdecl CWinApp::CWinApp(char const *)>]`

> This PR fixes #3608

### DIFF
--- a/src/dbg/symbolinfo.cpp
+++ b/src/dbg/symbolinfo.cpp
@@ -15,6 +15,8 @@
 #include "WinInet-Downloader/downslib.h"
 #include <shlwapi.h>
 
+void GetModuleInfo(MODINFO& Info, ULONG_PTR FileMapVA);
+
 duint symbolDownloadingBase = 0;
 
 struct SYMBOLCBDATA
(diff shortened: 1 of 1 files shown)


mrexodia (changes requested):
I think instead of this you can try something like:

```cpp
// Read the imported address from the IAT
duint exportAddress = 0;
DbgMemRead(addr, &exportAddress, sizeof(exportAddress));
SYMBOLINFO exportInfo = {};
if(SymbolFromAddressExact(exportAddress, exportInfo)
{
    // copy the 'exportInfo' into 'info'
    return true;
}
 ```

Alternatively this implementation also works, but the `GetModuleInfo` should be removed.

**Review by WindowsAPI — COMMENTED** [review]
this should be the proper fix you were envisioning duncan!

**Comment by mrexodia** [comment]
Sorry for the delay, thanks a lot for your contribution!

File: src/dbg/symbolinfo.cpp [hunk 1]
```diff
@@ -493,22 +493,38 @@ bool SymbolFromAddressExact(duint address, SYMBOLINFO* info)
...
+    // search in module imports (iat)
     {
         auto modImport = modInfo->findImport(rva);
         if(modImport != nullptr)
         {
+            // for imports by ordinal, try to resolve the real symbol
+            if(modImport->ordinal != -1)
+            {
+                duint exportAddress = 0;
+                if(DbgMemRead(address, &exportAddress, sizeof(exportAddress)) && exportAddress != 0)
```
d2k2-git, line 512: if exportAddress can be null, it should be checked for null before attempting to read it?
Although i think MemRead checks also for null destination buff aswell


image

maybe Buffer null check should be first before more heavy check operations (process reading) are done.

mrexodia: In this case `address` is what will be read, `exportAddress` is the destination and will never be 0

d2k2-git: i see. anyway SymbolFromAddress is doing a check for address == 0

so its checked twice currently for 0. not a big deal. up to you to clean this up or leave this defensive code.

image

PR #3704, opened 2025-09-26
Add `mod.fromoffset`, `mod.fromfileoffset` to command language

Windows81:
This function should perform the inverse of what `mod.offset` does: to get the virtual address from a file offset.

I added it because I'm trying to get the start of the `.data` region of an EXE.

Specifically, I am trying to use the `find` command to find a quick way to reconstruct [global data that gets initialised by `_initterm` upon runtime](https://medium.com/@naore32/beyond-the-main-unearthing-sneaky-functions-521f593ebeb3) (hope that doesn't raise too many questions).

However, there is no way to convert a file offset (in variable form) to a virtual offset.

The following should print out the file offset.

```
find :0,"2e 64 61 74 61"; dword:[($result+14)]
```

However, the following is invalid in the command syntax:

```
:#(dword:[($result+14)])
```

### DIFF
--- a/src/dbg/expressionfunctions.cpp
+++ b/src/dbg/expressionfunctions.cpp
@@ -72,7 +72,7 @@ void ExpressionFunctions::Init()
     RegisterEasy("mod.main,mod.mainbase", dbgdebuggedbase);
     RegisterEasy("mod.rva", modrva);
     RegisterEasy("mod.offset,mod.fileoffset", valvatofileoffset);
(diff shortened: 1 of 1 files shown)


Windows81:
@mrexodia: please confirm.

mrexodia:
In theory `:#(dword:[($result+14)])` should work, so I think that bug should be fixed. Will think about whether I want to merge this functionality, but it seems alright to me!

PR #3706, opened 2025-09-29
Use a pipe instead of temporary file for compressed database

torusrxxx:
This improved performance by multi-threading and also by eliminating temporary files.


d2k2-git:
Does this change also improves the performance for reading non compressed databases? i have usually database with 100+ MB which feels slow on startup.

torusrxxx:
> Does this change also improves the performance for reading non compressed databases? i have usually database with 100+ MB which feels slow on startup.

Yes, let's do it. Now it will try to load comments, breakpoints, functions and others concurrently. This will hopefully improve performance if you have lots of data in the database that can be loaded concurrently. If the database is small then the user probably doesn't care about the performance.

d2k2-git:
i have done some benchmarking with 300 MB (uncompressed) database files. Both versions take the same amount of time. but i like the idea of avoiding IO operations on the harddrive. maybe you can also skip saving if there has been no changes, currently x64dbg always saves the file on exit which takes 6 extra seconds before the application can be closed.

```
original
Loading database from T:\***.dd64 12593ms

yours
Loading database from T:\***.dd64 12109ms
```
```
original:
Saving database to T:\***.dd64 6015ms

yours:
Saving database to T:\***.dd64 5922ms
```

torusrxxx:
From what I can see, reading the JSON is the slowest part, and that isn't using multithreading yet. The code for reading JSON is at https://github.com/x64dbg/jansson and it is an outdated fork of https://github.com/akheron/jansson.

mrexodia:
I do not want to merge this change to be honest. Database loading has historically been the most tricky part and throwing in some `std::future` and multithreading seems like a really bad idea. There are also unrelated changes to `ModHashFromName`.

Issue #2831, opened 2022-02-09
Use different icon in sidebar for conditional breakpoint


torusrxxx:
Like Microsoft Visual Studio

PR #3713, opened 2025-10-29
Changed Conditional BP Color in Sidebar

ismaelvazquez182:
Issue #2831 

I added some code to change the color of conditional breakpoints in the sidebar to yellow. Please let me know if there's anything you would like me to change. 

image


ismaelvazquez182:
Okay, this should be ready to go now. I realized that the default colors weren't visible in light mode. This also led me to realize that conditional breakpoint colors were not editable in the appearance dialog which caused a  warning messagebox to appear. I fixed both of those issues. 

image

image

src/bridge/bridgemain.h
```diff
@@ -308,6 +308,7 @@ typedef enum
     DBG_LOOP_ADD,                   // param1=FUNCTION_LOOP_INFO* info,  param2=unused
     DBG_LOOP_DEL,                   // param1=FUNCTION_LOOP_INFO* info,  param2=unused
     DBG_IS_RUN_LOCKED,              // param1=unused,                    param2=unused
+    DBG_IS_BP_CONDITIONAL,          // param1=duint addr,                param2=unused
```
mrexodia: You cannot add new messages in the middle and this is generally the wrong approach. See https://github.com/x64dbg/x64dbg/blob/development/src/gui/Src/Utils/Breakpoints.cpp#L206C5-L206C28

src/gui/Src/Utils/Configuration.cpp
```diff
@@ -69,7 +69,9 @@ Configuration::Configuration() : QObject(), noMoreMsgbox(false)
     defaultColors.insert("SideBarCipLabelBackgroundColor", QColor("#4040FF"));
     defaultColors.insert("SideBarBulletColor", QColor("#808080"));
     defaultColors.insert("SideBarBulletBreakpointColor", QColor("#FF0000"));
+    defaultColors.insert("SideBarBulletConditionalBreakpointColor", QColor("#FFA500"));
     defaultColors.insert("SideBarBulletDisabledBreakpointColor", QColor("#00AA00"));
+    defaultColors.insert("SideBarBulletDisabledConditionalBreakpointColor", QColor("#008888"));
```
mrexodia, line 74: This also needs to be added to the dark theme.

mrexodia (changes requested):
Requires some changes, see comments.

src/gui/Src/Utils/Configuration.cpp
ismaelvazquez182: These has been added to the theme file.

SideBarBulletConditionalBreakpointColor=#FFA500
SideBarBulletDisabledConditionalBreakpointColor=#008888

PR #3714, opened 2025-11-01
Changed ctrl+a to select all items instead of selecting all characters of search filter.

### LINKED ISSUE #3698: Within the symbols tab, if you search for anything, CTRL+A will stop functioning on the updated result list
So I was on the journey of finding a specific call and this DLL had a bunch of imports and exports, most of which I did not care about (ntdll calls and the like) so I filtered it out by typing in the names that I need in the search bar on the bottom right, and upon attempting to then select the shown results with CTRL+A, there is no effect, nothing is selected. 

You actually need to scroll up, hold shift and select the first one, then scroll down and do the same again.

Strangely enough though, if you don't use the search box then it works.

ismaelvazquez182:
Closes #3698

This affects the searchbars(QLineEdit).

Current behavior:
If the user has NOT entered a search filter and they press ctrl+a, it selects all items. For example, if the user is looking at the references tab to see all intermodular calls, it will select all those calls.

If the user has entered a search filter, when they press ctrl+a, their search filter is selected instead.

Proposed change:
Always select all items when ctrl+a is pressed. (ctrl+a will no longer select the search filter)

image

### DIFF
--- a/src/gui/Src/BasicView/SearchListView.cpp
+++ b/src/gui/Src/BasicView/SearchListView.cpp
@@ -350,7 +350,6 @@ bool SearchListView::eventFilter(QObject* obj, QEvent* event)
                 return QWidget::eventFilter(obj, event);
             // Search box shortcuts reliant on mSearchBox not being empty
             case Qt::Key_X: //Ctrl+X
-            case Qt::Key_A: //Ctrl+A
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #3716, opened 2025-11-04
Changed Conditional BP Color in Sidebar

ismaelvazquez182:
This solves issue #2831.

Sorry, I'm new to git and my last pull request got really messy. I deleted it and created a new one. 

The requested changes from the previous PR #3713 have been made. The colors have been added to the dark theme and I'm using the suggested method to check whether a breakpoint is conditional.


mrexodia:
You can always squash all your commits:

```sh
git reset --soft origin/development
```

And then `git push -f` to update the pull request with new commits.

ismaelvazquez182:
Thank you! I'll save those commands for my next PR.

src/gui/Src/Gui/CPUSideBar.cpp
```diff
@@ -221,8 +223,13 @@ void CPUSideBar::paintEvent(QPaintEvent* event)
...
         duint instrVAEnd = instrVA + instr.length;
 
+        BP_REF ref;
+        DbgFunctions()->BpRefVa(&ref, bp_normal, instrVA);
+        std::string bpCondition;
+        ref.GetField(bpf_breakcondition, bpCondition);
+
         // draw bullet
-        drawBullets(&painter, line, DbgGetBpxTypeAt(instrVA) != bp_none, DbgIsBpDisabled(instrVA), DbgGetBookmarkAt(instrVA));
+        drawBullets(&painter, line, DbgGetBpxTypeAt(instrVA) != bp_none, DbgIsBpDisabled(instrVA), DbgGetBookmarkAt(instrVA), !bpCondition.empty());
```
mrexodia, line 232: Apologies, I didn't have time to review up to now.

There are some issues:
- `DbgGetBpxTypeAt(instrVA)` can return a hardware breakpoint, which would not be seen as conditional correctly
- `DbgIsBpDisabled(instrVA)` can also be queried with `ref.GetField(bpf_enabled, enabled)` which is much faster
- Doing `GetField` and then discarding the string is wasteful, especially in the rendering loop. You can do:

```cpp
DbgFunctions()->BpGetFieldText(&ref, bpf_breakcondition, [](const char* str, void* userdata)
    {
        *(bool*)userdata = str[0] != '\0';
    }, &hasCondition);
```

This avoids the unnecessary string allocations.

mrexodia (changes requested):

Issue #3599, opened 2025-06-22
Allow dll breakpoint with any name or (wildcard)


zsdex:
### Feature type

Quality of life

### Describe the feature

Would be very useful to automatically breakpoint on a dynamic lib that is loaded with deferent name each time.

![Image](https://github.com/user-attachments/assets/49bdc166-9714-4ea9-afd5-74337038967a)

mrexodia:
I think it will be easier to just enumerate all the files on disk and generate a script to set dll breakpoints on all of them with the [`bpdll`](https://help.x64dbg.com/en/latest/commands/breakpoint-control/LibrarianSetBreakpoint.html) command.

mrexodia:
I added some features that will become available in the next release.

### Testing with x64dbg

- Load the `build/Release/random_dll_test.exe` in x64dbg
- Run to the entry point
- Set a breakpoint on `LoadLibraryExW`
  - Break condition: `0`
  - Command condition: `strstr(utf16(arg(0)), ".xxl")`
  - Command text: `$bpdll {utf16(arg(0))}, a, 1` (this will generate the command `bpdll c:\temp\abcdef.xxl, a, 1`)
- Run, this should break on DLL load (1)
- Run, this should break on DLL entry (1)
- Run, this should break on DLL load (2)
- Run, this should break on DLL entry (3)
- Run, this should exit with code 0

torusrxxx:
This can still be supported, but with a different approach. We just add a new type of "wildcard dll breakpoint" that is named as "dll breakpoint" in GUI and commands. Then it will be O(1) for dll breakpoints and O(n) for "wildcard dll breakpoints".

mrexodia:
My current approach is to try and keep things as simple as possible. There are so many features in x64dbg, most of them are completely unused and it makes the maintenance much harder…

Wildcard DLL breakpoints _without scripting_ seems to be extremely niche. Wouldn’t the user be better off writing a C# of Python script at that point?

torusrxxx:
OP says he just need to pause at `*.xxl` loading, nothing else.

And yeah, some people avoid coding and prefer no-code very much. When a user must use scripting to do something, the user probably need to spend hours Googling, reading our obscure documentation and debugging, especially if there is no plugin available for it. Our users are very technical and specialized, losing productivity is not what they want.

As for maintenance, I think adding a feature is part of maintenance. Otherwise there will be little to no maintenance. Everything already works, what maintenance are needed? Why touch it?

PR #3728, opened 2025-11-12
Save graph image as PDF

torusrxxx:
It can now also save graph image as a PDF file. The quality is not so great, but it is the only vector/printing format supported and just needs a few lines of code thanks to QPdfWriter.

### DIFF
--- a/src/gui/Src/Gui/DisassemblerGraphView.cpp
+++ b/src/gui/Src/Gui/DisassemblerGraphView.cpp
@@ -16,6 +16,7 @@
 #include "StringUtil.h"
 #include "MiscUtil.h"
 #include <QMainWindow>
+#include <QPdfWriter>
 
 DisassemblerGraphView::DisassemblerGraphView(Architecture* architecture, QWidget* parent)
     : QAbstractScrollArea(parent),
@@ -2547,7 +2548,7 @@ void DisassemblerGraphView::refreshSlot()
     DbgCmdExec(QString("graph %1, force").arg(ToPtrString(this->cur_instr)));
 }
 
-QImage DisassemblerGraphView::getImage()
+void DisassemblerGraphView::paintImage(QPainter* painter)
 {
     //this->viewport()->update();
 
@@ -2566,12 +2567,7 @@ QImage DisassemblerGraphView::getImage()
     }
 
     //save viewport to image
-    auto scaleFactor = devicePixelRatioF();
-    QRect completeRenderRect = QRect(0, 0, renderWidth * scaleFactor, renderHeight * scaleFactor);
-    QImage img(completeRenderRect.size(), QImage::Format_ARGB32);
-    img.setDevicePixelRatio(scaleFactor);
-    QPainter painter(&img);
-    viewport()->render(&painter);
+    viewport()->render(painter);
 
     //restore changes made to viewport for full render saving
     viewport()->resize(size);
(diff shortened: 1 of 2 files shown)


mrexodia:
What is the use case for this? I don't really see the point...

torusrxxx:
This is just the first step of a bigger plan to also allow people copy and save images of disassembly, tables as well, without being limited to just screenshot what's on the screen. I am just testing the QPdfWriter capability from the beginning to evaluate whether this plan (printing tables) is feasible, but as it turns out not to be a giant bloat of features, codes or dependencies, then why not?

mrexodia:
The main problem for me: people will use this. It might be nice to allow more generic rendering of the table widgets to allow copying to an image, but wouldn’t changing the existing PNG exports achieve the same goal?

torusrxxx:
People will use this becomes a main problem? Probably you mean people complaining in issue tracker that they want to this but this does not work perfectly. But look at the issue tracker, we have more supported features that have unresolved issues for years. That's fine and part of open-source life. Just say that all we did is rendering it in QPdfWriter, and support should come from Qt.

What may bother me a bit is that if we want to accelerate this using OpenGL, which doesn't support PDF. But after all, OpenGL is a distant goal. I'll happily announce that we no longer support this tiny feature in exchange for OpenGL.

PNG exports for a table is indeed planned, but it is obviously not suitable for exporting the entire EXE. PDF is still needed.

mrexodia:
> But look at the issue tracker, we have more supported features that have unresolved issues for years.

Yeah, this is exactly the problem I am dealing with. Adding more and more features only creates more unresolved issues. Additionally it widens the surface area for bugs, even to existing features. My time on the project has been quite limited recently, due to other priorities in life, but I still would like to deliver a good product to our users.

Similar to the discussion in #3599 where you said:

> As for maintenance, I think adding a feature is part of maintenance. Otherwise there will be little to no maintenance. Everything already works, what maintenance are needed? Why touch it?

There are hundreds of issues, some of which contain pretty serious bugs in existing features. There is a lot of old (bad) code without tests to support refactoring them, the debug engine is still not moved from TitanEngine to something different and lots of abstractions are leaking making it very difficult to support things like remote debugging. Much of the UI was copied from Olly in 2017 and is a super bad user experience (#3692 as just an example).

Of course this PR is just a place for discussion, I don't see any real risks from merging this trivial change. However, I do notice that we have a different way of thinking about the project and I think it's good to discuss. You are the only consistent contributor after all!

torusrxxx:
Unresolved issues is only part of the problem. Actually I would like to consider "what are the use cases that we can support while not expanding project scope, complexity, number of bugs, compatibility problems etc." The top wanted features: small memory breakpoints, WINAPI SDK, arguments labels in stack, extract everything from PDB, etc, are still unsupported.
My contributions to this project are kind of random, probably because I decide to do it for something unrelated to this project, or I just did that for an unrelated project, all some totally outside factors. Recently, the only carefully thought coding tasks are the trace features and AVX-512 support. Surprisingly, nobody has raise an issue for the new trace features. Is it because these features are useless or unknown and everybody uses custom tracers, or because it works flawlessly? That's mystery as x64dbg does not phone home. Either way, I don't think some features, even major ones like the new trace features, have that much maintenance burden for you. Of course if you indeed have problems with it, let's discuss it.
And, if you want to invite me to complete your task list, feel free to do so by sharing your ideas!

mrexodia:
> The top wanted features: small memory breakpoints, WINAPI SDK, arguments labels in stack, extract everything from PDB, etc, are still unsupported.

Actually we are getting there
- [x] small memory breakpoints -> supported by commands
- [x] winapi sdk -> supported with the [ManyTypes](https://github.com/snowsnowsnows/ManyTypes) plugin (a lot of work was done for this
- [ ] arguments labels in stack -> not supported, but the infrastructure is (almost) there
- [ ] extract everything from PDB -> not supported, but the work on ManyTypes actually allow this fully now (I think)

> Surprisingly, nobody has raise an issue for the new trace features. 

Yeah this is the famous chicken-and-egg problem that plagues all x64dbg features. We do not have a well-funded marketing machine to market all the new ideas, so hardly anybody even knows about it. I recently added the 'Release notes' feature to try and improve this situation, as well as changing the release model to have bigger releases where we can announce things.

I think part of it is that the trace features in x64dbg have very limited use cases in my experience. For larger projects it is much easier to use something like PIN/Frida/DynamoRIO or even inject your own DLL to perform some hooking. This is why I personally hardly ever use these features...

> And, if you want to invite me to complete your task list, feel free to do so by sharing your ideas!

Thanks! I currently do not have a concrete plan, just scattered notes from private discussions with people, but I will try to share a more concrete roadmap in the coming year. Overall I am trying to focus on stability and the move to VS2022 because it can also attract new developers, but I agree that without a roadmap or clear direction it is kind of difficult to move the project forward...

mrexodia:
I copied this thread [here](https://github.com/orgs/x64dbg/discussions/3764) to maybe continue the discussion longer-term, this will just get lost...

PR #3729, opened 2025-11-13
Use https://github.com/actions/checkout v5

torusrxxx:
Use https://github.com/actions/checkout v5


mrexodia:
You missed `cross.yml` as well as a few other occurrences of outdated `actions/xxx`, so I added them.

PR #3736, opened 2025-11-17
Add two plugin callback for trace(StartTrace and StopTrace)

g0th1c54e4:
While developing an x64dbg module to enhance the existing code tracing functionality, I’ve been stuck for a long time on one problem: **retrieving the file path of the generated .trace32/.trace64 files.**

To address this, I came up with a possible solution. I propose adding two new callbacks to the current plugin SDK: one for the start of code tracing and one for the end of code tracing (**StartTrace** and **StopTrace**).

When the user begins code tracing, the StartTrace callback would be triggered and provide the plugin instance with the file path of the newly created .trace32 / .trace64 file. When code tracing stops, the StopTrace callback would be triggered accordingly.

```cpp
#include "plugin.h"

void traceStartCB(CBTYPE cbType, PLUG_CB_STARTTRACE* callbackInfo) {
    dprintf("open tracefile: %s\n", callbackInfo->traceFilePath);
}
void traceStopCB(CBTYPE cbType, PLUG_CB_STOPTRACE* callbackInfo) {
    dprintf("close tracefile\n");
}

bool pluginInit(PLUG_INITSTRUCT* initStruct) {
    _plugin_registercallback(pluginHandle, CB_STARTTRACE, (CBPLUGIN)traceStartCB);
    _plugin_registercallback(pluginHandle, CB_STOPTRACE, (CBPLUGIN)traceStopCB);
    return true;
}
void pluginStop() {
    _plugin_unregistercallback(pluginHandle, CB_STOPTRACE);
    _plugin_unregistercallback(pluginHandle, CB_STARTTRACE);
}

void pluginSetup() {}
```

image

I noticed that this seems to be the first time anyone has proposed adding these two callbacks, so are there any considerations or potential issues that should be taken into account?

### DIFF
--- a/src/dbg/_plugins.h
+++ b/src/dbg/_plugins.h
@@ -239,7 +239,6 @@ typedef struct
 typedef struct
 {
     const char* traceFilePath;
-    void* reserved;
 } PLUG_CB_STARTTRACE;
 
 typedef struct
--- a/src/dbg/_plugins.h
+++ b/src/dbg/_plugins.h
@@ -236,6 +236,17 @@ typedef struct
     GUIMENUTYPE hMenu;
 } PLUG_CB_MENUPREPARE;
 
+typedef struct
+{
+    const char* traceFilePath;
+    void* reserved;
+} PLUG_CB_STARTTRACE;
+
+typedef struct
+{
+    void* reserved;
+} PLUG_CB_STOPTRACE;
+
 typedef enum
 {
     ValueTypeNumber,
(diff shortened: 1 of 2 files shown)


src/dbg/_plugins.h
```diff
@@ -236,6 +236,17 @@ typedef struct
     GUIMENUTYPE hMenu;
 } PLUG_CB_MENUPREPARE;
 
+typedef struct
+{
+    const char* traceFilePath;
+    void* reserved;
```
mrexodia: This is not necessary, it's only there for structs that would otherwise be empty.

mrexodia (changes requested):
Seems like a reasonable use case, aside from a minor nit I am good with merging it!

mrexodia:
Thanks!

PR #3737, opened 2025-11-17
check Buffer first on MemRead

d2k2-git:
simple validation operations should be done before memory reading. i moved the buffer null check on top inside MemRead


mrexodia:
Sorry again for the confusion, thanks!

PR #3740, opened 2025-11-19
Add support to be able to edit the raw bytes of a register

ZehMatt:
The fact that we couldn't do this has bothered me for ages.

x64dbg_2025-11-19_23-58-031763589483


mrexodia:
Let's goo!

Issue #3733, opened 2025-11-16
Add 'Do not show again' option to MiscUtil


mrexodia:
People complained about the Windows 7 popup that it is an unsupported operating system. There should be a checkbox 'Do not show again' here that saves to the settings. Put this in `MiscUtil.cpp` as something that can easily be reused, because this is quite common.

amankumarhq:
I want to work on this. Kindly assign it to me.
Also, please let me know if there is any urgency or specific deadline.

mrexodia:
Thanks, go for it! There is no real deadline, I just want to add it before I do the next release (which hasn't been planned yet).

amankumarhq:
@mrexodia 
Sorry, I am unable to fix this issue. Kindly unassign me.

K0lin:
If possible, I would like to try to resolve this issue myself. I don't know if I will be able to, but if you are not in too much of a hurry, I can give it a try.

mrexodia:
yeah go for it 😅

mrexodia:
Not much to add to the original issue description.

d2k2-git:
Currently working on this. What is the proper place to store this in the configuration?

i use currently "Gui" section with  property  "HideErrorUnsupportedSystem"

to make it simple i would extend the existing SimpleErrorBox by this optional hide section and id parameters, when this parameters are used a checkbox will be used and a config check will be done to check whether to show the error.

<img width="574" height="132" alt="Image" src="https://github.com/user-attachments/assets/79346e60-e0d3-4ffd-8998-76c713fac53d" />

<img width="511" height="190" alt="Image" src="https://github.com/user-attachments/assets/fdf4d091-aa67-4325-88c8-bb1251ce289c" />

<img width="1101" height="474" alt="Image" src="https://github.com/user-attachments/assets/dbe94e0c-4439-4871-bc4c-928a952406ca" />

suxragi:
<x32gui.dll>
- Change "10 27" to "00 00" in columns 05-06 at address 000C5860

<x64gui.dll>
- Change "10 27" to "00 00" in columns 04-05 at address 000E8000

This alone will completely disable the alert, allowing you to continue using it comfortably even in older environments. Please try this if you want to further expand the possibilities of x64dbg.

mrexodia:
The next version will allow you to hide this message permanently.

PR #3745, opened 2025-11-22
Extend SimpleErrorBox with optional hide config parameters

d2k2-git:
Solves https://github.com/x64dbg/x64dbg/issues/3733

image

### DIFF
--- a/src/cross/widgets/MiscUtil.h
+++ b/src/cross/widgets/MiscUtil.h
@@ -12,7 +12,7 @@ QByteArray & ByteReverse(QByteArray & array);
 QByteArray ByteReverse(QByteArray && array);
 bool SimpleInputBox(QWidget* parent, const QString & title, QString defaultValue, QString & output, const QString & placeholderText, const QIcon* icon = nullptr);
 bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValue, const QStringList & choices, QString & output, bool editable, const QString & placeholderText, const QIcon* icon = nullptr, int minimumContentsLength = -1);
-void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text);
+void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const QString & hideConfigCategory = nullptr, const QString & hideConfigId = nullptr)
 void SimpleWarningBox(QWidget* parent, const QString & title, const QString & text);
 void SimpleInfoBox(QWidget* parent, const QString & title, const QString & text);
 QString getSymbolicName(duint addr);
--- a/src/cross/widgets/MiscUtil.h
+++ b/src/cross/widgets/MiscUtil.h
@@ -12,7 +12,7 @@ QByteArray & ByteReverse(QByteArray & array);
 QByteArray ByteReverse(QByteArray && array);
 bool SimpleInputBox(QWidget* parent, const QString & title, QString defaultValue, QString & output, const QString & placeholderText, const QIcon* icon = nullptr);
 bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValue, const QStringList & choices, QString & output, bool editable, const QString & placeholderText, const QIcon* icon = nullptr, int minimumContentsLength = -1);
-void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const QString & hideConfigId = {});
+void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const char* doNotShowAgainSetting = nullptr);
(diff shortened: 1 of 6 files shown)


src/cross/widgets/MiscUtil.h
```diff
@@ -12,7 +12,7 @@ QByteArray & ByteReverse(QByteArray & array);
 QByteArray ByteReverse(QByteArray && array);
 bool SimpleInputBox(QWidget* parent, const QString & title, QString defaultValue, QString & output, const QString & placeholderText, const QIcon* icon = nullptr);
 bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValue, const QStringList & choices, QString & output, bool editable, const QString & placeholderText, const QIcon* icon = nullptr, int minimumContentsLength = -1);
-void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text);
+void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const QString & hideConfigCategory = nullptr, const QString & hideConfigId = nullptr);
```
mrexodia: I think the interface should be simplified to only have `hideConfigId` and always use the `Gui` section

mrexodia: Also it should be `= {}` not `= nullptr`

src/gui/Src/Utils/MiscUtil.cpp
```diff
@@ -81,13 +83,34 @@ bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValu
...
+
     QMessageBox msg(QMessageBox::Critical, title, text, QMessageBox::NoButton, parent);
     msg.setWindowIcon(DIcon("fatal-error"));
     msg.setParent(parent, Qt::Dialog);
     msg.setWindowFlags(msg.windowFlags() & (~Qt::WindowContextHelpButtonHint));
+
+    QCheckBox* checkBox = nullptr;
+    if(useHideCheckBox)
+    {
+        checkBox = new QCheckBox(QObject::tr("Don't show this again"));
```
mrexodia: I think the text should be `Do not show again`

src/gui/Src/Utils/MiscUtil.cpp
```diff
@@ -81,13 +83,34 @@ bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValu
...
+        msg.setCheckBox(checkBox);
+    }
+
     msg.exec();
+
+    if(useHideCheckBox)
+    {
+        Config()->setBool(hideConfigCategory, hideConfigId, checkBox->isChecked());
```
mrexodia: Using `Config()` here has the downside that everything has to be added to the `Configuration.cpp`, I think using the `BridgeSettingSetUint` is fine here.

mrexodia:
Thanks a lot, just some minor comments.

d2k2-git:
> Thanks a lot, just some minor comments.

i resolved the mentioned points. also did execute format.bat before commit. but still there is a strange a error in the format check in the PR on a file which is not affected by this PR.

src/gui/Src/Utils/MiscUtil.cpp
```diff
@@ -83,14 +82,12 @@ bool SimpleChoiceBox(QWidget* parent, const QString & title, QString defaultValu
...
 }
 
-void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const QString & hideConfigCategory, const QString & hideConfigId)
+void SimpleErrorBox(QWidget* parent, const QString & title, const QString & text, const QString & hideConfigId)
 {
-    bool useHideCheckBox = hideConfigCategory != nullptr
-                           && !hideConfigCategory.isEmpty()
-                           && hideConfigId != nullptr
-                           && !hideConfigId.isEmpty();
+    bool useHideCheckBox = !hideConfigId.isNull();
```
mrexodia: Sorry for another nit, but this should be `.isEmpty()` like before 😅

Another (perhaps better) approach could be to change `const QString & hideConfigId = {}` to `const char* doNotShowAgainSetting = nullptr`, whatever you think is best!

mrexodia (changes requested):
.

src/gui/Src/Utils/MiscUtil.cpp
d2k2-git: no problem. i am very unexperienced in c++ and Qt Framework. Still learning. I just saw in the same class file an isNull check for QIcon, so i thought i will do it the same way.

image

i will take a look again.

d2k2-git: i changed to `char` now, makes probably more sense for the config api.

mrexodia:
Awesome, thanks!

PR #3748, opened 2025-11-26
Optimize expression evaluation for tracing performance

torusrxxx:
There wasn't an optimization in place for evaluating constants `0` and `1` in tracing, like conditional breakpoints. This patch added an optimized path for constant evaluation. This leads to about 10% more events per second in tracing but more testing is needed.

### DIFF
--- a/src/dbg/TraceRecord.cpp
+++ b/src/dbg/TraceRecord.cpp
@@ -233,7 +233,7 @@ static void HandleZydisOperand(const Zydis & zydis, int opindex, DISASM_ARGTYPE*
         const auto & mem = op.mem;
         if(mem.segment == ArchValue(ZYDIS_REGISTER_FS, ZYDIS_REGISTER_GS))
         {
-            *value += ThreadGetLocalBase(ThreadGetId(hActiveThread));
+            *value += ThreadGetLocalBase(GetDebugData()->dwThreadId);
         }
         *memorySize = op.size / 8;
         if(*memorySize <= memoryContentSize && DbgMemIsValidReadPtr(*value))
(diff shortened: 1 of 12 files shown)


torusrxxx:
What are your preferences to a possible future test suite?

torusrxxx:
Repeat the benchmark for 100000 times. Unoptimized branch takes 1875ms. Optimized branch takes 937ms. Optimized branch with 6ed5766ed96762c9566987cbeadbbb24ed7cdae8 reverted takes 1156ms. (tested 5 times and took the median value).
```C
duint value;
valfromstring("0", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring("1", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring("0x123abc", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring(".1234", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring("12+34", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring("bswap(1234)", &value, true, false, nullptr, nullptr, nullptr, false);
valfromstring("error_string", &value, true, false, nullptr, nullptr, nullptr, false);
```

d2k2-git:
i like the direction where this pull request is going. in general meta data of instructions should be handled completely without parsing/converting strings, which can be otherwise a performance issue on heavy usage.

torusrxxx:
I have been optimizing tracing speed, and it's becoming clear that a next version will be necessary.

The goals of trace files version 2 will be:

### AVX-512 support

With the new structure `REGDUMP_AVX512`, support for AVX-512 will be added, and redundant parts will be eliminated. This is a breaking change.

Module lists, thread lists, memory maps are also planned. These can use the JSON storage, so they are not completely breaking changes.

### Speed optimizations

First, the loading time will be completely eliminated. Trace files will open instantly as file offset indexes will be stored in a new block type rather than generated every time the file is opened. (Breaking change because x64dbg now only accepts just one block type)

Second, it has become evident that legacy `REGDUMP` includes useless fields, including winerror names, that are slow to generate. By removing those fields in `REGDUMP_AVX512`, performance is expected to be better.

Third, I will experiment with memory mapped trace file generation. It will likely improve performance due to fewer system calls.

### Size optimizations

The trace files will keep uncompressed in version 2. I will try to keep trace files small enough while not requiring other developers to write too much code to decode it and to make it friendly to generic compression algorithms.

torusrxxx:
It should be noted that with this change, expressions that are ambiguous will be interpreted as numbers first. If you add a label named "deadbeef", the expression "deadbeef" may evaluate to 0xdeadbeef not your label.

mrexodia:
Thanks!

PR #3752, opened 2025-12-01
Fix UI and crash issues in TraceFileReader when parsing trace files containing user-defined blocks.

g0th1c54e4:
TraceFileReader may cause UI display anomalies and application crashes when parsing trace files with user-defined blocks due to incorrect index updates.

image

Based on investigation, the issue lies in index updates. Before incrementing the index after calling readBlock(), the current block must be checked: only increment the index if the block is an instruction block; do not increment for user-defined blocks.

### DIFF
--- a/src/gui/Src/Tracer/TraceFileReader.cpp
+++ b/src/gui/Src/Tracer/TraceFileReader.cpp
@@ -457,14 +457,11 @@ void TraceFileParser::readFileHeader(TraceFileReader* that)
     }
 }
 
-static bool readBlock(QFile & traceFile)
+static bool readBlock(QFile & traceFile, unsigned char blockType)
 {
     if(!traceFile.isReadable())
         throw std::wstring(L"File is not readable");
-    unsigned char blockType;
     unsigned char changedCountFlags[3]; //reg changed count, mem accessed count, flags
-    if(traceFile.read((char*)&blockType, 1) != 1)
-        throw std::wstring(L"Read block type failed");
     if(blockType == 0)
     {
         if(traceFile.read((char*)&changedCountFlags, 3) != 3)
@@ -525,7 +522,10 @@ void TraceFileParser::run()
         while(!that->traceFile.atEnd())
         {
             quint64 blockStart = that->traceFile.pos();
-            bool isPageBoundary = readBlock(that->traceFile);
+            unsigned char blockType;
+            if(that->traceFile.read((char*)&blockType, 1) != 1)
+                throw std::wstring(L"Read block type failed");
+            bool isPageBoundary = readBlock(that->traceFile, blockType);
             if(isPageBoundary)
             {
                 if(lastIndex != 0)
@@ -539,7 +539,9 @@ void TraceFileParser::run()
                 if(this->isInterruptionRequested() && !that->traceFile.atEnd()) //Cancel loading
                     throw std::wstring(L"Canceled");
             }
-            index++;
+            //Increase the index only if the block is an instruction block(0), not a user-defined block(0x80 or higher)
+            if(blockType == 0)
+                index++;
         }
         if(index > 0)
             that->fileIndex.back().second.second = index - (lastIndex - 1);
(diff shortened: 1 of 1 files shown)


g0th1c54e4:
The following trace file sample was used to debug this bug. It contains user-defined blocks **(0x81, "Hello LingMo")** manually added using a hex editor.

[bug.zip](https://github.com/user-attachments/files/23857149/bug.zip)

src/gui/Src/Tracer/TraceFileReader.cpp
```diff
@@ -539,7 +539,9 @@ void TraceFileParser::run()
                 if(this->isInterruptionRequested() && !that->traceFile.atEnd()) //Cancel loading
                     throw std::wstring(L"Canceled");
             }
-            index++;
+            //Increase the index only if the block is an instruction block(0), not a user-defined block(0x80 or higher)
+            if(blockType == 0)
```
mrexodia: This should probably reuse the logic from `readBlock` and check `if(blockType < 0x80)` instead?

src/gui/Src/Tracer/TraceFileReader.cpp
```diff
@@ -600,7 +605,9 @@ void TraceFileReader::purgeLastPage()
                 lastIndex = index + 1;
                 isBlockExist = true;
             }
-            index++;
+            //Increase the index only if the block is an instruction block(0), not a user-defined block(0x80 or higher)
+            if(blockType == 0)
```
mrexodia: Same comment as above

mrexodia (changes requested):
Thanks for trying out this feature (as you can tell I clearly didn't). Some minor comments, but otherwise it looks good!

Another thing: should we document that the index is not being updated in the trace format documentation?

src/gui/Src/Tracer/TraceFileReader.cpp
g0th1c54e4: Umm, my understanding is that we should revert to letting readBlock() handle reading the blockType—meaning, the logic for reading blockType shouldn’t be lifted outside of readBlock()?

```cpp
static bool readBlock(QFile & traceFile, unsigned char* saveBlockType = nullptr)
{
    if(!traceFile.isReadable())
        throw std::wstring(L"File is not readable");
    unsigned char blockType;
    unsigned char changedCountFlags[3]; //reg changed count, mem accessed count, flags
    if(traceFile.read((char*)&blockType, 1) != 1)
        throw std::wstring(L"Read block type failed");
    if(saveBlockType)
        *saveBlockType = blockType;
...

void TraceFileParser::run()
{
        ...
        while(!that->traceFile.atEnd())
        {
            quint64 blockStart = that->traceFile.pos();
            unsigned char blockType;
            bool isPageBoundary = readBlock(that->traceFile, &blockType);
            if(blockType < 0x80)
            {
                if(isPageBoundary)
                {
                    if(lastIndex != 0)
                        that->fileIndex.back().second.second = index - (lastIndex - 1);
                    that->fileIndex.push_back(std::make_pair(index, TraceFileReader::Range(blockStart, 0)));
                    lastIndex = index + 1;
                    //Update progress
                    that->progress.store(that->traceFile.pos() * 100 / filesize);
                    if(that->progress == 100)
                        that->progress = 99;
                    if(this->isInterruptionRequested() && !that->traceFile.atEnd()) //Cancel loading
                        throw std::wstring(L"Canceled");
                }
                index++;
            }
        }
        if(index > 0)
            that->fileIndex.back().second.second = index - (lastIndex - 1);
...
```

mrexodia: No just changing the if(blockType==0) to check the the block type is not a user block.

g0th1c54e4:
> Another thing: should we document that the index is not being updated in the trace format documentation?

I think it can certainly be documented. Even if we don’t explicitly state it, users would probably assume this anyway: the presence of user-defined blocks does not affect the parsing behavior of TraceFileReader. These blocks are meant only for third-party plugins or scripts, allowing them to perform tasks beyond the core job of parsing the trace file.

mrexodia:
Thanks a lot!

PR #3766, opened 2025-12-24
Fix Fatal GUI Exception when using Ctrl+A in search boxes

mobile46:


mrexodia:
Why does this fix an exception?

mobile46:
Ctrl+A wasn't properly handled when the search box had text, causing the app to crash.

x64dbg_hEMZvCFfxt

justanotheranonymoususer:
While at it, can you perhaps make ctrl+c work there too?
And also shift+insert when it's empty.

BTW ctrl+a works for me...

mrexodia:
The whole point of this textbox is that it forwards the events to the table that it searches 😆 Ctrl+A is meant to select all symbols, not all text.

mobile46:
But this doesn't make any sense 🙂 When I click on the textbox, it has focus, so the shortcuts should apply to the textbox. If I wanted to select all symbols, I would click on the table and use the shortcut.

mrexodia:
It's a bit difficult, because when you focus the list and start typing it enters text in the search box to filter things down. This means there is no sense of focus on the text box really...

mobile46:
Okay I'll fix [selectAll()](https://github.com/x64dbg/x64dbg/blob/773482db87cdc084bb7ef0f3ded742e04e862a25/src/gui/Src/BasicView/AbstractStdTable.cpp#L680) instead

mrexodia:
Thanks a lot!

PR #3772, opened 2026-01-04
Prevent unintended execution after declining elevation

### LINKED ISSUE #3658: Dangerous execute as admin bug
When you open a binary that requires admin privileges, you get a pop up telling you that your x64dbg is not running as admin and asks you if you want to change that. If you then click on no, and then manually go to 

"File => Restart as Admin"

as soon as it has loaded as admin, it will execute that binary, although you've clicked on "No."

3rdit:
Clears the `gInitExe`, `gInitCmd`, and `gInitDir` init variables when user explicitly declines the elevation prompt, preventing the binary from being auto-executed on a manual "Restart as Admin".

This PR closes #3658.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -2877,6 +2877,14 @@ static void debugLoopFunction(INIT_STRUCT* init)
                     GuiCloseApplication();
                     return;
                 }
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks a lot!

PR #3777, opened 2026-01-10
Fix module loading when file path is stale or file is locked

### LINKED ISSUE #3756: "The module at ... triggers a weird bug, please report an issue"
### Operating System

Windows 11 64Bit, Build 26200

### x64dbg Version

Aug 19 2025

### Describe the issue

I was experimenting around with shared data segments and encountered this log. I'm pretty sure I understand why this error occurred, but the log says to report the bug so I figured I would. Basically, I managed to get a log message like "The module at ... triggers a weird bug, please report an issue" and now I am reporting the issue :)

The problem is not major because the memory of the module can still be viewed and even modified just fine, but it isn't labeled with section names like .text or .data like most modules are.

Image

### Steps to reproduce

Basically, I was playing around with if it would be possible to create a sort of "usermode MDL." Specifically, I was doing this by:
1. Marking a section of a DLL as shared using `#pragma comment(linker, "/section:SHARED,RWS")`
2. Loading the DLL normally with `LoadLibrary`.
3. Acquiring exclusive read/write access to the DLL with `CreateFile` to prevent any other processes from loading it.
4. Mapping the same DLL into another spot with `CreateFileMapping`/`SEC_IMAGE` and `MapViewOfFile`.

This allowed me to load a DLL in such a way that it is registered as a module (as opposed to only using mapped views which wouldn't accomplish this,) map the same DLL section to two different virtual addresses in the same process with different protections on them, edit a value at one address, and have it update at the other. This is an extremely bizarre set of circumstances (though it does totally work how I wanted which is awesome.)

Here's a simple program that causes the error. First you need a DLL (DllMain and exports can just be default/empty) and it needs to include this header:

```
#pragma once
#pragma data_seg("SHARED")
unsigned long sharedVariable = 0xAABBCCDD;
#pragma data_seg()
#pragma comment(linker, "/section:SHARED,RWS")
```

Then you have an EXE that loads it:

```
#include <windows.h>

int main(int argc, char** argv) {
	HANDLE shared = LoadLibrary(TEXT("shared.dll"));

	if (!shared) {
		return GetLastError();
	}

	HANDLE fileHandle = CreateFile(
		TEXT("shared.dll"),
		GENERIC_READ,
		0,
		NULL,
		OPEN_EXISTING,
		FILE_ATTRIBUTE_NORMAL | FILE_FLAG_RANDOM_ACCESS,
		NULL
	);

	if (!fileHandle || fileHandle == INVALID_HANDLE_VALUE) {
		return GetLastError();
	}

	HANDLE fileMapping = CreateFileMapping(fileHandle, NULL, PAGE_READONLY | SEC_IMAGE, 0, 0, NULL);

	if (!fileMapping) {
		return GetLastError();
	}

	LPVOID filePointer = MapViewOfFile(fileMapping, FILE_MAP_READ, 0, 0, 0);

	if (!filePointer) {
		return GetLastError();
	}
	return 0;
}
```

I don't know if the shared memory is actually a contributing factor to the bug at all. I think it's probably mostly just caused by the unusual circumstance of mapping a DLL that can't be read (because `CreateFile` locked it without allowing `FILE_SHARE_READ` permission.) I was tempted to just ignore the issue, because without being able to read the raw file, I kind of doubt it is solvable in any reasonable way. But the log said please report an issue so, you know, here it is ^^'

### Attachments

_No response_

3rdit:
This PR addresses the "weird bug" message that appears when module headers can't be loaded.

The issue occurs in two scenarios. First, kernel section caching can return stale file paths after a file is renamed or moved, causing the file open to fail. Second, if a file is locked with exclusive access, the debugger can't open it by path even though the path is correct.

The fix adds a fallback chain in `ModLoad`. It first tries opening the file by path as before. If that fails and we have the debug API's file handle, it maps directly from that handle instead of trying to reopen. If both fail, it reads from process memory as a last resort, reading the PE header first to determine the correct module size (since callers pass `Size=1` as a placeholder).

The path resolution in `cbLoadDll` now also validates that paths actually exist before using them, uses `FSCTL_CHECK_FOR_SECTION` to invalidate stale kernel section cache, and tries to open the file ourselves when the debug API provides a NULL file handle.

This PR closes #3756

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1844,14 +1844,46 @@ static void cbLoadDll(LOAD_DLL_DEBUG_INFO* LoadDll)
     hActiveThread = ThreadGetHandle(GetDebugData()->dwThreadId);
     void* base = LoadDll->lpBaseOfDll;
 
+    // Retrieve the DLL path using a fallback
+    // https://github.com/x64dbg/x64dbg/issues/3756
     char DLLDebugFileName[MAX_PATH] = "";
-    if(!GetFileNameFromHandle(LoadDll->hFile, DLLDebugFileName, _countof(DLLDebugFileName)))
+    bool validPath = false;
+
+    //1 - Try kernel path from file handle
+    if(GetFileNameFromHandle(LoadDll->hFile, DLLDebugFileName, _countof(DLLDebugFileName)) &&
+       FileExists(DLLDebugFileName))
     {
-        if(!GetFileNameFromModuleHandle(fdProcessInfo->hProcess, HMODULE(base), DLLDebugFileName, _countof(DLLDebugFileName)))
-            strcpy_s(DLLDebugFileName, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "??? (GetFileNameFromHandle failed)")));
+        validPath = true;
     }
 
-    ModLoad((duint)base, 1, DLLDebugFileName);
+    //2 - Invalidate section cache and retry
+    // FSCTL_CHECK_FOR_SECTION releases cached image sections. This primarily helps step 3 (GetMappedFileNameW)
+    // which uses NtQueryVirtualMemory. The retry here is kinda speculative - GetFileNameFromHandle uses
+    // NtQueryInformationFile which may not benefit, but we retry anyway in case it helps.
+    // https://github.com/x64dbg/x64dbg/issues/3756
+    if(!validPath && LoadDll->hFile)
+    {
+        IO_STATUS_BLOCK iosb = {};
+        constexpr ULONG FSCTL_CHECK_FOR_SECTION = 0x90348;
+        NtFsControlFile(LoadDll->hFile, nullptr, nullptr, nullptr, &iosb, FSCTL_CHECK_FOR_SECTION, nullptr, 0, nullptr, 0);
+        if(GetFileNameFromHandle(LoadDll->hFile, DLLDebugFileName, _countof(DLLDebugFileName)) &&
+           FileExists(DLLDebugFileName))
+        {
+            validPath = true;
+        }
+    }
+
+    //3 - Try GetMappedFileNameW (with internal fallback to PEB)
+    if(!validPath)
+    {
+        validPath = GetFileNameFromModuleHandle(fdProcessInfo->hProcess, HMODULE(base), DLLDebugFileName, _countof(DLLDebugFileName));
+    }
+
+    //Give up
+    if(!validPath)
+        strcpy_s(DLLDebugFileName, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "??? (GetFileNameFromHandle failed)")));
+
+    ModLoad((duint)base, 1, DLLDebugFileName, true, LoadDll->hFile);
 
     // Update memory map
     MemUpdateMapAsync();
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1851,7 +1851,7 @@ static void cbLoadDll(LOAD_DLL_DEBUG_INFO* LoadDll)
 
     //1 - Try kernel path from file handle
     if(GetFileNameFromHandle(LoadDll->hFile, DLLDebugFileName, _countof(DLLDebugFileName)) &&
-       FileExists(DLLDebugFileName))
+            FileExists(DLLDebugFileName))
     {
         validPath = true;
     }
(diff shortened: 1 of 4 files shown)


src/dbg/debugger.cpp
```diff
@@ -1844,14 +1844,58 @@ static void cbLoadDll(LOAD_DLL_DEBUG_INFO* LoadDll)
...
-    ModLoad((duint)base, 1, DLLDebugFileName);
+    //Give up
+    if(!validPath)
+        strcpy_s(DLLDebugFileName, GuiTranslateText(QT_TRANSLATE_NOOP("DBG", "??? (GetFileNameFromHandle failed)")));
+
+    // If hFile is NULL but we have a valid path, try to open the file ourselves
+    HANDLE effectiveHFile = LoadDll->hFile;
+    if(!effectiveHFile && validPath)
+    {
+        effectiveHFile = CreateFileA(DLLDebugFileName, GENERIC_READ,
```
mrexodia: Is this actually necessary? The code inside of ModLoad also opens the handle inside of `StaticFileLoadW` from the path.

src/dbg/module.cpp
```diff
@@ -928,25 +928,114 @@ bool ModLoad(duint Base, duint Size, const char* FullPath, bool loadSymbols)
...
         if(StaticFileLoadW(wszFullPath.c_str(), UE_ACCESS_READ, false, &info.fileHandle, &info.loadedSize, &info.fileMap, &info.fileMapVA))
         {
             // Fix an anti-debug trick, which opens exclusive access to the file
             CloseHandle(info.fileHandle);
             info.fileHandle = (HANDLE)1; // Set to non-zero for TitanEngine compatibility
+            fileLoaded = true;
+        }
 
-            GetModuleInfo(info, info.fileMapVA);
+        // 2. If path-based loading failed and we have a file handle, try mapping from it
```
mrexodia: I'm actually thinking about this, maybe it would be better to use the `hFile` when available first and fall back to loading the file from the path? Kind of a design change, but the `hFile` in the debug event comes from the kernel so it's more likely to be correct.

3rdit: Addressed in https://github.com/x64dbg/x64dbg/pull/3777/commits/e6bfcc791dbfbfa1d1b7e754b45e04802ba9f827

src/dbg/debugger.cpp
3rdit: Addressed in https://github.com/x64dbg/x64dbg/pull/3777/commits/e6bfcc791dbfbfa1d1b7e754b45e04802ba9f827

mrexodia:
Thanks a lot!

diversenok:
I like the idea of having a fallback mechanism that reads the image from memory when the file is inaccessible. The other steps, however, won't help.

Let's recap what causes the problem: the image section being mapped references a file object that (for one reason or another) cached an outdated name. `NtQueryVirtualMemory` returns the wrong name because that's what this file object tells it (since the function follows VAD -> section control area -> file object -> file name). The debugging subsystem has the identical problem: it knows the image section object, and that's it. The handle it returns in the debug event is merely an attempt to reopen the file based on the (potentially bogus) name. See `DbgkpSectionToFileHandle`, which consists of `MmGetFileNameForSection` followed by `ZwOpenFile`. The name querying logic here uses the same code path as in `NtQueryVirtualMemory`.

So the returned file handle doesn't help (aside from saving a few function calls). If the referenced name was correct, we could open it ourselves. If the name was incorrect, the debug event would include no handle or a handle to the wrong file. Similarly, `FSCTL_CHECK_FOR_SECTION` can't help because there is no way to issue it against the correct file when needed. And, either way, this FSCTL releases the cached (broken) file object for the control area, but only if there are **no section objects** left (otherwise, it returns `STATUS_SHARING_VIOLATION`). As a result, `FSCTL_CHECK_FOR_SECTION` can only help when issued **before** creating sections, by preventing these future sections from reusing the stale file object. It doesn't change names returned from any file objects; it merely changes which file object sections will consult when asked for the filename. The current use of this FSCTL will either fail (when issued against the correct file since it does have a section) or have no meaningful effect (when applied to the wrong file).

3rdit:
@diversenok Perhaps I misread the wall of conversation that mrexodia posted in that issue. I'll review this tomorrow and make a follow up change if what you say is correct. Thanks for the clarification and help!

diversenok:
If you're interested, I have a bit more context for that conversation [in the slides](https://diversenok.github.io/slides/RomHack-2025-Image-Load-Callbacks.pdf) I made later for my conference talk on that topic. The image load debug events are very similar to the image load notification callbacks; the same tricks would work there as well.

You can observe how the debugging subsystem reopens the file by name using procmon:
image

This behavior is problematic because it doesn't guarantee that the request will end up opening the correct file. Files could've been renamed, transactions rolled back, etc. The only way (that I know) to identify whether the returned file is the same as the mapped image is via `NtAreMappedFilesTheSame`. Unfortunately, it doesn't work cross-process. And, of course, it cannot help when the debug event returns a NULL file handle.

3rdit:
@diversenok I believe even the conversation said this too so this is just my fault for not properly reading it :sweat_smile: I'll make a PR to clean this up this coming week.

PR #3780, opened 2026-01-13
Add option to detach instead of terminate when attaching to new process

### LINKED ISSUE #648: Wrong detach behavior when attaching to another process
1. Open any 2 applications.
2. Open the debugger, press `Alt + A` and attach it to the first application.
3. Again, press `Alt + A` and select the second application.

The first application will be terminated instead of detaching the debugger from it.

Btw, how are you ?? :smile:
it's been a while.

EDIT: if this one is too silly/trivial, please feel free to close here immediately.

3rdit:
Adds "Detach from process when attaching to another" setting in Engine tab. When enabled, attaching to a new process will detach from current process and leave it running instead of terminating it.

Closes #648

### DIFF
--- a/src/dbg/commands/cmd-debug-control.cpp
+++ b/src/dbg/commands/cmd-debug-control.cpp
@@ -264,8 +264,6 @@ bool cbDebugAttach(int argc, char* argv[])
             CloseHandle(hDebugLoopThread);
             hDebugLoopThread = nullptr;
         }
-
-        HistoryClear();
     }
     else
     {
@@ -361,6 +359,7 @@ bool cbDebugDetach(int argc, char* argv[])
         dputs(QT_TRANSLATE_NOOP("DBG", "Detached!"));
     _dbg_animatestop(); // Stop animating
     unlock(WAITID_RUN); // run to resume the debug loop if necessary
+    HistoryClear();
     return true;
 }
 
--- a/src/dbg/commands/cmd-debug-control.cpp
+++ b/src/dbg/commands/cmd-debug-control.cpp
@@ -252,7 +252,26 @@ bool cbDebugAttach(int argc, char* argv[])
         return false;
 
     EXCLUSIVE_ACQUIRE(LockDebugStartStop);
-    cbDebugStop(argc, argv);
+
+    // Detach instead of terminate when attaching to a new process (if setting enabled)
+    if(bIsDebugging && settingboolget("Engine", "DetachOnAttach", false))
+    {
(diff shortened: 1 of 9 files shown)


src/dbg/commands/cmd-debug-control.cpp
```diff
@@ -252,7 +252,26 @@ bool cbDebugAttach(int argc, char* argv[])
...
+        cbDebugDetach(argc, argv);
+
+        if(hDebugLoopThread)
+        {
+            WaitForSingleObject(hDebugLoopThread, INFINITE);
+            CloseHandle(hDebugLoopThread);
+            hDebugLoopThread = nullptr;
+        }
+
+        HistoryClear();
```
mrexodia: Isn't this handled inside of `cbDebugDetach`?

mrexodia:
I think overall this is useful, but there are a few considerations:

1. This is only when you attach, closing the debugger when attached does not detach automatically (at least iirc)
2. There is no UI feedback to the user. Maybe we should add the confirmation dialog on exit (and also make that save the current preference)

src/dbg/commands/cmd-debug-control.cpp
3rdit: `cbDebugDetach` doesn't call `HistoryClear`. It handles the detach operation itself, I did `HistoryClear` to match what `cbDebugStop` does, too.

mrexodia: So probably it should be moved in `cbDebugDetach` instead? I thought this was happening at the end of the debug loop...

3rdit: Resolved in https://github.com/x64dbg/x64dbg/pull/3780/commits/8fdfb68ebf207f3bc15d1d96bfb5b8c16aa0fe18

3rdit:
@mrexodia I believe for your second consideration you meant this: https://github.com/x64dbg/x64dbg/pull/3780/commits/8e9e4b9ee124f356e77f43e0588d20cae988dc4c

This shows in the event the `ShowExitConfirmation` setting is enabled only.

mrexodia:
Yeah it's already much better, but I think for users the setting might be too hidden. Probably we want:

1. On exit ask what to do (+ option to remember and never pop up again)
2. When you actually click attach in the UI ask what to do (+ option to remember and never pop up again)

For (2), the UI flow would then run `detach + attach` or `stop + attach` depending on the setting. This way the user never has to know the setting exists, because it would come up naturally when they find themselves in this situation.

PR #3786, opened 2026-01-24
Hexdump format option for log

tayxd:
##### Summary
New "hexdump" formatting option for log.
It produces a canonical hex dump output (address|hex|ascii) for a given memory address, similar to the x64dbg dump window or linux "xxd" command.
##### Motivation / Use Case
Currently, the "mem" formatter outputs raw hex strings, and "ascii" outputs raw text. However, when logging buffer data or inspecting memory via the log window, it is often necessary to see both the hex representation and the ASCII representation side-by-side along with address to identify patterns, strings, or file headers, etc.
##### Example usage
Breakpoint's "Log Text": Breakpoint hit:{\n}{hexdump;0x20;@rax+40}
image

### DIFF
--- a/src/dbg/formatfunctions.cpp
+++ b/src/dbg/formatfunctions.cpp
@@ -132,6 +132,43 @@ void FormatFunctions::Init()
         });
     });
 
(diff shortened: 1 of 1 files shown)


mrexodia:
This is a plugin feature.

Issue #3651, opened 2025-07-14
Add "copy address" hotkey (for references tab)


loqstrz:
### Feature type

Quality of life

### Describe the feature

After finding all the references, I end up having to click on each and then do a CTRL+C just to get the address...
If you do CTRL+C while in the references tab it copies the Address + Disassembly.

sumiroymallarpur-hub:
Rc chh uugfgjn hhj

PR #3792, opened 2026-02-04
Added References tab copy address shortcut

bbugdigger:
Issue: https://github.com/x64dbg/x64dbg/issues/3651

Added a References-only “Copy Address” action with Ctrl+Shift+C


mrexodia:
Thanks!

Issue #2408, opened 2020-07-27
Move all content of faq.x64dbg.com into CHM documentation


torusrxxx:

ahkrichards:
@mrexodia @torusrxxx Implemented a submodule + patch + shim generator based implementation for syncing wiki into CHM in #3794.

Let me know if you like the approach or would like a more static version or what not.

ahkrichards:
One thing to consider: I haven’t shimmed any of the wiki’s remote images into local copies. I believe this means the CHM will only show them if the viewer has internet access. If we want full offline support, we should either pull those images locally or file a follow‑up issue.

mrexodia:
From #3794:
> I think a better approach would be to move the whole wiki repository into docs/wiki and then archive the wiki repository and create a github workflow to push to the x64dbg.wiki.git repository (https://github.com/x64dbg/workflows/blob/main/.github/workflows/sync_wiki.yml). All the images could then also be put into the repository directly.

PR #3794, opened 2026-02-08
FEAT: Sync all Wiki FAQs content into CHM docs (#2408)

ahkrichards:
Feature PR #2408

## Summary
Move all content of `faq.x64dbg.com` into the CHM documentation by embedding the FAQ wiki content into the Sphinx output while keeping the existing docs toolchain intact.

## Details
- Add a build-time shim that converts `docs/faqs/Frequently-Asked-Questions.md` into RST for inclusion.
- Wire the generator into `docs/makechm.bat`.
- Preserve the wiki git project as a submodule so it can remain in its current state without duplicating tracked files; uses `docs/wiki.patch` to apply the necessary changes in `docs/makechm.bat`.
- Patch also applies MD header changes to match wiki navigation menu
- Patch includes the generated FAQ content in `docs/faqs/index.rst`.

> [!WARNING]
> This hasn't shimmed any of the wiki’s remote images into local copies. I believe this means the CHM will only show them if the viewer has internet access. If we want full offline support, we should either pull those images locally or file a follow‑up issue.

## Preview

image

image


mrexodia:
I think a better approach would be to move the whole wiki repository into `docs/wiki` and then archive the `wiki` repository and create a github workflow to push to the x64dbg.wiki.git repository (https://github.com/x64dbg/workflows/blob/main/.github/workflows/sync_wiki.yml). All the images could then also be put into the repository directly.

PR #3796, opened 2026-02-13
Fix patch import undo bug

### LINKED ISSUE #3461: Patch Applies Incorrectly When Original Byte is 0x0
I've auto-generated a `bytes.1337` file to patch a binary. It works great, however when trying to replace a `0x0` byte with `0x48`, the order is swapped around to `0x48 -> 0x0`.

In the screenshot below:
- `bytes.1337` is visible on the left
- Yellow highlights the incorrect byte replacement
- Green highlights another occurence

byte_patch_error

Is this happening because the 'original bytes' in `bytes.1337` are all `00`? Since I'm generating `bytes.1337` myself, I don't have access to the original bytes. The patches work though, just not when the original byte is `00`.

I tried changing my patch file to `000000000099A2B3:FF->48` (notice it's `FF->48` instead of `00->48`), but it imported the patch as `48->FF`.

As a workaround, I've changed my generation script to always output the same byte pairs, e.g. `48->48`:

image

I originally [accidentally posted](https://github.com/x64dbg/DotX64Dbg/issues/102) this on the DotX64Dbg repo. I received a reply stating where the issue may be:
> You can check the function responsible for importing the patch [here](https://github.com/x64dbg/x64dbg/blob/9785d1a62e104a2720c21736db8fc8c6f7311ce3/src/gui/Src/Gui/PatchDialog.cpp#L493). In [this](https://github.com/x64dbg/x64dbg/blob/9785d1a62e104a2720c21736db8fc8c6f7311ce3/src/gui/Src/Gui/PatchDialog.cpp#L607C9-L607C10) line, you can see the behavior that you have described

ali-asly101:
Closes #3461

I changed it so theres "actualbyte" thats in the IMPORTSTATUS struct, which takes a "snapshot" of the byte on import.

So it can actually compared the original byte on import, not compare it to the original byte in the patch file which can be wrong.
This basically handles the edge case error where a user can type out a wrong original byte, but correct newbyte, which causes it to take the original byte in the .1337 as the real byte and actually swaps it when it "undoes" patching, resulting in lets say a 00 original byte which was never in the file, to be replaced.

New build:
image
image

Old build:
image

Notice how it switches CC for 35? When the newbyte should have been CC, as it is.

### DIFF
--- a/src/gui/Src/Gui/PatchDialog.cpp
+++ b/src/gui/Src/Gui/PatchDialog.cpp
@@ -501,6 +501,7 @@ void PatchDialog::on_btnImport_clicked()
     {
         bool badoriginal;
         bool alreadypatched;
+        unsigned char actualbyte; // byte snapshot on import
     } IMPORTSTATUS;
     QList<QPair<DBGPATCHINFO, IMPORTSTATUS>> patchList;
     DBGPATCHINFO curPatch;
@@ -558,6 +559,7 @@ void PatchDialog::on_btnImport_clicked()
             IMPORTSTATUS status;
             status.alreadypatched = (checkbyte == newbyte);
             status.badoriginal = (checkbyte != oldbyte);
+            status.actualbyte = checkbyte;  
             if(status.alreadypatched)
                 bAlreadyDone = true;
             else if(status.badoriginal)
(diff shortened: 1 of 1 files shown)


ali-asly101:
please let me know if any problems

mrexodia:
You just need to run format.bat, otherwise I'm happy to merge it. Also change `Basically #3461` to `Closes #3461`, then it will auto-close the issue on merge.

ali-asly101:
alright just updated it

mrexodia:
Thanks!

PR #3798, opened 2026-02-14
faster search list filtering

d2k2-git:
For improving the UI experience while using the search filter this PR will adjust the typing timer values for average human typing speed. 

In addition this fixes a performance issue in `StdTableSearchList::filter`. 

I have changed also some variable names like "i" and "j" to make the code easier to understand.

### DIFF
--- a/src/gui/Src/BasicView/SearchListView.cpp
+++ b/src/gui/Src/BasicView/SearchListView.cpp
@@ -238,7 +238,7 @@ void SearchListView::searchTextEdited(const QString & text)
         if(rowCount <= 10000)
             return 0;   // seems to be OK for up to 10000 rows
         else
-            return 400; // 400 ms between keypresses should be ok for the average user typing speed.
+            return 250; // 250 ms between keypresses should be ok for the average user typing speed.
     }(mAbstractSearchList->list()->getRowCount()));
     mAbstractSearchList->unlock();
     mTypingTimer->start(); // This will fire filterEntries after interval ms.
--- a/src/gui/Src/BasicView/SearchListView.cpp
+++ b/src/gui/Src/BasicView/SearchListView.cpp
@@ -238,7 +238,7 @@ void SearchListView::searchTextEdited(const QString & text)
         if(rowCount <= 10000)
             return 0;   // seems to be OK for up to 10000 rows
         else
-            return 250; // 250 ms between keypresses should be ok for the average user typing speed.
+            return 400; // 400 ms between keypresses should be ok for the average user typing speed.
     }(mAbstractSearchList->list()->getRowCount()));
     mAbstractSearchList->unlock();
     mTypingTimer->start(); // This will fire filterEntries after interval ms.
--- a/src/gui/Src/BasicView/SearchListView.cpp
+++ b/src/gui/Src/BasicView/SearchListView.cpp
@@ -235,11 +235,8 @@ void SearchListView::searchTextEdited(const QString & text)
     mAbstractSearchList->lock();
     mTypingTimer->setInterval([](dsint rowCount)
     {
-        // These numbers are kind of arbitrarily chosen, but seem to work
         if(rowCount <= 10000)
-            return 0;
-        else if(rowCount <= 600000)
-            return 100;
+            return 0;   // seems to be OK for up to 10000 rows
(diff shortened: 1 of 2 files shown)


src/gui/Src/BasicView/StdTableSearchList.cpp
```diff
@@ -5,22 +5,33 @@ void StdTableSearchList::filter(const QString & filter, FilterType type, duint s
...
     StdIconTable* mIconList = qobject_cast<StdIconTable*>(mList);
-    mSearchList->setRowCount(0);
+
     auto rows = mList->getRowCount();
     auto columns = mList->getColumnCount();
-    for(duint i = 0, j = 0; i < rows; i++)
+
+    // collect matching row indices
+    std::vector<duint> matchingRows;
+    matchingRows.reserve(rows);
```
mrexodia: This might allocate a lot of memory and `push_back` is amortized cheap enough anyway so it shouldn't be necessary

mrexodia (changes requested):
Other than that it would be good to test with a huge PDB. Some examples I have on my drive:

- C:\Qt-online\5.15.2\msvc2019_64\bin\Qt5WebEngineCored.pdb
- D:\Games\Palworld.Early.Access\Pal\Binaries\Win64\Palworld-Win64-Shipping.pdb

[00007FFC930F0000, qt5webenginecored.dll] Loaded 4303939 symbols in 107.047s

Search for `vector` or something

src/gui/Src/BasicView/SearchListView.cpp
```diff
@@ -235,13 +235,10 @@ void SearchListView::searchTextEdited(const QString & text)
...
     {
-        // These numbers are kind of arbitrarily chosen, but seem to work
         if(rowCount <= 10000)
-            return 0;
-        else if(rowCount <= 600000)
-            return 100;
+            return 0;   // seems to be OK for up to 10000 rows
         else
-            return 350;
+            return 250; // 250 ms between keypresses should be ok for the average user typing speed.
```
ZehMatt, line 241: This threshold is too low in my opinion, I think something like 500ms is much better to be more confident that the user stopped typing, typing speed varies so its best to aim for slower users rather than going by average.

d2k2-git: might be true, in combination of the fix for the filtering performance, this timer value is not so relevant anymore. as before. higher values would be result in 500ms waiting for every search result even its not required, since the search is fast now. but as mrexodia mentioned, it shoulod be tested with extremly high row counts, my project has "only" 120k rows. i will do more testings. in the end everything else than the current dev snapshot is an improvement.

d2k2-git:
> Other than that it would be good to test with a huge PDB. Some examples I have on my drive:
> 
> * C:\Qt-online\5.15.2\msvc2019_64\bin\Qt5WebEngineCored.pdb
> * D:\Games\Palworld.Early.Access\Pal\Binaries\Win64\Palworld-Win64-Shipping.pdb
> 
> [00007FFC930F0000, qt5webenginecored.dll] Loaded 4303939 symbols in 107.047s
> 
> Search for `vector` or something

maybe i can just write a script to generate a lot of comments for every assembly line. should be then some millions of rows for testing. i will see what i can do. but already just by testing with only 120k rows there is already a noticeable improvement while typing.

src/gui/Src/BasicView/SearchListView.cpp
d2k2-git: made some test. 500ms still fine for me, allthough i think it might be too high. any other opinions? lets say we do 350ms or 400ms. it probably does not matter much in the end as long as the search itself is faster than the timer value.

mrexodia: I’m happy with 250

d2k2-git:
Benchmark for old and new "StdTableSearchList::filter"
search 700.000 strings references for "a"

old:
Execution time:  5633970900 ns -> also, ui freezes couple seconds (10-20) after search finished

new:
Execution time:  1065006000 ns

not sure if there could be done further optimizations. but with this changes its already a lot better ui experience.

mrexodia:
Let's do it, thanks a lot!

PR #3799, opened 2026-02-14
Use run trace to assist correct disassembly (experimental)

torusrxxx:
This is part of original run trace (same name as Ollydbg) feature roadmap. It saves the boundaries of executed instructions in run trace. Previously, this byte type information was unused. Now, the disassembler starts to use the byte execution type to ensure correct disassembly, especially in obfuscated code.
Before:
before
After:
after
This is currently working and experimental.

### DIFF
--- a/src/cross/widgets/Bridge.cpp
+++ b/src/cross/widgets/Bridge.cpp
@@ -94,7 +94,7 @@ DBGFUNCTIONS* DbgFunctions()
         };
         f.GetTraceRecordByteType = [](duint addr)
         {
-            return TRACERECORDBYTETYPE::Unknown;
+            return TRACERECORDBYTETYPE::TraceRecordByteTypeUnknown;
         };
         f.ModBaseFromAddr = [](duint addr) -> duint
         {
--- a/src/cross/widgets/Bridge.cpp
+++ b/src/cross/widgets/Bridge.cpp
@@ -92,6 +92,10 @@ DBGFUNCTIONS* DbgFunctions()
         {
             return 0;
         };
+        f.GetTraceRecordByteType = [](duint addr)
+        {
+            return TRACERECORDBYTETYPE::Unknown;
+        };
         f.ModBaseFromAddr = [](duint addr) -> duint
         {
             return 0;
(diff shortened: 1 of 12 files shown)


mrexodia:
Just needs some changes to the `cross` project to stub out `GetTraceRecordByteType` in the fake `DBGFUNCTIONS`, but looks good otherwise!

torusrxxx:
There are some minor details wrong. For example, two EIP arrows are on the screen. I'm on holiday, won't come back this week.

mrexodia (changes requested):
I pushed a small follow-up to fix the cross/windows CI break caused by the new  enum value colliding with the Windows SDK in unity builds.\n\nOne remaining correctness issue: the new run-trace path in  /  trusts trace byte metadata without first verifying that the *current* byte is actually a traced instruction start. In the back-disassembler we frequently scan from the middle of an instruction; if  happens to be traced, the new loop can synthesize a shorter instruction boundary, skip the real start, and return the wrong previous instruction. The trace-assisted fast path should only be taken when  /  is  or , and the lookahead should also be clamped to the remaining buffer / architectural max instruction length.

mrexodia (changes requested):
I pushed a small follow-up to fix the cross/windows CI break caused by the new `Unknown` enum value colliding with the Windows SDK in unity builds.

One remaining correctness issue: the new run-trace path in `QZydis::DisassembleBack()` / `DisassembleNext()` trusts trace byte metadata without first verifying that the *current* byte is actually a traced instruction start. In the back-disassembler we frequently scan from the middle of an instruction; if `base + addr + 1` happens to be traced, the new loop can synthesize a shorter instruction boundary, skip the real start, and return the wrong previous instruction. The trace-assisted fast path should only be taken when `GetTraceRecordByteType(base + addr)` / `GetTraceRecordByteType(base + ip)` is `InstructionHeading` or `InstructionOverlapped`, and the lookahead should also be clamped to the remaining buffer / architectural max instruction length.

mrexodia:
Ignore the earlier malformed review body from my shell quoting mistake; the later CHANGES_REQUESTED review contains the actual feedback.

mrexodia (approved):
Addressed the remaining concern by only trusting run-trace boundaries when the current byte is a traced instruction start and by clamping the lookahead to the architectural maximum / remaining buffer.

mrexodia:
I resolved the `cross` build issue and merged, thanks!

PR #3801, opened 2026-02-15
Add flush() to TaskThread to sync logs before savelog

### LINKED ISSUE #3609: Savelog command doesn't wait for printstack command to finish
### Operating System

Wine

### x64dbg Version

Mar 15 2025, 15:54:24

### Describe the issue

I am running a script to save stacktrace into a file:

```
ClearLog; printstack; savelog /tmp/stack
```

```
27 7ll stack frames (RIP = 00000001487D0CC8 , RSP = 000000000011F060 , RBP = 000000000011F100 ):
000000000011F088 myexe.00000001487D0CC8
000000000011F118 myexe.00000001487F3E3D
000000000011F338 dinput8.GetdfDIJoystick+F2D
Log have been saved as /tmp/stack
000000000011F378 myexe.000000014C1EF824
000000000011F4B8 myexe.0000000141A2673B
000000000011F4E8 myexe.000000014C1EDE81
000000000011F668 myexe.000000014C1EF7BE
```

Only first 4 lines are saved in a file, rest of the stack is printed but ignored.

### Steps to reproduce

Run the commands

### Attachments

_No response_

ali-asly101:
Closes #3609

Added a flush() method to TaskThread, which is called from the main thread inside cbSaveLog, which pauses execution of the main thread until the worker thread ( triggered by the cbInstrPrintStack via dprintf_untranslated ) finishes processing its current buffer and prints messages at the time savelog is called.
Before:
image

The log is empty because the save dialog command returns before the worker thread has finished processing its buffer.

After:
image
It prints the current call stack at the time savelog is called, ensuring that everything already in the buffer is flushed.

For now, this does not guarantee that all future call stack logs will be printed, since additional wakeup calls may occur after savelog returns. It only ensures that the buffer at the moment savelog is called is flushed.

### DIFF
--- a/src/dbg/console.cpp
+++ b/src/dbg/console.cpp
@@ -6,12 +6,16 @@
 #include "console.h"
 #include "taskthread.h"
 
+static StringConcatTaskThread_<void(*)(const std::string &)>* logTask = nullptr;
+static StringConcatTaskThread_<void(*)(const std::string &)>* logHtmlTask = nullptr;
+
 static void GuiAddLogMessageAsync(_In_z_ const char* msg)
 {
     static StringConcatTaskThread_<void(*)(const std::string &)> task([](const std::string & msg)
     {
         GuiAddLogMessage(msg.c_str());
     });
+    logTask = &task;
     task.WakeUp(msg);
 }
 
@@ -21,6 +25,7 @@ static void GuiAddLogMessageHtmlAsync(_In_z_ const char* msg)
     {
         GuiAddLogMessageHtml(msg.c_str());
     });
+    logHtmlTask = &task;
     task.WakeUp(msg);
 }
 
@@ -134,4 +139,12 @@ void dprintf_html(_In_z_ _Printf_format_string_ const char* Format, ...)
 void dlogprint_untranslated(const char* Text)
 {
     GuiAddLogMessageAsync(Text);
+}
+
+// flushing
+
+void dflush()
+{
+    if(logTask) logTask->Flush();
+    if(logHtmlTask) logHtmlTask->Flush();
 }
--- a/src/dbg/console.cpp
+++ b/src/dbg/console.cpp
@@ -7,7 +7,6 @@
 #include "taskthread.h"
 
 static StringConcatTaskThread_<void(*)(const std::string &)>* logTask = nullptr;
-static StringConcatTaskThread_<void(*)(const std::string &)>* logHtmlTask = nullptr;
 
 static void GuiAddLogMessageAsync(_In_z_ const char* msg)
 {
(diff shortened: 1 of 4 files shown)


src/dbg/console.cpp
```diff
@@ -6,12 +6,16 @@
 #include "console.h"
 #include "taskthread.h"
 
+static StringConcatTaskThread_<void(*)(const std::string &)>* logTask = nullptr;
+static StringConcatTaskThread_<void(*)(const std::string &)>* logHtmlTask = nullptr;
```
mrexodia: Shouldn't this be a single pointer instead?

mrexodia (changes requested):
I think this has some fundamental issue: the fact that GuiAddLogMessage executed does not mean the GUI got that message and handled the signal

ali-asly101:
> I think this has some fundamental issue: the fact that GuiAddLogMessage executed does not mean the GUI got that message and handled the signal

i changed the savedialog command to be a callback in the gui thread to be executed after the addlogmessage events, and now it works better since the whole callstack is printed
image

src/dbg/console.cpp
ali-asly101: yea i deleted the second pointer its useless for now

ali-asly101:
is everything here fine?

mrexodia:
> is everything here fine?

CI does not pass, so no 🙂

mrexodia:
I think this solution is fundamentally wrong. Likely something like this _should_ do the same (didn't try) without requiring changes to `taskthread.h` or other places:

```
auto hEvent = CreateEventW(nullptr, false, false, nullptr);
GuiExecuteOnGuiThreadEx([](void* event)
{
    SetEvent((HANDLE)param);
}, hEvent);
WaitForSingleObject(hEvent);
CloseHandle(hEvent);
// log flushed here
```

All the messages are pumped through the GUI thread, so once we reach `SetEvent` we know that all log messages up to that point have been processed.

ali-asly101:
> auto hEvent = CreateEventW(nullptr, false, false, nullptr);
> GuiExecuteOnGuiThreadEx([](void* event)
> {
>     SetEvent((HANDLE)param);
> }, hEvent);
> WaitForSingleObject(hEvent);
> CloseHandle(hEvent);
> // log flushed here

is adding a Flush() to taskthread acceptable? ur gui solution is genius btw it works, but i need the flush function to flush everything before moving onto the gui thread

mrexodia:
I don’t like that flush method there too much, that code is very brittle and it has worked reasonably well for years…

will need to look into this in more detail to come up with something clean.

ali-asly101:
alright thank u for ur help boss

PR #3804, opened 2026-02-19
Fix invalid pointer dereference in setfpuvalue for Opmask/ZMM registers

### LINKED ISSUE #3803: Crash in setfpuvalue: integer value dereferenced as pointer for Opmask/ZMM registers
### Operating System

Windows 10 64bit, Build 19041

### x64dbg Version

x64dbg Version: snapshot_2025-08-19_19-40

### Describe the issue

**Describe the issue:**

`setfpuvalue()` in `value.cpp` dereferences the `duint value` parameter as a pointer in the Opmask and ZMM paths:

```cpp
// Opmask path (~line 2304)
context.Opmask[registerindex] = *(ULONGLONG*)value;

// ZMM path (~line 2316)
context.ZmmRegisters[registerindex] = *(ZmmRegister_t*)value;
```

`value` comes from `valtostring(const char* string, duint value, bool silent)` which passes a plain integer, not a pointer. This causes an access violation.

All other FPU paths in the same function (x87, MMX, XMM, YMM) use `value` as a plain integer, so this looks like a copy-paste mistake

### Steps to reproduce

Steps to reproduce:

1. Open x64dbg (x64), debug any 64-bit executable (e.g. notepad.exe)
2. Wait for system breakpoint
3. Type _K0=5 in the command bar and press Enter
4. Debugger crashes with access violation

WinDbg crash log:

Image

Image

Image

Call chain:

handleAssignment -> valtostring("_K0", 5) -> setfpuvalue("K0", 5) -> *(ULONGLONG*)5 -> crash

Oblivionsage:
`setfpuvalue` was dereferencing `value` as a pointer (`*(ULONGLONG*)value`) instead of casting it directly. This crashes the debugger when setting Opmask/ZMM registers via expressions like `_K0=5`.

- Opmask: cast directly to `ULONGLONG`
- ZMM: zero-init + `memcpy` since `ZmmRegister_t` is larger than `duint`

Fixes : #3803

### DIFF
--- a/src/dbg/value.cpp
+++ b/src/dbg/value.cpp
@@ -2285,7 +2285,7 @@ static void setfpuvalue(const char* string, duint value)
             }
             else
             {
-                context.Opmask[registerindex] = *(ULONGLONG*)value;
+                context.Opmask[registerindex] = (ULONGLONG)value;
                 SetAVX512Context(hActiveThread, &context);
             }
         }


src/dbg/value.cpp
```diff
@@ -2303,7 +2303,9 @@ static void setfpuvalue(const char* string, duint value)
             }
             else
             {
-                context.ZmmRegisters[registerindex] = *(ZmmRegister_t*)value;
```
mrexodia, line 2306: This is likely incorrect, @torusrxxx did some hacks here and I think it is being passes by pointer for ZMM registers.

mrexodia (changes requested):

src/dbg/value.cpp
Oblivionsage: Makes sense for ZMM, I'll revert that change. The Opmask fix is still valid though , `_K0=5` crashes the debugger since ULONGLONG fits in duint on x64. Want me to update the PR with just the Opmask fix?

torusrxxx:
x32dbg limits `duint` to 32-bit but the opmask register is still 64-bit on x32dbg. Varying pass-by-value and pass-by-pointer styles for x64dbg and x32dbg should be avoided, so all AVX512 registers are passed by pointers including the opmask register.

mrexodia:
I think we need to change this function to take some `union RegisterValue*` instead and pass everything by pointer, because this is very unclear...

mrexodia:
@torusrxxx what do you want to do with this?

torusrxxx:
Keep the pointer dereference. Fix the caller.

mrexodia:
Thanks for the PR @Oblivionsage, I will address it in a different way on development directly.

PR #3809, opened 2026-03-20
Fix #3467: can't load DLL with no entrypoint

### LINKED ISSUE #3467: x64dbg can't load DLL with no entrypoint
### Operating System

Windows 11 23H2 22631.4317

### x64dbg Version

Oct 18 2024

### Describe the issue

It is possible to create a DLL without an entrypoint. x64dbg fails to load such a DLL, but if I call `LoadLibraryW()` to load it _with exactly the same path_, my program doesn't fail.

I actually tried compiling https://github.com/x64dbg/x64dbg/blob/9785d1a62e104a2720c21736db8fc8c6f7311ce3/src/loaddll/loaddll.cpp with the path hard-coded to my DLL but it did not fail. Not sure what is special about the prebuilt ones in x64dbg.

### Steps to reproduce

1. Create a source file:
    ```c
    // a.c
    int foo(void) { return 1; }
    ```
2. In MSYS2 UCRT64 shell, compile it with gcc:
    > gcc a.c -shared -nostdlib -o a.dll
    ld.exe: warning: cannot find entry symbol DllMainCRTStartup; not setting start address

3. Load this file with x64dbg. It says 'Failed to load DLL (LastError: 193)' which is `ERROR_BAD_EXE_FORMAT`.

### Attachments

_No response_

PedroLlussa:
x64dbg tries to set a breakpoint on the entry point. However, DLL files may not have an entry point defined, therefore we need to check if it is defined before setting a breakpoint so that the file can be properly loaded. This fix implements a guard that prevents the breakpoint from being set if there is no entry point.

### DIFF
--- a/src/dbg/debugger.cpp
+++ b/src/dbg/debugger.cpp
@@ -1901,7 +1901,7 @@ static void cbLoadDll(LOAD_DLL_DEBUG_INFO* LoadDll)
         bIsDebuggingThis = true;
         pDebuggedBase = (duint)base;
         DbCheckHash(ModContentHashFromAddr(pDebuggedBase)); //Check hash mismatch
-        if(settingboolget("Events", "EntryBreakpoint", true))
+        if(pDebuggedEntry != 0 && settingboolget("Events", "EntryBreakpoint", true))
         {
             bAlreadySetEntry = true;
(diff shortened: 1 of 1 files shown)


mrexodia:
Thanks!

PR #3813, opened 2026-03-22
Crash when pasting/loading script due to wrong Order of evaluation

OldGamesCracking:
The program can crash for some special scripts.

I'm not fully sure when or why this happens, but it's possible that `fromIp` starts from `maxIp - 1`. In this case it loops one time and then it crashes on the next iteration for 'obvious reasons'.

The fix is simple, change the order of evaluation.

Minimal example to provoke the crash:

```
jmp error
error:
end:
```

### DIFF
--- a/src/dbg/simplescript.cpp
+++ b/src/dbg/simplescript.cpp
@@ -187,7 +187,10 @@ static bool scriptCreateLineMap(const char* filename, bool gui)
     if(*temp)
     {
         entry = {};
-        strcpy_s(entry.raw, temp);
+        int add = 0;
+        while (isspace(temp[add]))
+            add++;
+        strcpy_s(entry.raw, temp + add);
         scriptLineMap.push_back(entry);
     }
     int linemapsize = (int)scriptLineMap.size();
--- a/src/dbg/simplescript.cpp
+++ b/src/dbg/simplescript.cpp
@@ -144,6 +144,7 @@ static bool scriptCreateLineMap(const char* filename, bool gui)
     auto len = filedata.length();
     String temp;
     LINEMAPENTRY entry = {};
+    size_t rawElements = sizeof(LINEMAPENTRY::raw) / sizeof(LINEMAPENTRY::raw[0]);
     scriptLineMap.clear();
     for(size_t i = 0; i < len; i++) //make raw line map
     {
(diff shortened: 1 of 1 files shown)


mrexodia:
@codex

mrexodia:
@OldGamesCracking I appreciate you fixing things here, but that code needs to be rewritten in terms of `std::string` vs hardcoded buffer lengths

OldGamesCracking:
> @OldGamesCracking I appreciate you fixing things here, but that code needs to be rewritten in terms of `std::string` vs hardcoded buffer lengths

Ok, I can try it. Are there already some helper methods for the 'lstrip/ltrim'-part? I don't want to hardcode a `temp.erase ...` each time.

OldGamesCracking:
I have written a preliminary version using std::string

mrexodia:
I was thinking to redo this:

```cpp
struct LINEMAPENTRY
{
    SCRIPTLINETYPE type;
    char raw[256];
    union
    {
        char command[256];
        SCRIPTBRANCH branch;
        char label[256];
        char comment[256];
    } u;
};
```

The `union` is kinda hacky I think when using `std::string` we can pay the overhead of the empty structures do maybe something like this:

```cpp
struct LINEMAPENTRY
{
    SCRIPTLINETYPE type;
    String raw;
    // NOTE: one of these
    String command;
    SCRIPTBRANCH branch;
    String label;
    String comment;
};
```

There are helpers in `StringUtils::` for trimming/stripping not sure if they allow in-place but could be added.

OldGamesCracking:
> I was thinking to redo this:

I guess it makes sense to replace the char buffers with a String.
~~Are unions discouraged in C++ or why do you want to remove it?~~ (because of alloc/free stuff I guess)

Besides that I guess my changes are done now, it won't get much better without changing anything outside `scriptCreateLineMap` (stuff inside LINEMAPENTRY etc.)

By the way, I f*cked up and forgot to rebase and make a second pull request. I have accidentally cramped two different issues in one:

1. Crash (original issue)
2. 'Swallowed' Jumps as last instruction of the script (with preceeding whitespace and without trailing newline) like follows:

```
loop:
    jmp loop
```

In this scenario  the `jmp` was not displayed.

mrexodia:
I merged #3814 but this is now conflicting...

mrexodia:
I merged it manually, let me know if everything works now as expected @OldGamesCracking

PR #3814, opened 2026-03-22
Use more std::string in simplescript

OldGamesCracking:
As discussed in #3813 I gave it a shot and moved `LINEMAPENTRY` from using char-buffers to Strings.
Fingers crossed, nothing broke in the process :)

~~Slight changes/adaptions that were made:~~
- ~~Also allow ";" to be used as appended line comment~~

### DIFF
--- a/src/dbg/simplescript.cpp
+++ b/src/dbg/simplescript.cpp
@@ -187,7 +187,10 @@ static bool scriptCreateLineMap(const char* filename, bool gui)
     if(*temp)
     {
         entry = {};
-        strcpy_s(entry.raw, temp);
+        int add = 0;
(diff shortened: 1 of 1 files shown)


mrexodia:
The `;` is already reserved for executing multiple commands so using it for a comment isn't possible (eg `push rcx; push rbx` is a valid script command that would now become invalid)

OldGamesCracking:
> The `;` is already reserved for executing multiple commands so using it for a comment isn't possible (eg `push rcx; push rbx` is a valid script command that would now become invalid)

Makes sense, removed.

mrexodia:
Apologies for the delay, thanks for the contribution!

PR #3815, opened 2026-03-23
Add patch comments

### LINKED ISSUE #3783: Add comments to .1337 file that can be viewed in the patch window
### Feature type

Quality of life

### Describe the feature

When you reverse something, it may be useful to not only be able to comment in the main CPU window, but also in the patch file so that when you load the .1337 file, you instantly see where is what directly within the patching window. 

You then don't need to click on every single patch to then find the comment and only get an understanding this way.

Would save you some time on a large reversing project.

codereversing:
Closes #3783

This adds the ability to attach comments to patches, both for a module and for individual addresses. This is done by introducing a "Comments" text box to the current Patches dialog where a user can now type in notes/comments about what the patch changes are doing. By default this text box will be disabled, but enabled when a comment is eligible to be made (there is a module with patches present).

A few screenshots of what this ends up looking like:

image

image

In the `.1337` file, each comment starts with a `;` and is either it's own line preceding the module name in the case of a module-level comment, or is added after the patch information for an address-level comment (example below). These changes remain backwards compatible with the existing `.1337` format parsing so older versions of x64dbg can still parse `.1337` files that have comments embedded in them.

At a high-level, this feature works by keeping track of three things:
- Which comment type (if any) is active and should be saved/displayed: `mActiveCommentType`.
- A map of module name to a comment for that module: `mModulePatchComments`.
- A map of module name to another map that maps each byte patch to a comment: `mAddressPatchComments`.

The last one is a bit more complicated so I'll try to explain my reasoning here... Comments are separate from the underlying `DBGPATCHINFO` for a patch, but we need a way to tie them together and keep them in sync despite them not knowing of each other. The most convenient way I could think of to do this was to tie the text representation of the patch, i.e. `000000000011F530:65->90` to a comment. This ends up making it a lot simpler and can give a unique representation for every patch.

Take this example `.1337` file with some comments:
```
;My module-level comment
>ntdll.dll
000000000011F530:65->90
000000000011F531:48->90 ;Comment for second entry
000000000011F532:8B->90
000000000011F533:0C->90
000000000011F534:25->90 ;Comment for fifth entry
000000000011F535:60->90
000000000011F536:00->90
000000000011F537:00->90
000000000011F538:00->90
```

This would have
```
mModulePatchComments:
Key = "ntdll.dll", Value = "My module-level comment"
```
```
mAddressPatchComments:
Key = "ntdll.dll",
Value = 
    Key = "000000000011F531:48->90", Value = "Comment for second entry"
    Key = "000000000011F534:25->90", Value = "Comment for fifth entry"
```

This mapping logic works the same for imported patches from `.1337` files or patches created at runtime. This map is what is used to save/load comment information and what is referenced when doing an export.

A lot of the other UI-related code changes are mostly for handling edge cases related to the UI:
- Navigating to a new item in the Modules or Patches list should save the comment that is being navigated away from and then load the comment for the newly selected item.
- Comments should be saved/loaded on selection changes (keyboard navigation or clicking). But they also need to be saved/loaded on item clicks because a user can click already selected items in the Modules and Patches windows as well.
- Comments should persist if a user closes and re-opens the Patches window.
- If a user is entering a comment then it needs to be saved on an export or window close. This is separate from the auto-saving that's done when a user clicks to another list item.
- Restoring all patch bytes in all modules should disable the comments window because there is nothing to comment on.

Its been a while since I've done Qt development so I tried to make the UI state management changes as small as possible, but happy to hear suggestions if this can be improved.

### DIFF
--- a/src/gui/Src/Gui/PatchDialog.cpp
+++ b/src/gui/Src/Gui/PatchDialog.cpp
@@ -809,13 +809,12 @@ void PatchDialog::showEvent(QShowEvent* event)
 void PatchDialog::closeEvent(QCloseEvent* event)
 {
     Q_UNUSED(event);
-    if(!isEligibleForComment(mActiveCommentType))
-        return;
-
-    if(isCommentForModule(mActiveCommentType))
-        syncComment(nullptr, ui->listModules->currentItem(), mActiveCommentType, mActiveCommentType);
+    if(isCommentUninitialized(mActiveCommentType) || isCommentStale(mActiveCommentType))
+        syncComment(nullptr, getSelectedOrFirst(ui->listModules), mActiveCommentType, mActiveCommentType);
+    else if(isCommentForModule(mActiveCommentType))
+        syncComment(nullptr, getSelectedOrFirst(ui->listModules), mActiveCommentType, mActiveCommentType);
     else if(isCommentForAddress(mActiveCommentType))
-        syncComment(nullptr, ui->listPatches->currentItem(), mActiveCommentType, mActiveCommentType);
+        syncComment(nullptr, getSelectedOrFirst(ui->listPatches), mActiveCommentType, mActiveCommentType);
 
     mActiveCommentType = ActiveCommentType::Stale;
 }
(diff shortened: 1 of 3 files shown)


mrexodia:
There is already support for comments in the database, we should use that instead of storing additional state in the GUI.

codereversing:
I'll take a look at `DbgSetCommentAt` / `DbgGetCommentAt`. The changes will change enough things in this PR that it will be pretty different -- I'm going to close this and re-open another when they're ready.

PR #3821, opened 2026-04-13
add initial Linux debugger with ElfBug ptrace engine and cross widget support

3rdit:
This PR introduces the initial implementation of a Linux x64dbg implementation that leverages the cross-platform widgets through a custom "ElfBug" ptrace-based debug engine.

Currently supports process launch, software breakpoints, single-stepping, register/memory access and multi-thread tracing. The engine is a static library.

Some widgets, especially RegistersView, needed to be adapted to allow for cross-platform functionality with ifdef guards and so on, but both work.

After this, the next steps would be to continue creating the interface and features of the CPU page to match Windows functionality.


mrexodia:
@codex

3rdit:
image

For historical context: This is this version of it!

justanotheranonymoususer:
Linux before ARM64 windows? wow!

mrexodia:
The problem is that ARM64 on windows is used by 3 people in the world in total 😆

PR #3826, opened 2026-04-16
ElfBug - initial architecture detect (reject everything else for now)

3rdit:
Added a small `ProcessArch` module to identify the architecture of the tracee and reject i386/anything else for now. The goal being to eventually support the architecture. Also fix minor quirks.


mrexodia:
@codex

PR #3839, opened 2026-04-21
Added CPUStack (Linux)

3rdit:
_This PR's branch name is not really relevant._

- Adds the CPUStack panel to the CPU tab, next to the HexDump, the right-click context menu all works as expected, the goal being to finish it within this PR andn get it out of the way entirely.
- Other ElfBug general bug fixes 

- No stack frame visuals like Windows yet, and some other features Windows has I haven't added in this PR

Looks like this:
image


mrexodia:
@codex

src/gui/Src/Gui/WordEditDialog.cpp
```diff
@@ -172,8 +173,8 @@ void WordEditDialog::on_signedLineEdit_textEdited(const QString & arg1)
 
 void WordEditDialog::on_unsignedLineEdit_textEdited(const QString & arg1)
 {
-    LONGLONG value;
-    if(sscanf_s(arg1.toUtf8().constData(), "%llu", &value) == 1)
+    unsigned long long value;
+    if(sscanf(arg1.toUtf8().constData(), "%llu", &value) == 1)
```
mrexodia: I would change this to `uint64_t` and use `PRIu64` so it's properly standard-compliant

src/gui/Src/Gui/WordEditDialog.cpp
```diff
@@ -90,7 +91,7 @@ void WordEditDialog::expressionChanged(bool validExpression, bool validPointer,
         unsigned char* word = (unsigned char*)&mWord;
         // ascii
         int asciiWidth = 0;
-#ifdef _WIN64
+#if defined(_WIN64) || defined(__x86_64__) || defined(__aarch64__)
```
mrexodia: I think this should be based on `sizeof(duint)` instead

src/gui/Src/Gui/WordEditDialog.cpp
```diff
@@ -156,8 +157,8 @@ void WordEditDialog::expressionChanged(bool validExpression, bool validPointer,
 
 void WordEditDialog::on_signedLineEdit_textEdited(const QString & arg1)
 {
-    LONGLONG value;
-    if(sscanf_s(arg1.toUtf8().constData(), "%lld", &value) == 1)
+    long long value;
+    if(sscanf(arg1.toUtf8().constData(), "%lld", &value) == 1)
```
mrexodia: same as below

src/dbg/types.h
```diff
@@ -153,7 +153,7 @@ namespace Types
         void Enum(std::vector<Summary> & typeList) const;
         std::string StructUnionPtrType(const std::string & alias) const;
 
-private:
+    private:
```
mrexodia: This is sus, might be an AStyle bug 😩

3rdit: I was using the wrong version :grin:

mrexodia:
@codex

mrexodia:
@codex

PR #3845, opened 2026-04-27
Add RtlExitUserProcess

yinsel:


mrexodia:
Thanks, but it is not used by x64dbg.

PR #3848, opened 2026-05-07
Add module rva address display mode in CPU

ZehMatt:
Clicking two times on the address column now uses a module rva mode:
image

For addresses that have no module associated it will display a question mark:
image

For me this simplifies working with binaries that have dynamic image base enabled, sometimes I want to copy a couple rows of the disassembly to keep track of things and if it holds the absolute address it was rather annoying to find location again with a new base address, also simplifies documenting disassembly in reports, typically don't want to use absolute address on those.

It also preserves the mode now rather than resetting it when navigating to another module or memory page, the relative mode can now be used to quickly compute the relative jump address with this, enable relative at where you jump, navigate to where you want the jump head to, and thats your relative address.


mrexodia:
@codex

PR #3853, opened 2026-05-14
feat: assemble multiple instructions in the *Assemble* window

Windows81:
## Summary

Should be able to use the *Assemble* GUI window to assemble multiple sequential instructions separated by semicolons.

## Why?

I write patching guides which require the user to input instructions one at a time.  Adding support for parsing multiple instructions at a time will save poeple's time making sure that their lines are in order.

## Tests?

None as of 2026-05-14 08:41 UTC.

### DIFF
--- a/src/dbg/assemble.cpp
+++ b/src/dbg/assemble.cpp
@@ -137,12 +137,14 @@ bool assembleat(duint addr, const char* instruction, int* size, char* error, boo
                 // Return false if an unrelated error has occured
                 return false;
             }
-            
+
             if(!nextInstruction)
             {
                 break;
             }
+         
             currentInstruction = nextInstruction + 1;
+            destIndex += destSize;
         }
     }
     else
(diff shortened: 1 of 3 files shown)


mrexodia:
Thanks, but I do not want to merge this feature. You can use https://ramensoftware.com/multimate-assembler

Windows81:
Thank you for the prompt response.

Would you be able to outline why multi-instruction support shouldn't be added to x64dbg?

Best.

mrexodia:
Because doing it properly requires a lot of work and not some clanker-maxxing. That work was already put into a plugin that you can use super easily today and has worked for years.

Windows81:
Makes sense. 

I wrote the code completely on my own without any LLM support.  However, I'm not too acquainted with the x64dbg codebase, as it's only my second contribution. 

Thank you for making me aware of that plugin.

mrexodia:
If you want to spend significant time on porting the full feature set of the plugin I would be willing to take it as a contribution. Things like labels should be supported properly.

Issue #1937, opened 2018-04-25
Shufps instruction, there is a better way to display


flarejune:
the Shufps instruction, there is a better way to display: 
for example:
Shufps xmm0, xmm1,0x4E 
can be displayed as: 
Shufps xmm0, xmm1,1032

For example, the above 0x4E, 
the binary is 01001110, divided into 2 groups are represented as: 
01 (1) 00 (0) 11 (3) 10 (2)
After representing the Shuf, 
the numerical exchange position of the XMM register. 
The final results: 
xmm0[1] - > xmm0[3] 
xmm0[0] - > xmm0[2] 
xmm1[3] - > xmm0[1] 
xmm1[2] - > xmm0[0] 
Code for calculation:
v = 0x4E;
v1 = v & 0x3;
v2 = (v & 0xC) >> 2;
v3 = (v & 0x30) >> 4;
v4 = (v & 0xC0) >> 6;
Result:{v4}{v3}{v2}{v1} = 1032

flarejune:
string to shufps imm:
s = "1032";
imm = s[3] - '0' | (s[2] - '0') << 2 | (s[1] - '0') << 4 | (s[0] - '0') << 6;

torusrxxx:
Doing this would also mean assemblers should support such notation.
However, this is not the case. So, this feature is probably not suitable
for x64dbg.

flarejune:
When inputting the instruction "shufps" to the assembly,
If the "imm" string length is 4, need string conversion processing
The default length is 1 (does not contain "0x"),Can clearly distinguish

IssuehuntBot:
@0maxxam0 funded this issue with $6. [Visit this issue on Issuehunt](https://issuehunt.io/repos/33791743/issues/1937)

mrexodia:
Not planned.

PR #3854, opened 2026-05-16
feat(disasm): display SHUFPS immediate as lane pattern

ncthuc2004:
## Summary
- format `SHUFPS` control immediates as lane-pattern digits (`abcd`) instead of raw hex in disassembly text
- expose the same lane-pattern format in operand text for immediate operands
- keep behavior scoped to `SHUFPS` so unrelated instructions are unaffected

## Why
Issue #1937 asks for a clearer representation of shuffle control values.  
Example: `SHUFPS xmm0, xmm1, 0x4E` now displays as `SHUFPS xmm0, xmm1, 1032`, which directly reflects lane selection.

## Notes
- lane pattern is derived from 2-bit selector fields in immediate byte: bits `[7:6][5:4][3:2][1:0]`
- no broad formatter behavior changes; this is an instruction-specific display improvement

### DIFF
--- a/src/gui/Src/Gui/CPUDisassembly.cpp
+++ b/src/gui/Src/Gui/CPUDisassembly.cpp
@@ -27,6 +27,30 @@
 #include "BrowseDialog.h"
 #include "Tracer/TraceBrowser.h"
 
+static QString NormalizeShufpsLanePatternExpression(const QString & expression)
+{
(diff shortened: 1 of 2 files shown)


mrexodia:
> Doing this would also mean assemblers should support such notation.

What happens if you press `space` on the instruction, can it be reassembled?

PR #3860, opened 2026-05-20
Move toggle FPU button into menu

### LINKED ISSUE #2621: Remove "Hide FPU" button and add it in to menu
That button is placed at the title bar of registers view, but it doesn't read like a title. Either remove it or rename it "Registers (FPU)" like OllyDbg.

XanderSteyn:
## Issue

The CPU/Registers and Trace views expose a "Change view" button for toggling FPU visibility. This is redundant because the same functionality already exists in the registers context menu (onChangeFPUViewAction), and the button introduces an inconsistent element in the view header.

Fixes #2621.

## Changes

**CPU / Registers View**

- Removed the "Hide/Show FPU" button from the registers view header and its associated UI wiring.
- Kept the existing context menu action as the single entry point for toggling FPU visibility.

**Trace View**

- Removed the "Hide/Show FPU" button from the trace registers view header and its associated UI wiring.
- Fixed missing right-click context menu in trace registers view by ensuring it is executed at the cursor position.

## Result

FPU visibility is now toggled only via the context menu in both CPU/Register and Trace views, which removes redundancy and keeps UI behavior consistent.

### DIFF
--- a/src/gui/Src/Gui/CPUWidget.cpp
+++ b/src/gui/Src/Gui/CPUWidget.cpp
@@ -69,14 +69,9 @@ CPUWidget::CPUWidget(Architecture* architecture, QWidget* parent)
     mGeneralRegs->setFrameShape(QFrame::NoFrame);
     mGeneralRegs->ShowFPU(true);
 
-    QPushButton* button_changeview = new QPushButton("", this);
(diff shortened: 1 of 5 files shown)


mrexodia:
Cheers!

Issue #3859, opened 2026-05-19
Reproducible crash in TraceFileReader


meck-gd:
### Operating System

Windows 11 x64, Build 26100

### x64dbg Version

Apr 20, 2026

### Describe the issue

I found a problem where x64dbg reproducibly crashes due to an access violation when scrolling through a recorded trace.

The crash seems to happen in this line: https://github.com/x64dbg/x64dbg/blob/46fed4403e1a69139535d0d73405a6eebd59aef1/src/gui/Src/Tracer/TraceFileReader.cpp#L359
The output pointer (3rd argument) that the API tries to write to is invalid (it's not NULL, but apparently an unmapped heap pointer).

### Steps to reproduce

1. Load the attached trace64.
2. Scroll up and down across the whole trace range until the problem occurs (for me, around 10-15 times, or when the process is about to reach 1 GB of usage).

I can't provide a crash dump, but I hope you'll be able to reproduce it with just the trace file.

### Attachments

[trace-20260519-135541.zip](https://github.com/user-attachments/files/28017186/trace-20260519-135541.zip)

torusrxxx:
I haven't reproduced this crash by dragging around the scrollbar for 5 minutes. But I think this code needs to be completely rewritten. It should use a linked list to put recently used pages on the top, rather than sort on access time.

meck-gd:
> I haven't reproduced this crash by dragging around the scrollbar for 5 minutes. But I think this code needs to be completely rewritten. It should use a linked list to put recently used pages on the top, rather than sort on access time.

Perhaps it's relevant to mention that I'm using a VM with 4 GB of total memory. Since available memory seems to have a direct impact on the page caching mechanism.

Edit: Yes, it's relevant. I bumped up the VM to 16 GB and the fault does not occur.

torusrxxx:
fixed in #3863

PR #3863, opened 2026-05-26
Fix memory fault in trace reader when memory status change

torusrxxx:
Fixed memory fault #3859. Memory status change can be simulated. I still plan to completely rewrite this but don't have time.


torusrxxx:
Do you think it's a good idea to try Rust in this project? It might fix many untested memory faults like this while rewriting and reduce memory consumption (currently we don't use C++20 `string_view`).

mrexodia:
Personally, I do not like rust and linking it with an existing C++ project is a pain. I want to do the C++ upgrade, but I need to measure it carefully because it impacts compile times severely. C++17 is enough for `string_view` and there are many third-party implementations of it for lower C++ versions as well we could use.

PR #3867, opened 2026-05-29
Add option to allow exporting trace files in the program directory

g0th1c54e4:
This PR adds a new option that allows users to export trace files into the program directory by default instead of the database (DB) directory.

Personally, I often find it inconvenient that x64dbg always initializes the trace export path inside the DB directory. I usually have to manually modify the path every time in order to export the files into the directory where I actually want to store them. Compared to exporting into the DB directory, exporting alongside the target application may be more convenient for organizing related files. I can also imagine (perhaps) that other users may feel the same way, since trace exports for all applications are otherwise placed under the same DB directory.

Based on this, I came up with a compromise solution: add an option which, when enabled, initializes export paths in the program directory instead of the DB directory.

At the moment, this PR only affects trace file exporting, which is already sufficient for my own use case. However, I intentionally avoided limiting the naming specifically to trace exporting, since similar changes may also be useful in other places in the future.


torusrxxx:
That's not working if the program is installed in "Program files" or read-only paths. Probably what's needed is more than a boolean configuration, it should remember the export path.

g0th1c54e4:
> That's not working if the program is installed in "Program files" or read-only paths.

Oh, yes — I completely forgot about this specific case.

> it should remember the export path.

That approach is indeed better overall. I'll try to rework the implementation tomorrow.

mrexodia:
There is already an existing setting for creating the database in the program directory. If you enable this it _should_ already put the trace files there as well, but if not that can be changed instead of adding a second option.

g0th1c54e4:
> There is already an existing setting for creating the database in the program directory.

Yes, I was aware of that option, but initially I was overly focused on the fact that it was limited specifically to "database".

At the moment, I think there may be a better approach. As @torusrxxx suggested, I could instead remember the user's previously selected export path. Different export features would manage their own last-used paths independently — for example, "Export Trace File" and "Export Memory Data" would each maintain their own path history without affecting each other.

This approach would probably give users more flexibility and also scale much better for future export-related features.

The only thing I'm still unsure about is whether this behavior should be controlled by an additional option to enable or disable path remembering. Without an explicit option, the behavior change would take effect immediately, and some users might find that confusing — although I may just be overthinking that part.

mrexodia:
I think saving databases/traces/autorun scripts/whatever to the program directory makes sense as a setting. If we remember the previous folder you can end up in a confusing situation where you save to `C:\programA` when debugging programB. This is more confusing to me than saving to a consistent location depending on the user's preference.

For a while I have been thinking how to nicely model per-database settings in a transparent way, which would solve this issue somewhat but this would require a much broader effort.

g0th1c54e4:
> For a while I have been thinking how to nicely model per-database settings in a transparent way, which would solve this issue somewhat but this would require a much broader effort.


Hey, after thinking about it some more, I think this approach would indeed solve the problem to a large extent.

The conceptual solution I have in mind right now is the following: when the user needs to choose a save location for a trace file, x64dbg would first try to retrieve the previously used path from a project-level configuration. If no such path exists, it would fall back to using the program directory as the default location.

For use cases such as "changing the default export path for trace files" and similar features, I believe they should generally be stored as project-level settings rather than global settings. This would avoid situations where a user debugs Program A, then later debugs Program B and finds the export path still pointing to Program A's directory. The most natural place to store such project-specific settings would probably be the program's own debugging database (`.dd32` / `.dd64`).

As for cases where the program directory is located under "Program Files" or another read-only location, I see the program directory merely as a sensible default that works in most situations. In the remaining cases, users would still need to choose a different location themselves.

mrexodia:
If you want to database-specific settings a shot I would do that first as a separate PR, but it's going to be quite difficult and a lot of work. For example when you open the settings, we need to clearly display whether we are changing global settings of per-database settings. The `BridgeSettingXyz` APIs need to work transparently (eg they need to become aware of the scope without changing the interface for consumers) and we need to think about storage. The database is conceptually the right place, but I suspect you very often want to get settings before the database was loaded, so maybe a mechanism similar to the command line caching is needed (or maybe it needs to be unified)?

For this change specifically I think it's okay to base the default on the db/program directory depending on the settings and I would merge it as-is, without requiring the per-database settings. That can be a later addition fairly easily.

g0th1c54e4:
I accidentally messed up the commits. I’ll close this PR myself later and open a new one.

PR #3871, opened 2026-06-01
feat #2259: add CB_DBOPERATION callbacks for plugin database synchronization

### LINKED ISSUE #2259: [SDK - Feature request] Add callbacks for labels + comments
Hi, first of all thanks for this awesome tool.

In order to be ablefor my plugin to catch function naming + comments adding, I would require the SDK to expose these events with the appropriate callback functions

From the doc : https://help.x64dbg.com/en/latest/developers/plugins/Callbacks/index.html

Current callbacks:

https://github.com/x64dbg/x64dbg/blob/1e7d4cec1a84e61f83b25857f01466cee108ff4b/src/dbg/plugin_loader.cpp#L181-L210

I'm very new to this SDK but I'd happily give a hand to implement this feature request if someone could provide me some basic infos on the workflow to do so (or pointers to the doc).

Cheers!

rafaelrfreitas:
Extended the database operation callback system to notify plugins when labels, comments, bookmarks, functions, loops and arguments are added or removed, passing the relevant data by value.

Implemented bulk callbacks, meaning commands that cause massive amounts of database changes (eg: auto analysis, clearing all or delete range) will be batched together in a single plugin callback.

Added basic unit tests for all implemented database callback types, asserting that all fields match the expected values.

Closes #2259 
Co-authored-by: Pedro Filipe Llussa Tavares <pedro.llussa.tavares@tecnico.ulisboa.pt>


src/dbg/_exports.cpp
```diff
@@ -1491,18 +1496,21 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
...
     case DBG_DELETE_LABEL_RANGE:
     {
+        DbCallbackBatcher batcher;
         LabelDelRange((duint)param1, (duint)param2, true);
     }
     break;
 
     case DBG_DELETE_BOOKMARK_RANGE:
     {
+        DbCallbackBatcher batcher;
```
mrexodia: These should be moved inside the function, not on the export. Otherwise internal range deletions will spam the callbacks.

src/dbg/argument.cpp
```diff
@@ -75,7 +76,12 @@ bool ArgumentAdd(duint Start, duint End, bool Manual, duint InstructionCount)
     argument.manual = Manual;
     argument.instructioncount = InstructionCount;
 
-    return arguments.Add(argument);
+    if(arguments.Add(argument))
+    {
+        DbCbNotifyArgument(DbOperationType::Add, argument.modhash, argument.start, argument.end, InstructionCount, Manual);
```
mrexodia: From what I remember the object for `arguments` is generic, can't we do the notification mechanism in there transparently?

mrexodia (changes requested):
I think this needs some more discussion about the implementation. If we delete 1M comments, what does the memory usage look like?

rafaelrfreitas:
> I think this needs some more discussion about the implementation. If we delete 1M comments, what does the memory usage look like?

Memory usage is mainly dictated by the DB_MAX_CB_BATCH_SIZE macro. Two vectors (one for the actual operation structs and one for saving strings) are reserved with this size. We could reduce the memory footprint if we assured that an addition/removal of a comment/label (at the same address/module) couldn't coexist in the same batch callback. (Assuming that this can happen is why the string vector exists and why we pass every operation by value. This way, the entire operation notification is preserved as is, and when a plugin loops through the callback it will have all the correct information).

If a user calls 'commentclear' and has 1M comments removed, the memory usage would remain the same as if he had DB_MAX_CB_BATCH_SIZE comments, as the deletions would be distributed across various batch callbacks. REMOVE operations don't carry their relevant data, only the ADD operations. If however, something created 1M comments (in a batch), then the usage would be bigger depending on the size of the actual comments' strings, although the callbacks would still be distributed.

We appreciate further feedback.

src/dbg/_exports.cpp
rafaelrfreitas: You're right, and this will be corrected. The main catch with moving the definition inside the functions is how it affects things like the analysis SetMarkers routines, where we already instantiate a DbCallbackBatcher at the top level to capture all function additions and removals.

If we add another batcher inside FunctionDelRange itself, we end up with two batcher objects on the stack at the same time. While it won't duplicate the callbacks, it does mean we're wasting memory by allocating extra vectors that aren't actually needed.

We can fix this if the inner batcher checks if an active batcher already exists on the thread. If it does, it can simply re-use the existing one and skip its own vector reservations entirely.

src/dbg/argument.cpp
rafaelrfreitas: Do you mean we should wrap Add/Delete locally inside the Arguments struct? If this is the case, wouldn't it make more sense to do this for all other types as well? Thanks for the question.

src/dbg/argument.cpp
```diff
@@ -147,7 +167,15 @@ void ArgumentCacheLoad(JSON Root)
 
 void ArgumentClear()
 {
+    std::vector<ARGUMENTSINFO> allArguments;
+    ArgumentGetList(allArguments);
```
mrexodia: I was referring to this with the memory usage, doesn't this just copy all of the arguments?

src/dbg/bookmark.cpp
```diff
@@ -58,7 +78,15 @@ bool BookmarkEnum(BOOKMARKSINFO* List, size_t* Size)
 
 void BookmarkClear()
 {
+    std::vector<BOOKMARKSINFO> allBookmarks;
+    BookmarkGetList(allBookmarks);
```
mrexodia: Same, copies all the bookmarks

src/dbg/comment.cpp
```diff
@@ -85,7 +105,15 @@ bool CommentEnum(COMMENTSINFO* List, size_t* Size)
 
 void CommentClear()
 {
+    std::vector<COMMENTSINFO> allComments;
+    CommentGetList(allComments);
```
mrexodia: etc

src/dbg/argument.cpp
rafaelrfreitas: Oh, yes of course. That was a complete oversight. We will fix this, using a similar method to the DelRange functions. Thanks!

mrexodia:
@codex

src/dbg/_plugins.h
```diff
@@ -246,6 +246,15 @@ typedef struct
     void* reserved;
 } PLUG_CB_STOPTRACE;
 
+/* Forward declaration because definition is in database.h */
```
mrexodia: How does this get included in the pluginsdk? The database.h file is internal to x64dbg and not distributed

src/dbg/argument.cpp
```diff
@@ -75,7 +76,12 @@ bool ArgumentAdd(duint Start, duint End, bool Manual, duint InstructionCount)
     argument.manual = Manual;
     argument.instructioncount = InstructionCount;
 
-    return arguments.Add(argument);
+    if(arguments.Add(argument))
```
mrexodia: Just had a little more time, I meant the `SerializableMap` base class. That should implement a generic database notification handler for every type.

mrexodia (changes requested):
Had a little more time for reviewing

src/dbg/_plugins.h
rafaelrfreitas: Fair point. I guess since only plugins will interact with these types we can fully define the operation struct and the enums in `_plugins.h`.

src/dbg/argument.cpp
rafaelrfreitas: Alright, we'll do that then. Thanks!

src/dbg/serializablemap.h
```diff
@@ -252,6 +278,8 @@ class SerializableTMap
 protected:
     virtual const char* jsonKey() const = 0;
     virtual TKey makeKey(const TValue & value) const = 0;
+    virtual void notifyAdd(const TValue &) const {}
```
mrexodia: Isn't it better to always require an implementation by doing `= 0` instead, or would that fail compilation currently?

src/dbg/database_cb_batcher.cpp
```diff
@@ -0,0 +1,238 @@
...
+DbCallbackBatcher* DbCallbackBatcher::Get()
+{
+    return DbCallbackBatcher::tActiveBatcher;
+}
+
+void DbCbNotifyLabel(DbOperationType opType, duint modhash, duint address, const char* text, bool manual)
+{
+    auto batcher = DbCallbackBatcher::Get();
+
+    DbOperation op = {};
```
mrexodia: Wouldn't it be cleaner to let the virtual functions fill in the `DbOperation` so that we can remove all of the specialized `DbCbNotifyXyz` functions? Then we might also be able to reduce it to a single virtual function and let the SerializableMap be responsible for the notification type/lifecycle.

src/dbg/_plugins.h
```diff
@@ -246,6 +246,49 @@ typedef struct
...
+            const char* text; // comments and labels
+        };
+
+        struct
+        {
+            duint end, instructioncount, parent; // functions, arguments and loops
+            int depth; // loops only (nesting depth)
+        };
+    };
+    duint address, modhash;
```
mrexodia: I would put this on separate lines for clarity and also probably move the `bool manual` to the bottom for better padding usage.

mrexodia (changes requested):
Getting much better, but I think it can become cleaner still.

src/dbg/_plugins.h
rafaelrfreitas: Will do!

src/dbg/serializablemap.h
rafaelrfreitas: It doesn't compile if an implementation is always required, this is due to the fact that EncodeMap, ModulePartyInfo and Xrefs implement SerializableTMap.

src/dbg/database_cb_batcher.cpp
rafaelrfreitas: Yeah, that would be much cleaner. The only caveat I can see with removing the specialized DbCbNotify functions is that `loop.cpp` still uses the old system since it doesn't implement SerializableTMap. I guess it could be possible to define a static function there that creates the DbOperation and sends it to the batcher.

For the second part of your statement, do you mean that finding a batcher object and adding an operation to the batcher would be the responsibilities of `SerializableTMap`, with the virtual function only serving to fill the correct fields (item type, etc) in a `DbOperation`?

mrexodia: Yeah for `loop.cpp` it would require some additional work, but I don't think it will be a huge amount of code if architected right.

> with the virtual function only serving to fill the correct fields (item type, etc) in a DbOperation?

Exactly! Do you foresee any issues with this approach?

src/dbg/serializablemap.h
mrexodia: I would lean towards doing = 0 anyway and just putting empty functions in those classes with a comment.

src/dbg/database_cb_batcher.cpp
rafaelrfreitas: No, it actually worked quite well, even allowing the `Remove` callbacks to carry relevant data seamlessly.

As for the `loop.cpp`, I ended up just making a static function for populating a `DbOperation` with loop info and doing the notification mechanism in the same places where they were already done.

src/dbg/serializablemap.h
rafaelrfreitas: I'll do that right now and commit.

src/dbg/database_cb_batcher.cpp
```diff
@@ -0,0 +1,85 @@
...
+    }
+}
+
+void DbCallbackBatcher::Add(DbOperation & op)
+{
+    auto batcher = DbCallbackBatcher::Get();
+
+    if(!mOwner && batcher != nullptr)
+    {
+        return batcher->Add(op);
```
mrexodia: I do not quite understand this. If there is no owner and there is already a thread-local batcher we use that batcher instead. Won't that recurse infinitely?

src/dbg/serializablemap.h
```diff
@@ -134,8 +149,29 @@ class SerializableTMap
...
+            EXCLUSIVE_ACQUIRE(TLock);
+            auto found = mMap.find(key);
+            erased = found != mMap.end();
+            if(erased)
+            {
+                value = std::move(found->second);
+                mMap.erase(found);
+            }
+        }
+        if(erased)
```
mrexodia: Here we remove the item first and then execute the callback.

src/dbg/serializablemap.h
```diff
@@ -144,7 +180,17 @@ class SerializableTMap
...
+            {
+                DbOperation op {};
+                op.opType = DbOperationType::Remove;
+
+                if(populateDbOperation(op, itr->second))
+                {
+                    DbCbNotify(op);
+                }
+
                 itr = mMap.erase(itr);
```
mrexodia: Here the callback is executed first and then the remove happens (plus the locking behavior is different here too since it is held for the whole loop vs just to delete)

src/dbg/serializablemap.h
```diff
@@ -199,7 +257,18 @@ class SerializableTMap
             deserializer.SetJson(jsonValue);
             TValue value;
             if(deserializer.Load(value))
-                addNoLock(value);
+            {
+                if(addNoLock(value))
+                {
+                    DbOperation op {};
+                    op.opType = DbOperationType::Add;
```
mrexodia: This would spam add operations on startup for every entry already in the database, is that intended?

mrexodia (changes requested):

src/dbg/database_cb_batcher.cpp
rafaelrfreitas: Yeah, it looks a bit confusing, but that situation will never happen because the thread-local `tActiveBatcher` is always guaranteed to be the owner. The `!mOwner` check evaluates the current batcher object, which might be a nested child (and thus not the owner). Once it delegates the call to the active batcher, that target instance is guaranteed to be the owner, so the check fails and it doesn't recurse.

src/dbg/serializablemap.h
rafaelrfreitas: Yes, this is the intended behavior.

src/dbg/serializablemap.h
rafaelrfreitas: My original idea was for it not to waste extra memory, but it seems like my implementation is inherently flawed. Would a better behavior be to store the deleted objects in a vector or another data structure, populate it with the lock held and when the lock is freed, simply iterate and notify?

src/dbg/serializablemap.h
rafaelrfreitas: It's not intended. In what situations does this happen? I could make `addNoLock` return whether it actually put something on the map/updated the original value or not.

src/dbg/database_cb_batcher.cpp
mrexodia: But wouldn’t it be better to let the DbNotify static function handle this (or make this function static)?

rafaelrfreitas: I agree there's an architectural flaw with the `Add` function calling itself. It would indeed be better to have `DbCbNotify` figure out which `DbCallbackBatcher` object to use. Making `Add` static would also be a good solution, since this way we could simply call `DbCallbackBatcher::Add` and remove the `DbCbNotify` function completely.

mrexodia:
Sorry for the long delay, could you do another review of the changes I pushed?

rafaelrfreitas:
Everything looks great, I didn't think about reentrant callbacks causing issues. I've pushed a PR which makes all the notifications in `loop` use the `notifyDbOperation` function for cleanliness. `CacheLoad` now also only notifies after the lock is released and the item added to match all other functions, although it still can't know if a new key was made, or if an existing object was modified (this would require comparison overloads).

mrexodia:
Hm yeah I think for the `CacheLoad` case we need to add `bool dbload` next to `bool manual` to indicate that this happened during database loading. If I understand correctly, right now if you open x64dbg and load a database with N entries already in it, there will be a callback for everything as 'added'.

Say you have a plugin that sends notifications over websocket, this could spam 1M comments every time you restart...

rafaelrfreitas:
Yes it would. A commit has been pushed that populates a `dbload` field in a DbOperation and sets it to `true` on `CacheLoad`.

What are your thoughts on the current behavior of `CacheLoad`? Is it necessary to only send an `Add` operation when either: key doesn't exist, key exists but object is different (`Remove` + `Add` or implement `Update`)?

mrexodia (approved):
CI passes on both architectures and engines. Conflict-resolution regressions and callback ordering/reentrancy issues are fixed and covered by tests.

mrexodia:
Thanks again @rafaelrfreitas! I made a few changes:
- Added a `dboplog` command to the test plugin and print every operation for debugging
- Fix a bug where `DbClear` on debuggee termination would incorrectly report everything being deleted
- Minimize overhead when nobody has registered the callback
- Introduced a separate `CB_DBLOADOPERATION` to avoid more unnecessary overhead
- Made API resistant to future changes (some follow-ups in #3909)
- Change reentrancy design, currently it uses `.swap` to keep the vector local and appends to the next batch

The final API surface looks like this:

```c
ENUM_U8_BEGIN(DbItemType)
{
    DbItemTypeBookmark,
    DbItemTypeLabel,
    DbItemTypeComment,
    DbItemTypeFunction,
    DbItemTypeLoop,
    DbItemTypeArgument,
    DbItemTypeAddressColor,
}
ENUM_U8_END(DbItemType);

ENUM_U8_BEGIN(DbOperationType)
{
    DbOperationTypeAdd,
    DbOperationTypeRemove,
}
ENUM_U8_END(DbOperationType);

#undef ENUM_U8_BEGIN
#undef ENUM_U8_END

typedef struct
{
    uint8_t reserved;
} DbItemBookmark;

typedef struct
{
    const char* text;
} DbItemLabel;

typedef struct
{
    const char* text;
} DbItemComment;

typedef struct
{
    duint end;
    duint parent;
    uint32_t icount;
} DbItemFunction;

typedef struct
{
    duint end;
    duint parent;
    uint32_t icount;
    int32_t depth;
} DbItemLoop;

typedef struct
{
    duint end;
    uint32_t icount;
} DbItemArgument;

typedef struct
{
    duint end;
    duint color;
} DbItemAddressColor;

typedef struct
{
    DbOperationType opType;
    DbItemType itemType;
    bool manual;
    duint modhash; // Use DbgFunctions()->ModNameFromHash
    duint address; // RVA if modhash != 0 else VA
    union
    {
        DbItemBookmark bookmark;
        DbItemLabel label;
        DbItemComment comment;
        DbItemFunction function;
        DbItemLoop loop;
        DbItemArgument argument;
        DbItemAddressColor addressColor;
    };
} DbOperation;

#ifdef __cplusplus
static_assert(offsetof(DbOperation, modhash) == sizeof(void*), "");
#endif // __cplusplus

typedef struct
{
    const DbOperation** operations;
    size_t count;
    uint32_t batchId; // >0 if batched, 0 otherwise
} PLUG_CB_DBOPERATION;
```

The `const DbOperation**` is to allow future changes to append members to `DbOperation` without breaking existing plugins doing `sizeof(DbOperation)` (eg to iterate the list). Also changed the union to use explicit structs for every DbItemType.

Apologies it took so long to review this properly, but the changes were very invasive and it needed a lot of development time.

rafaelrfreitas:
Hello! I've looked at the final API and it does look much cleaner, it did really need some refactoring. Turning `DbLoad` into its own callback is also something that I didn't think about, but it saves so much hassle.
I have also been busy, but I'm glad that it took this much time to be ready because it means it was well thought out.
Thanks for your help getting this merged!

PR #3875, opened 2026-06-04
Improve "SaveDatabaseInProgramDirectory" for trace file exports

g0th1c54e4:
repeat #3867


mrexodia:
I'm cool with it, but need to run `format.bat`

Issue #3876, opened 2026-06-04
Output a warning when the log file becomes too large and needs to be truncated


YuHuanTin:
### Operating System

Win10 21H2

### x64dbg Version

May 27 2026,12:09:19

### Describe the issue

When outputting logs that are too long, add a warning or additional information to indicate that the log has been truncated.
https://github.com/x64dbg/x64dbg/blob/6373153bf1b27c1a52e6dd0031881b80a9dbfc18/src/gui/Src/Gui/LogView.cpp#L360

If possible, I’ll submit a pull request right away :)

### Steps to reproduce

no

hiSandog:
It would help to define whether truncation is based on bytes, lines, or rendered blocks and whether the oldest or newest content is retained. The warning should be emitted once per truncation event and remain visible even if the affected text has already been removed. Tests with multibyte text and repeated appends near the threshold would prevent off-by-one and warning-spam behavior.

PR #3877, opened 2026-06-04
GUI: Output a warning when truncating the log buffer

YuHuanTin:
Display a warning when truncating the log buffer to better notify users and developers

https://github.com/x64dbg/x64dbg/issues/3876


mrexodia:
Thanks!

PR #3883, opened 2026-06-21
Don't overwrite explicit command line with DB-loaded one on launch

winstonwolfrce:
When a command line is explicitly provided at startup, skip loading the saved command line from the database so the explicit one is used.


mrexodia:
Thank you!

PR #3884, opened 2026-06-21
Add GuiLogGet to read log window contents as a string

winstonwolfrce:
Follows the same BridgeResult pattern as GuiGetGlobalNotes/GuiGetDebuggeeNotes: GUI_GET_LOG emits a queued signal to LogView, which allocates the log text via BridgeAlloc and signals completion synchronously via BridgeResult::GetLog. Caller frees the returned buffer with BridgeFree.

### DIFF
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -2279,6 +2279,11 @@ BRIDGE_IMPEXP void GuiFlushLog()
     _gui_sendmessage(GUI_FLUSH_LOG, nullptr, nullptr);
 }
 
+BRIDGE_IMPEXP void GuiLogGet(char** text)
+{
+    _gui_sendmessage(GUI_GET_LOG, text, nullptr);
+}
+
 BRIDGE_IMPEXP void GuiReferenceAddCommand(const char* title, const char* command)
 {
     _gui_sendmessage(GUI_REF_ADDCOMMAND, (void*)title, (void*)command);
--- a/src/bridge/bridgemain.h
+++ b/src/bridge/bridgemain.h
@@ -1393,6 +1393,7 @@ BRIDGE_IMPEXP bool DbgTypeVisit(const TYPEVISITDATA* data);
     msg(GUI_MENU_SET_NAME, int hMenu, const char* name) \
     msg(GUI_MENU_SET_ENTRY_NAME, int hEntry, const char* name) \
     msg(GUI_FLUSH_LOG, unused, unused) \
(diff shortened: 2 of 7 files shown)


mrexodia:
Why do you need this feature?

winstonwolfrce:
There are multiple use-cases:
* Using x64dbg for mass analysis/validation using AI (https://github.com/dariushoule/x64dbg-automate).
* There is a new anti-anti-debug plugin under development where `headless` is used for unit testing. During development AI also needs to know what has happened. Copy-pasting or constant redirection is annoying and time consuming. (Here the feature is: use x64dbg to find bug in x.y test)

Several plugins are writing info into log when something is blowing up or just tracing. Such case AI can auto-fetch and act on that. The alternative would be subscribe to events in `x64dbg-automate` and build the same data or constant file redirect/cleanup.

If you have better idea I would buy that.

torusrxxx:
This feature was omitted for performance reasons:
* It would be very slow to copy 100MB log like this. Of course, it's still faster than saving the log and then reading it in your application, but as soon as the log is updated, you will need to get that 100MB log again just to immediately discard the old 99 MB log.
* The best performance alternative is to redirect the log to a named pipe or file. In that way, you won't constantly get 99 MB of old logs over and over again. You just read from the pipe like a console.

mrexodia:
I think for `headless` all of the logs are redirected already: https://github.com/x64dbg/x64dbg/blob/development/src/headless/headless.cpp#L304-L308

If you want to actually use the GUI for testing, I guess the best way is to call log redirection commands on startup automatically and tell the LLM to read the logs from that file. Does that work for you?

winstonwolfrce:
> Does that work for you?

Everything works just not convenient. The source of truth must be somewhere no matter how we solve that and this is x64dbg since it stores this info already. Fetching existing info by saying not performant can be valid under some circumstances. Malware analysis and plugin development cases we're talking about Kbyte scale rather than hundreds of Mbytes. I can say for sure that from user perspective waiting couple of seconds is better than coding this around as many times as needed. If you have long fundamental consideration about this topic then I'll close this. Please notify me and we'll go that direction.

mrexodia:
In general it's a harness issue, which is the idea behind the `headless.exe` (which should also work with x64dbg-automate, just fyi). A GUI simply isn't a good platform to interact with for an agent so it's better to spend a little extra time setting up a good automation environment for the agent then to try to force the square peg in the round hole so to speak.

When it comes to the source of truth, in fact this proposed change will not work for the headless case without that message being implemented as reading the existing log file it created.

PR #3885, opened 2026-06-23
Improve db/dw/dd/dq instructions

m417z:
Allow multiple values, negative values, and packed bytes.

Previously implemented:
```assembly
db 12 ; single 0x12 byte
dw 1234 ; single 0x1234 word
```

This PR implements:
```assembly
db -2 ; one byte: 0xFE
db 12 34 ; two bytes: 0x12 0x34
dw 1234 56 ; two words: 0x1234 0x0056
db 123456 ; three bytes: 0x12 0x34 0x56
db 123 ; two bytes: 0x01 0x23
```

The packed syntax (no spaces) is only implemented for `db`.

Context: https://github.com/m417z/Multiline-Ultimate-Assembler/issues/13.

Note: The total size is still limited to 16 bytes. This also revealed an existing issue - some commands don't work, e.g. `ymmword 0`.

### DIFF
--- a/src/dbg/datainst_helper.cpp
+++ b/src/dbg/datainst_helper.cpp
@@ -161,25 +161,70 @@ bool tryassembledata(duint addr, unsigned char* dest, int destsize, int* size, c
     switch(di.type)
     {
     case enc_byte:
+        // For "db" only, a single all-hex operand longer than one byte is "packed" into
+        // several bytes, e.g. "db 11223344" (four bytes). An odd number of digits treats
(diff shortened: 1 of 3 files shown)


mrexodia:
Isn't `dw 1234 56` parsed as `dw 123456` in the command parser?

m417z:
If that was the case, `dw 12 34` would be equivalent to `dw 1234`, right? But `dw 1234` works, and `dw 12 34` fails with an error (before this PR).

mrexodia:
Thanks!

PR #3887, opened 2026-06-23
Use UTF-8 code page via exe manifest

m417z:
Reference:
https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

Fixes this issue in my plugin:
https://github.com/m417z/Multiline-Ultimate-Assembler/issues/15
And makes the life of plugin authors easier in general.

Also, a second commit fixes a duplicate manifest, see commit message for details.

### DIFF
--- a/src/dbg/_exports.cpp
+++ b/src/dbg/_exports.cpp
@@ -1188,7 +1188,7 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
         duint newStringAlgorithm = 0;
         if(!BridgeSettingGetUint("Engine", "NewStringAlgorithm", &newStringAlgorithm))
         {
-            auto acp = GetACP();
+            auto acp = BridgeGetAnsiCodePage();
             newStringAlgorithm = acp == 932 || acp == 936 || acp == 949 || acp == 950 || acp == 951 || acp == 1251;
         }
         bNewStringAlgorithm = !!newStringAlgorithm;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -993,6 +993,7 @@ endif()
 
 if(MSVC) # msvc
 	target_link_options(exe PRIVATE
+		"/MANIFEST:NO"
 		"/DELAYLOAD:user32.dll"
 		"/DELAYLOAD:wintrust.dll"
 		"/DELAYLOAD:$<TARGET_FILE_NAME:bridge>"
@@ -1153,6 +1154,12 @@ if(CMAKE_SIZEOF_VOID_P EQUAL 4) # x86
 		Shlwapi
 		Comctl32
 	)
+
+	if(MSVC) # msvc
+		target_link_options(launcher PRIVATE
+			"/MANIFEST:NO"
+		)
+	endif()
 
 	set_target_properties(launcher PROPERTIES
 		MSVC_RUNTIME_LIBRARY
--- a/cmake.toml
+++ b/cmake.toml
@@ -201,6 +201,9 @@ link-libraries = [
     "Wintrust",
 ]
 msvc.link-options = [
+    # resource.rc already embeds manifest.xml as RT_MANIFEST, so disable the
+    # linker's default manifest to avoid embedding a second one.
+    "/MANIFEST:NO",
     "/DELAYLOAD:user32.dll",
     "/DELAYLOAD:wintrust.dll",
     "/DELAYLOAD:$<TARGET_FILE_NAME:bridge>",
@@ -269,6 +272,11 @@ sources = [
     "src/exe/icon.rc",
     "src/exe/resource.rc",
     "src/exe/strings_utf8.rc",
+]
+msvc.link-options = [
+    # resource.rc already embeds manifest.xml as RT_MANIFEST, so disable the
+    # linker's default manifest to avoid embedding a second one.
+    "/MANIFEST:NO",
 ]
 link-libraries = [
     "Shlwapi",
--- a/src/bridge/bridgemain.cpp
+++ b/src/bridge/bridgemain.cpp
@@ -7,6 +7,7 @@
 #include "_global.h"
 #include "bridgemain.h"
 #include <stdio.h>
+#include <stdlib.h>
 #include <ShlObj.h>
 #include "../dbg/_dbgfunctions.h"
 #include "Utf8Ini.h"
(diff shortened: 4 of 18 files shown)


torusrxxx:
x64dbg is one of the few applications that cannot use the UTF-8 code page even though it uses UTF-8. It needs to be aware of system default locale to be compatible with ANSI debuggees,

Did you test functionalities that depend on system locale, eg. `{ansi@address}`? Can it display ANSI strings in the debuggee with the system locale especially when the system default locale is Windows-1251 and multi-byte character set?

torusrxxx:
I tested myself on Windows 11 (Chinese). I found no issues, but I still recommend you to test Russian Windows.

----
Edit: Re-built everything and retested, it is broken.

torusrxxx:
Probably also a breaking change for other plugins.

m417z:
It might be a breaking change for some plugins, but it also might be a fixing change for others which didn't think about non-English texts, like my plugin.

mrexodia:
> It needs to be aware of system default locale to be compatible with ANSI debuggees,
@torusrxxx did you confirm if `CP_ACP` will change to UTF-8 as well or does it keep using the 'The system default Windows ANSI code page.' as mentioned in MSDN?

mrexodia:
Looks like no, so we need to query the system code page on startup and use that in a few places.

ZehMatt:
Definitely no, this is going to break more than it fixes.

mrexodia:
I don't think anything will break if we handle it right. There is no ANSI API usages in x64dbg itself and internally all the char* pointers to strings are already utf-8 (including things passed to plugins). My system code page has been UTF-8 for 6+ years and never ran into any problems.

ZehMatt:
> I don't think anything will break if we handle it right. There is no ANSI API usages in x64dbg itself and internally all the char* pointers to strings are already utf-8 (including things passed to plugins). My system code page has been UTF-8 for 6+ years and never ran into any problems.

Yes but plugins aren't always using the x64dbg API only and this is a process wide change.

mrexodia:
Currently if a plugin does `CreateFileA` from some data received in a plugin callback it will just fail silently with unicode paths, changing the process code page would fix this and simplify some things. We can of course test plugins with the change, but right now I don't have the infrastructure for this...

ZehMatt:
Yes but that is a plugin issue not an x64dbg issue so we shouldn't just flip a switch that changes the behavior in some way for everything. If there is a problem with the plugin then the plugin should fix it.

torusrxxx:
The plugin should do UTF-8 conversion themselves. It has been in the documentation for years (https://help.x64dbg.com/en/latest/developers/plugins/notes.html). Considering the exact behavior likely changes with Windows version, I don't think it's a good idea to implicitly rely on that and break some systems unknowingly.

I rebuild everything and retested. It is definitely broken.

ANSI-UTF8

mrexodia:
@m417z I'm fine with the change, but at least the usages of `CP_ACP` related to string conversion need to be replaced with a global constant coming from `HKLM\SYSTEM\CurrentControlSet\Control\Nls\CodePage\ACP` to make sure the `ansi` function above works. There is also a call to `GetACP()` to determine what string detection algorithm to use per default, which also needs to be replaced.

m417z:
Done. Didn't have the time to fully test it yet.
**Edit:** I made some basic tests and it seems to be working.

mrexodia:
@torusrxxx could you also confirm it works properly now?

torusrxxx:
No, the x64dbg GUI cannot recognize correct ASCII and System codepage now, so still broken.
ANSI-UTF8

Also, many Chinese plugins might be broken, because their source code contained something like `MessageBoxA(hWnd, "😂", "😂", MB_OK);`. You didn't notice that because `MessageBoxA(hWnd, "Ok", "Ok", MB_OK);` still works, but that breaks as soon as the string contains "😂" unless you add `/utf-8` compiler flag or migrate to `MessageBoxA(hWnd, (LPCSTR)u8"😂", (LPCSTR)u8"😂", MB_OK);`. And all plugins will also need the registry hack you just added to access ANSI strings in the debuggee.

At this point I agree with [ZehMatt](https://github.com/ZehMatt). Definitely no. The current character encoding model of x64dbg is great. I can help you fix your plugin instead, target UTF-8 explicitly rather than relying on UTF-8 manifest hack.

m417z:
Thanks for trying it, I pushed a change which hopefully fixes it.

> I can help you fix your plugin instead, target UTF-8 explicitly rather than relying on UTF-8 manifest hack.

Thanks, I know how to fix it. I created this PR not because that's the easiest way to get my plugin fixed. I created it because I felt that it's a better, safer, and simpler design overall. Without a UTF-8 manifest, plugin authors have to be very careful about any string they receive from the debugger, and differentiate between these UTF-8 strings and Windows ANSI strings. That's fragile and easy to miss. My plugin's `MessageBox` usage and @mrexodia's `CreateFile` examples demonstrate that.

You refer to it as a "manifest hack", but I think that having to do manual UTF-8 to UTF-16 conversion for each string received from the debugger feels more like a hack, while the UTF-8 manifest is an elegant solution in comparison.

> Also, many Chinese plugins might be broken, because their source code contained something like `MessageBoxA(hWnd, "😂", "😂", MB_OK);`.

I agree that it's a real concern, although that means that these plugins only work correctly if the system is configured to use the right encoding, which is sub-optimal. As a mitigation, perhaps we can add `x64dbg-ansi.exe` in addition to `x64dbg.exe`, which can be used during a transition period until the affected plugins are updated.

torusrxxx:
Is it not possible to use the “System” codec anymore, and the user needs to remember the system code page?

When I change the code page in hex editor, then close the hex editor, then reopen the hex editor, the remembered code page is off by one (previously I selected “System”, but the remembered code page is UTF-16).

m417z:
Fixed the off-by-one list index, now the list stays unchanged, and existing `LastCodepage` settings keep working.

torusrxxx:
Dump view and `ansi` string formatter tests passed.

Now the question is backward compatibility. This will certainly break some plugins as well as fix some other plugins. And there's the issue of UTF-8 manifest was only introduced in Windows 1903. What happens if x64dbg is run on Windows 7, Windows 8 and Windows 10 pre-1903, should we print a warning in the log?

An upside of this change is that we can dynamically change the “ANSI” code page to UTF-8 when the debuggee uses UTF-8 manifest itself. Again, this upside has nothing to do with x64dbg UTF-8 manifest itself. Any code that deals with strings from multiple sources should be careful about exact encoding anyway, there is no shortcut especially in a debugger context.

m417z:
> What happens if x64dbg is run on Windows 7, Windows 8 and Windows 10 pre-1903, should we print a warning in the log?

These versions are mostly end-of-life, except for extended support for Win10 1607 and 1809 (until Jan 2029). There's value in supporting end-of-life versions too, but it's naturally lower priority going forward.

We can add a log message when running on these versions if there's at least one plugin installed. We can mention `x64dbg-ansi.exe` if you decide it's worth having it for some time.

> Any code that deals with strings from multiple sources should be careful about exact encoding anyway, there is no shortcut especially in a debugger context.

Right. With the UTF-8 manifest, WinAPI and the debugger agree on the encoding, so the only major source left for plugins is an ANSI debuggee. Many other things - error messages, file paths, etc. - just work.

mrexodia:
I don't want this `x64dbg-ansi.exe` that will be confusing. We should test on Windows 7 though, but I expect it will just not use UTF-8 and everything will work transparently? Otherwise I am conceptually find with the change, don't really see any downsides since we correctly detect the system locale in all cases now.

Will leave some review comments for the code itself, but otherwise I'm inclined to merge unless @torusrxxx has strong objections.

src/dbg/_exports.cpp
```diff
@@ -1188,7 +1188,7 @@ extern "C" DLL_EXPORT duint _dbg_sendmessage(DBGMSG type, void* param1, void* pa
         duint newStringAlgorithm = 0;
         if(!BridgeSettingGetUint("Engine", "NewStringAlgorithm", &newStringAlgorithm))
         {
-            auto acp = GetACP();
+            auto acp = StringUtils::AnsiCodePage();
```
mrexodia: I think we should use `BridgeGetAnsiCodePage` consistently everywhere and get rid of this extra method.

mrexodia (changes requested):

src/dbg/_exports.cpp
m417z: Done.

torusrxxx:
I have a 9 year old public plugin (https://github.com/torusrxxx/EasyDump) that will stop working as it sends ANSI strings converted into Unicode from the debuggee over a socket connection, and some vanity private plugins. It is the same for any plugin that reads ANSI strings in the debuggee or contains ANSI strings in source code. I'm not gonna update that plugin soon. But if a plugin will be updated, then they will stop working for old x64dbg versions that lacks `BridgeGetAnsiCodePage`. And none of the plugins will work correctly on both Windows 7 and Windows 11 unless they avoid ANSI strings altogether like x64dbg. Windows 7 seems to work now unless in the future the new plugin versions start to mix up ANSI and UTF-8 and break. So effectively Windows 10 the minimum requirement, and plugins targetting x64dbg version before this and x64dbg version after this are mutually exclusive.

mrexodia:
Alright I had some more time to review and think about this issue and unfortunately I will not merge this in the end @m417z I think it is more likely to cause compatibility issues than actually fix something. Sincere apologies for wasting your time 😔

torusrxxx:
Your changes could still be beneficial, when x64dbg is migrated to Linux or adds support for debuggees with UTF-8 manifest themselves. Just not add the UTF-8 manifest to x64dbg.exe that changes the global code page. May be a second one can be accepted.

PR #3889, opened 2026-06-24
Expose the process memory map over the ElfBug C API

arslan-charyyev:
## Summary

The Linux port's Memory Map feature needs the process memory layout from the engine, but ElfBug only exposed whether a region was executable. It already parsed `/proc/<pid>/maps` internally on every stop, yet kept a single `executable` bool per region and offered no way to read the list back out.

This extends the engine's internal region records to capture the full `rwxp` permissions (read/write/execute plus shared/private) and adds `ElfBugGetMemoryMap` to the C API, which copies the address-ordered region list (start, end, permissions, pathname) across the C ABI. It is the engine layer of the Memory Map vertical slice; the adapter, the Bridge shim, and the MemoryMapView widget build on top of it.

Decisions worth calling out:

- **Two-call count contract.** `ElfBugGetMemoryMap` returns the total region count and fills up to a caller-supplied capacity, with a null buffer to query the count first. This mirrors the out-parameter style of the surrounding C API and lets a caller size its buffer without guessing.
- **Fixed pathname buffer.** Each region carries its pathname inline in a fixed `char[ELFBUG_MAX_PATH]`, keeping the result a flat POD array with no ownership or lifetime concerns across the ABI.

The new Catch2 test drives the C API against a live fixture and checks the returned regions line-by-line against an independent parse of `/proc/<pid>/maps` (addresses, r/w/x and shared/private flags, pathname), plus the capacity-truncation and post-exit-cleared behavior.

### DIFF
--- a/src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
+++ b/src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
@@ -25,7 +25,7 @@ struct ElfBugDebugger : ElfBug::Debugger
     struct MemRegion
     {
         uint64_t start, end;
-        bool executable;
+        bool read, write, execute, shared;
         std::string pathname;
     };
     mutable std::mutex mapMutex;
@@ -93,7 +93,8 @@ struct ElfBugDebugger : ElfBug::Debugger
                 }
             }
 
-            newMaps.push_back({start, end, perms[2] == 'x', std::move(pathname)});
+            newMaps.push_back({start, end, perms[0] == 'r', perms[1] == 'w',
+                               perms[2] == 'x', perms[3] == 's', std::move(pathname)});
         }
         fclose(f);
 
@@ -491,7 +492,7 @@ extern "C" {
 
         std::lock_guard lock(dbg->mapMutex);
         const auto* region = dbg->findRegion(addr);
-        return region && region->executable;
+        return region && region->execute;
     }
 
     bool ElfBugMemIsValidPtr(const ElfBugDebugger* dbg, const uint64_t addr)
@@ -503,6 +504,33 @@ extern "C" {
 
         std::lock_guard lock(dbg->mapMutex);
         return dbg->findRegion(addr) != nullptr;
+    }
+
+    size_t ElfBugGetMemoryMap(const ElfBugDebugger* dbg, ElfBugMemRegion* out, const size_t maxCount)
+    {
+        if(!dbg)
+            return 0;
+        if(!dbg->active.load(std::memory_order_acquire))
+            return 0;
+
+        std::lock_guard lock(dbg->mapMutex);
+        const size_t total = dbg->memoryMap.size();
+        const size_t n = out ? std::min(total, maxCount) : 0;
+        for(size_t i = 0; i < n; ++i)
+        {
+            const auto & r = dbg->memoryMap[i];
+            ElfBugMemRegion & o = out[i];
+            o.start = r.start;
+            o.end = r.end;
+            o.read = r.read;
+            o.write = r.write;
+            o.execute = r.execute;
+            o.shared = r.shared;
+            const size_t len = std::min(r.pathname.size(), sizeof(o.path) - 1);
+            memcpy(o.path, r.pathname.data(), len);
+            o.path[len] = '\0';
+        }
+        return total;
     }
 
     bool ElfBugModBaseFromAddr(const ElfBugDebugger* dbg, const uint64_t addr, uint64_t* base)
--- a/src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
+++ b/src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
@@ -7,6 +7,7 @@
 #include <cinttypes>
 #include <cstdio>
 #include <cstring>
+#include <fstream>
 #include <mutex>
 #include <set>
 #include <shared_mutex>
@@ -53,31 +54,30 @@ struct ElfBugDebugger : ElfBug::Debugger
             return;
         }
 
-        char path[64];
-        snprintf(path, sizeof(path), "/proc/%d/maps", mProcess->pid);
-        FILE* f = fopen(path, "r");
-        if(!f)
+        std::ifstream maps("/proc/" + std::to_string(mProcess->pid) + "/maps");
+        if(!maps)
         {
             std::lock_guard lock(mapMutex);
             memoryMap.clear();
             return;
         }
 
-        char line[512];
-        while(fgets(line, sizeof(line), f))
+        // Read whole lines: a maps pathname can be longer than any fixed buffer.
+        std::string line;
+        while(std::getline(maps, line))
         {
             uint64_t start = 0, end = 0;
             char perms[8] = {};
             int pathOffset = 0;
             // %n captures the byte offset after the fixed prefix so we can take the pathname verbatim (it may contain spaces).
-            if(sscanf(line, "%" SCNx64 "-%" SCNx64 " %4s %*x %*x:%*x %*u %n",
+            if(sscanf(line.c_str(), "%" SCNx64 "-%" SCNx64 " %4s %*x %*x:%*x %*u %n",
                       &start, &end, perms, &pathOffset) < 3)
                 continue;
 
             std::string pathname;
-            if(pathOffset > 0 && pathOffset < static_cast<int>(sizeof(line)))
+            if(pathOffset > 0 && pathOffset <= static_cast<int>(line.size()))
             {
-                const char* p = line + pathOffset;
+                const char* p = line.c_str() + pathOffset;
                 while(*p == ' ' || *p == '\t') ++p;
                 size_t len = strlen(p);
                 while(len > 0 && (p[len - 1] == '\n' || p[len - 1] == '\r' || p[len - 1] == ' '))
@@ -96,7 +96,10 @@ struct ElfBugDebugger : ElfBug::Debugger
             newMaps.push_back({start, end, perms[0] == 'r', perms[1] == 'w',
                                perms[2] == 'x', perms[3] == 's', std::move(pathname)});
         }
-        fclose(f);
+        // findRegion() binary-searches this, so keep it sorted by start address
+        // rather than relying on the kernel emitting /proc/<pid>/maps in order.
+        std::sort(newMaps.begin(), newMaps.end(),
+        [](const MemRegion & a, const MemRegion & b) { return a.start < b.start; });
 
         std::lock_guard lock(mapMutex);
         memoryMap = std::move(newMaps);
--- a/src/cross/ElfBug/ElfBug/api/elfbug_api.h
+++ b/src/cross/ElfBug/ElfBug/api/elfbug_api.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <cstdint>
+#include <cstddef>
 #include <sys/types.h>
 
 #ifdef __cplusplus
@@ -75,6 +76,24 @@ ELFBUG_EXPORT bool ElfBugMemWrite(const ElfBugDebugger* dbg, uint64_t addr, cons
 ELFBUG_EXPORT bool ElfBugMemFindBaseAddr(const ElfBugDebugger* dbg, uint64_t addr, uint64_t* base, uint64_t* size);
 ELFBUG_EXPORT bool ElfBugMemIsCodePtr(const ElfBugDebugger* dbg, uint64_t addr);
 ELFBUG_EXPORT bool ElfBugMemIsValidPtr(const ElfBugDebugger* dbg, uint64_t addr);
+
+#define ELFBUG_MAX_PATH 4096
+
+// One region from /proc/<pid>/maps (snapshot refreshed on every stop).
+typedef struct
+{
+    uint64_t start;
+    uint64_t end;
+    bool read;
+    bool write;
+    bool execute;
+    bool shared;                // true = shared ('s'), false = private ('p')
+    char path[ELFBUG_MAX_PATH]; // empty for anonymous/[heap]/[stack]/... regions
+} ElfBugMemRegion;
+
+// Copies up to `maxCount` address-ordered regions into `out` (pass null to query
+// the count) and returns the total number of regions in the map.
+ELFBUG_EXPORT size_t ElfBugGetMemoryMap(const ElfBugDebugger* dbg, ElfBugMemRegion* out, size_t maxCount);
 
 // Lowest mapped start address of the module containing `addr`, or 0 if the
 // region is anonymous (heap/stack/vdso/etc).
--- a/src/cross/ElfBug/ElfBug/api/elfbug_api.h
+++ b/src/cross/ElfBug/ElfBug/api/elfbug_api.h
@@ -91,8 +91,8 @@ typedef struct
     char path[ELFBUG_MAX_PATH]; // empty for anonymous/[heap]/[stack]/... regions
 } ElfBugMemRegion;
 
-// Copies up to `maxCount` address-ordered regions into `out` (pass null to query
-// the count) and returns the total number of regions in the map.
+// Copies up to `maxCount` regions into `out` (pass null to query the count) and
+// returns the total number of regions in the map. Regions are sorted by start address.
 ELFBUG_EXPORT size_t ElfBugGetMemoryMap(const ElfBugDebugger* dbg, ElfBugMemRegion* out, size_t maxCount);
 
 // Lowest mapped start address of the module containing `addr`, or 0 if the
--- a/src/cross/ElfBug/tests/tests.cpp
+++ b/src/cross/ElfBug/tests/tests.cpp
@@ -1,14 +1,22 @@
 #include <catch2/catch_test_macros.hpp>
 #include "TestHarness.h"
 #include "SymbolHelper.h"
+#include <ElfBug/api/elfbug_api.h>
 #include <string>
 #include <chrono>
+#include <cinttypes>
+#include <condition_variable>
 #include <csignal>
 #include <cstdint>
+#include <cstdio>
+#include <cstring>
+#include <fstream>
 #include <future>
+#include <mutex>
 #include <optional>
 #include <set>
 #include <thread>
+#include <vector>
 #include <unistd.h>
 
 #define FIXTURE(name) (std::string(ELFBUG_TESTS_TARGETS_DIR "/") + (name))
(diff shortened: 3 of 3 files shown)


src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
3rdit: Not sure why this code exists (this is my fault), we should probably make it read whole lines instead of capping at 512. As a result of this line your test can flake on a long path because its truncated at 512. If you could fix it here that would be good.

src/cross/ElfBug/ElfBug/api/elfbug_api.h
```diff
@@ -76,6 +77,24 @@ ELFBUG_EXPORT bool ElfBugMemFindBaseAddr(const ElfBugDebugger* dbg, uint64_t add
...
+    char path[ELFBUG_MAX_PATH]; // empty for anonymous/[heap]/[stack]/... regions
+} ElfBugMemRegion;
+
+// Copies up to `maxCount` address-ordered regions into `out` (pass null to query
+// the count) and returns the total number of regions in the map.
+ELFBUG_EXPORT size_t ElfBugGetMemoryMap(const ElfBugDebugger* dbg, ElfBugMemRegion* out, size_t maxCount);
+
 // Lowest mapped start address of the module containing `addr`, or 0 if the
 // region is anonymous (heap/stack/vdso/etc).
 ELFBUG_EXPORT bool ElfBugModBaseFromAddr(const ElfBugDebugger* dbg, uint64_t addr, uint64_t* base);
```
3rdit, line 100: "address-ordered"? What sorts `memoryMap`? I guess the kernel gives `/proc/maps` sorted? I would remove this or actually sort it.

src/cross/ElfBug/tests/tests.cpp
```diff
@@ -55,6 +63,53 @@ namespace
         }
         return false;
     }
+
+    // Reference parse of /proc/<pid>/maps, normalized the same way the engine
+    // normalizes pathnames, so an exact line-by-line comparison is meaningful.
+    std::vector<ElfBugMemRegion> ParseProcMaps(const pid_t pid)
```
3rdit: This is nearly a copy of the engine parser in `refreshMemoryMap`. It's also very weak. The problem of this is that it's not really an independent chec k, if engine parses something wrong, this copy gets it wrong the same exact way and the test is accepted. Kind of defeating the point.

Better to parse the raw fields so a real bug will show up. And don't just pull it into one shared parser because the engine tests will just compare to itself (and will never fail).

3rdit (changes requested):

mrexodia:
```cpp
#pragma once

#include <cerrno>
#include <cinttypes>
#include <cstdint>
#include <cstdio>
#include <cstring>
#include <limits>
#include <string>
#include <string_view>
#include <type_traits>
#include <utility>
#include <vector>
#include <algorithm>

#include <unistd.h>
#include <fcntl.h>

namespace MapsParsers {
    inline bool has_prefix(const char *str, const char *prefix) {
        return strncmp(str, prefix, strlen(prefix)) == 0;
    }

    typedef bool (*LineParseFunc)(const char *line, void *user_data);

    inline bool
    parse_lines(const char *path, char *buffer, size_t buffer_size, LineParseFunc parse_line,
                void *user_data) {
        if (buffer_size < 2) {
            return false;
        }

        auto fd = open(path, O_RDONLY | O_CLOEXEC);
        if (fd == -1) {
            return false;
        }

        size_t fill_amount = 0;
        while (true) {
            ssize_t n;
            do {
                n = read(fd, buffer + fill_amount, buffer_size - fill_amount - 1);
            } while (n < 0 && errno == EINTR);

            if (n < 0) {
                close(fd);
                return false;
            }

            const bool eof = n == 0;
            if (n > 0) {
                fill_amount += static_cast<size_t>(n);
                buffer[fill_amount] = '\0';
            }

            auto cursor = buffer;
            auto end = buffer + fill_amount;
            while (cursor < end) {
                auto next_newline = static_cast<char *>(
                    memchr(cursor, '\n', static_cast<size_t>(end - cursor)));
                if (next_newline == nullptr) {
                    break;
                }

                if (next_newline > cursor && next_newline[-1] == '\r') {
                    next_newline[-1] = '\0';
                } else {
                    *next_newline = '\0';
                }

                if (!parse_line(cursor, user_data)) {
                    close(fd);
                    return true;
                }
                cursor = next_newline + 1;
            }

            const auto remaining = static_cast<size_t>(end - cursor);
            if (cursor != buffer) {
                memmove(buffer, cursor, remaining);
                fill_amount = remaining;
                buffer[fill_amount] = '\0';
            }

            if (eof) {
                // Accept a final line without a trailing newline.
                if (fill_amount != 0) {
                    if (buffer[fill_amount - 1] == '\r') {
                        buffer[fill_amount - 1] = '\0';
                    }
                    if (!parse_line(buffer, user_data)) {
                        close(fd);
                        return true;
                    }
                }
                break;
            }

            if (fill_amount == buffer_size - 1) {
                // NOTE: This happens if a single line does not fit in the buffer
                close(fd);
                return false;
            }
        }

        close(fd);
        return true;
    }

    struct Range {
        uintptr_t start = 0;
        uintptr_t end = 0;

        template<typename T>
        typename std::enable_if_t<std::is_base_of_v<Range, T>, typename std::vector<T>::const_iterator>
        static find(const std::vector<T> &ranges, uintptr_t address) {
            auto compare = [](const auto &region, uintptr_t value) {
                return region.end <= value;
            };
            auto itr = std::lower_bound(ranges.begin(), ranges.end(), address,
                                        compare);
            if (itr == ranges.end() || address < itr->start)
                return ranges.end();
            return itr;
        }
    };

    // Reference: https://man7.org/linux/man-pages/man5/proc_pid_maps.5.html
    struct MapsRange : Range {
        struct {
            bool r: 1; // readable
            bool w: 1; // writable
            bool x: 1; // executable
            bool p: 1; // private
        } perms = {};
        uint64_t offset = 0;
        uint32_t dev_major = 0;
        uint32_t dev_minor = 0;
        uint64_t inode = 0;
        std::string path;

        bool parse_line(const char *line) {
            // Extract the basic maps line information
            char temp[5] = "";
            int path_index = 0;
            if (sscanf(line,
                       "%" SCNxPTR "-%" SCNxPTR " %4s %" SCNx64
                       " %" SCNx32 ":%" SCNx32 " %" SCNu64 "%n",
                       &start, &end, temp, &offset, &dev_major,
                       &dev_minor, &inode, &path_index) != 7) {
                return false;
            }

            // Extract permissions
            perms = {};
            if (temp[0] == 'r') {
                perms.r = true;
            }
            if (temp[1] == 'w') {
                perms.w = true;
            }
            if (temp[2] == 'x') {
                perms.x = true;
            }
            if (temp[3] == 'p') {
                perms.p = true;
            }

            // Extract the path
            auto path_str = line + path_index;
            while (*path_str == ' ') {
                path_str++;
            }
            path = path_str;

            return true;
        }
    };

    struct AbstractMapsParser {
        bool parse_line(const char *line) {
            MapsRange range;
            if (range.parse_line(line)) {
                return collect(std::move(range));
            }
            return true;
        }

        static bool parse_line(const char *line, void *user_data) {
            return ((AbstractMapsParser *) user_data)->parse_line(line);
        }

    protected:
        virtual bool collect(MapsRange &&range) = 0;

    public:
        virtual ~AbstractMapsParser() = default;

        bool parse() {
            char buffer[4096] = {};
            return parse_lines("/proc/self/maps", buffer, sizeof(buffer), parse_line, this);
        }
    };

    // References:
    // - https://man7.org/linux/man-pages/man5/proc_pid_smaps.5.html
    // - https://docs.kernel.org/filesystems/proc.html
    struct SMapsRange : MapsRange {
        static constexpr size_t UNSET = std::numeric_limits<size_t>::max();

        std::string name; // TODO: do we need this?
        size_t size = UNSET;
        size_t kernel_page_size = UNSET;
        size_t mmu_page_size = UNSET;
        size_t rss = UNSET;
        size_t pss = UNSET;
        size_t shared_clean = UNSET;
        size_t shared_dirty = UNSET;
        size_t private_clean = UNSET;
        size_t private_dirty = UNSET;
        size_t referenced = UNSET;
        size_t anonymous = UNSET;
        size_t lazy_free = UNSET;
        size_t anon_huge_pages = UNSET;
        size_t shmem_pmd_mapped = UNSET;
        size_t file_pmd_mapped = UNSET;
        size_t shared_hugetlb = UNSET;
        size_t private_hugetlb = UNSET;
        size_t swap = UNSET;
        size_t swap_pss = UNSET;
        size_t locked = UNSET;
        bool thp_eligible = false;
        struct {
            bool rd: 1;  // readable
            bool wr: 1;  // writable
            bool ex: 1;  // executable
            bool sh: 1;  // shared
            bool mr: 1;  // may read
            bool mw: 1;  // may write
            bool me: 1;  // may execute
            bool ms: 1;  // may share
            bool gd: 1;  // stack segment grows down
            bool pf: 1;  // pure PFN range
            bool dw: 1;  // disabled write to the mapped file
            bool lo: 1;  // pages are locked in memory
            bool io: 1;  // memory mapped I/O area
            bool sr: 1;  // sequential read advise provided
            bool rr: 1;  // random read advise provided
            bool dc: 1;  // do not copy area on fork
            bool de: 1;  // do not expand area on remapping
            bool ac: 1;  // area is accountable
            bool nr: 1;  // swap space is not reserved for the area
            bool ht: 1;  // area uses huge tlb pages
            bool sf: 1;  // perform synchronous page faults (since Linux 4.15)
            bool nl: 1;  // non-linear mapping (removed in Linux 4.0)
            bool ar: 1;  // architecture specific flag
            bool wf: 1;  // wipe on fork (since Linux 4.14)
            bool dd: 1;  // do not include area into core dump
            bool sd: 1;  // soft-dirty flag (since Linux 3.13)
            bool mm: 1;  // mixed map area
            bool hg: 1;  // huge page advise flag
            bool nh: 1;  // no-huge page advise flag
            bool mg: 1;  // mergeable advise flag
            bool um: 1;  // userfaultfd missing pages tracking (since Linux 4.3)
            bool uw: 1;  // userfaultfd wprotect pages tracking (since Linux 4.3)
        } vm_flags = {};

        static bool extract_size(const char *line, const char *prefix, size_t &value) {
            // Skip values that are already set
            if (value != UNSET) {
                return false;
            }

            // Check the prefix
            if (!has_prefix(line, prefix)) {
                return false;
            }

            // Skip spaces
            auto prefix_length = strlen(prefix);
            auto value_str = line + prefix_length;
            while (*value_str == ' ') {
                value_str++;
            }

            return sscanf(value_str, "%zu kB", &value) == 1;
        }
    };

    class AbstractSMapsParser {
        SMapsRange range;
        bool in_range = false;

        bool flush() {
            if (!in_range) {
                return true;
            }

            auto result = collect(std::move(range));
            range = {};
            in_range = false;
            return result;
        }

        bool parse_line(const char *line) {
            auto ch = line[0];
            if ((ch >= '0' && ch <= '9') || (ch >= 'a' && ch <= 'f')) {
                SMapsRange next_range;
                if (next_range.parse_line(line)) {
                    // Flush previous region
                    if (!flush()) {
                        return false;
                    }

                    range = std::move(next_range);
                    in_range = true;
                    return true;
                }
            }

            if (in_range) {
                if (!features.smaps) {
                    return true;
                }

                // Extract the

                if (has_prefix(line, "Name:")) {
                    if (!features.name) {
                        return true;
                    }

                    auto name_str = strchr(line, ' ');
                    if (name_str != nullptr) {
                        // Skip spaces
                        while (*name_str == ' ') {
                            name_str++;
                        }
                        range.name = name_str;
                    }
                    return true;
                }
#define EXTRACT(prefix, value) if (SMapsRange::extract_size(line, prefix, value)) return true
                EXTRACT("Size:", range.size);
                EXTRACT("KernelPageSize:", range.kernel_page_size);
                EXTRACT("MMUPageSize:", range.mmu_page_size);
                EXTRACT("Rss:", range.rss);
                EXTRACT("Pss:", range.pss);
                EXTRACT("Shared_Clean:", range.shared_clean);
                EXTRACT("Shared_Dirty:", range.shared_dirty);
                EXTRACT("Private_Clean:", range.private_clean);
                EXTRACT("Private_Dirty:", range.private_dirty);
                EXTRACT("Referenced:", range.referenced);
                EXTRACT("Anonymous:", range.anonymous);
                EXTRACT("LazyFree:", range.lazy_free);
                EXTRACT("AnonHugePages:", range.anon_huge_pages);
                EXTRACT("ShmemPmdMapped:", range.shmem_pmd_mapped);
                EXTRACT("FilePmdMapped:", range.file_pmd_mapped);
                EXTRACT("Shared_Hugetlb:", range.shared_hugetlb);
                EXTRACT("Private_Hugetlb:", range.private_hugetlb);
                EXTRACT("Swap:", range.swap);
                EXTRACT("SwapPss:", range.swap_pss);
                EXTRACT("Locked:", range.locked);
#undef EXTRACT
                if (has_prefix(line, "THPeligible:")) {
                    auto value_str = strchr(line, ' ');
                    if (value_str != nullptr) {
                        while (*value_str == ' ') {
                            value_str++;
                        }
                        range.thp_eligible = *value_str == '1';
                    }
                    return true;
                }
                if (has_prefix(line, "VmFlags:")) {
                    if (!features.vm_flags) {
                        return true;
                    }

                    auto value_str = strchr(line, ' ');
                    if (value_str != nullptr) {
                        // Skip spaces
                        while (*value_str == ' ') {
                            value_str++;
                        }

                        // Split
                        auto begin = value_str;
                        auto end = begin;
                        while (true) {
                            auto ch = *value_str++;
                            if (ch == ' ' || ch == '\0') {
                                std::string_view name(begin, static_cast<size_t>(end - begin));
                                if (!name.empty()) {
#define VM_FLAG(flag) if (name == #flag) range.vm_flags.flag = true
                                    VM_FLAG(rd);
                                    VM_FLAG(wr);
                                    VM_FLAG(ex);
                                    VM_FLAG(sh);
                                    VM_FLAG(mr);
                                    VM_FLAG(mw);
                                    VM_FLAG(me);
                                    VM_FLAG(ms);
                                    VM_FLAG(gd);
                                    VM_FLAG(pf);
                                    VM_FLAG(dw);
                                    VM_FLAG(lo);
                                    VM_FLAG(io);
                                    VM_FLAG(sr);
                                    VM_FLAG(rr);
                                    VM_FLAG(dc);
                                    VM_FLAG(de);
                                    VM_FLAG(ac);
                                    VM_FLAG(nr);
                                    VM_FLAG(ht);
                                    VM_FLAG(sf);
                                    VM_FLAG(nl);
                                    VM_FLAG(ar);
                                    VM_FLAG(wf);
                                    VM_FLAG(dd);
                                    VM_FLAG(sd);
                                    VM_FLAG(mm);
                                    VM_FLAG(hg);
                                    VM_FLAG(nh);
                                    VM_FLAG(mg);
                                    VM_FLAG(um);
                                    VM_FLAG(uw);
#undef VM_FLAG
                                }
                                begin = end = value_str;
                                if (ch == '\0') {
                                    break;
                                }
                            } else {
                                end++;
                            }
                        }
                    }
                }
            }
            return true;
        }

        static bool parse_line(const char *line, void *user_data) {
            return ((AbstractSMapsParser *) user_data)->parse_line(line);
        }

    protected:
        virtual bool collect(SMapsRange &&range) = 0;

    public:
        virtual ~AbstractSMapsParser() = default;

        struct SMapsFeatures {
            bool smaps: 1;
            bool path: 1;
            bool name: 1;
            bool vm_flags: 1;

            SMapsFeatures() {
                smaps = true;
                path = true;
                name = true;
                vm_flags = true;
            }
        } features;

        bool parse() {
            range = {};
            in_range = false;

            char buffer[4096] = {};
            if (!parse_lines("/proc/self/smaps", buffer, sizeof(buffer), parse_line, this)) {
                range = {};
                in_range = false;
                return false;
            }

            // Flush the last range. A false collect() result means stop successfully.
            (void) flush();
            return true;
        }
    };
}
```

Here is some paste

src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
arslan-charyyev: Fixed in 17a09306 — `refreshMemoryMap` now reads whole lines via `std::ifstream`/`std::getline` instead of the 512-byte `fgets` buffer, so a long pathname can't be truncated (and the test can no longer flake on one).

src/cross/ElfBug/ElfBug/api/elfbug_api.h
arslan-charyyev: You're right — it was only relying on the kernel emitting `/proc/<pid>/maps` in order. In 17a09306 `refreshMemoryMap` now explicitly `std::sort`s by start address (`findRegion`'s binary search depends on it) and I reworded the comment to state that guarantee rather than assume it.

src/cross/ElfBug/tests/tests.cpp
arslan-charyyev: Good catch. Rewrote the oracle in 17a09306 so it no longer mirrors the engine: it tokenizes the line into raw whitespace fields and parses each by hand (`std::stoull` base-16 for the range, direct perm chars, `getline` for the path) — no `sscanf`/`%n` — so a bug in the engine's extraction now surfaces as a mismatch instead of being reproduced identically. The only thing shared is the engine's documented path *contract* (brackets/anon → empty, strip ` (deleted)`), expressed in a separate `ExpectedEnginePath` helper rather than a shared parser.

arslan-charyyev:
Thanks, this is really helpful — much more robust than what I have (whole-line buffering, raw `read`, the smaps fields). I kept this PR's fix minimal (whole-line `getline`) and would like to adopt `MapsParsers` as a shared parser in a follow-up rather than expand it here.

src/cross/ElfBug/ElfBug/api/elfbug_api.cpp
```diff
@@ -96,7 +96,10 @@ struct ElfBugDebugger : ElfBug::Debugger
             newMaps.push_back({start, end, perms[0] == 'r', perms[1] == 'w',
                                perms[2] == 'x', perms[3] == 's', std::move(pathname)});
         }
-        fclose(f);
+        // findRegion() binary-searches this, so keep it sorted by start address
+        // rather than relying on the kernel emitting /proc/<pid>/maps in order.
+        std::sort(newMaps.begin(), newMaps.end(),
```
mrexodia, line 101: /proc/maps is guaranteed to be already sorted by the kernel

mrexodia:
Thanks for your interest, but this type of LLM-assisted contribution is not welcome. I will update the CONTRIBUTING.md to reflect this, but I can run Claude myself and don't need to have a back-and-forth with it.

If you want to learn about building a Linux debugger I heard this is a good book btw: https://nostarch.com/building-a-debugger

3rdit:
Refer to the new: https://github.com/x64dbg/x64dbg?tab=contributing-ov-file#ai-policy

Although the time is appreciated, it is expected that **you** have some involvement in the work you make for this project. When I said AI contributions are welcome I should have been more explicit in the fact it is conditional on you having at least some involvement in what you submit, so you can learn from the contributions and stand behind them.

arslan-charyyev:
> Refer to the new: https://github.com/x64dbg/x64dbg?tab=contributing-ov-file#ai-policy
> 
> Although the time is appreciated, it is expected that **you** have some involvement in the work you make for this project. When I said AI contributions are welcome I should have been more explicit in the fact it is conditional on you having at least some involvement in what you submit, so you can learn from the contributions and stand behind them.

But I _did_ have involvement. I _did_ read every single PR feedback and spent time understanding the problem and solutions. I _did_ evaluate your proposals and I did make decisions myself, even though some of them ended up being wrong, such as not copy-pasting the entirety of mrexodia's paste. I merely let claude automate the writing of comment replies (which reflect _my_ opinion, wrong as it may be), as well as the actual code (the way I would have written if I had the time to do it by hand). Both of which are purely mechanical parts. You guys have completely misunderstood the meaning of keeping the human in the loop. In my mind it is letting the human make correct judgement calls, not monkey coding or writing prose, which appears to be your expectations.

Furthermore, I was very explicit before even starting about the _way_ in which I was able to contribute in this proposal:  https://github.com/orgs/x64dbg/discussions/3881#discussion-10286032. And I did make serious effort into planning the work before diving head first into this blindly. But instead, I kept getting contradictory signals:
- _"AI contributions are welcome"_ - **Turns out they're not.**
- _"The logical next steps for the Linux port is to create more of the cross components and set up the interface to match the Windows further."_ - **Turns out it's not, it's end-to-end feature porting.**
- _"Here is some paste"_ - **Thanks, but what do I do with it?**
- _"no it needs to be adopted as-is"_ - **Fine, but this approach contradicts with the previously establish precedent of adding component on as-needed basis**: _"the porting of the widgets should be as-necessary because otherwise the use case isn't clear and the wrong abstraction is likely to be used"_

Regardless, I express my gratitude for the learnings I gained during this brief journey and apologize for taking your time. I'm wishing you success going forward!

mrexodia:
The fact that your reply is LLM-generated shows you are completely missing the point 😅

PR #3895, opened 2026-06-30
Fix #2575: Track patches from Fill/memset command

### LINKED ISSUE #2575: Patches with Fill command are ignored
Hello mrexodia,

It's so nice to see you being so great. My problem is with 64 bit version. 

Debugger Version: Latest and all
OS: Windows 10
Description: Patch does not get the patches assembled with fill function. Also fill doesnot mark it as changed bytes. It works if assembled manually.
Reproduction Steps: Run latest x64dbg 64bit, use fill to patch some bytes in debugee and try patch.

Thank you for your interest in improving this community project!

Also exodia if you come to Kremlin come meet me. 

Regards,
VV

0x355:
Fixes #2575.

`FillDebugMemory` calls `MemWrite` directly, so bytes written via the `Fill`/`memset` command are not registered with `PatchSet` and never show up in the Patches view. The GUI `Binary -> Fill` already goes through `MemPatch` via `MemoryPage::write`, this just brings the command path in line.

Introduced in 9a1fe10a ("Move Fill out of TitanEngine").

Verified locally: before, `Fill cip, 90, 10` leaves the Patches view empty; after, it shows 16 entries.

### DIFF
--- a/src/dbg/commands/cmd-memory-operations.cpp
+++ b/src/dbg/commands/cmd-memory-operations.cpp
@@ -14,11 +14,17 @@ static bool FillDebugMemory(duint addr, duint size, uint8_t value)
     if(size == 0)
         return true;
 
+    constexpr duint patchTrackThreshold = 0x100;
+    const bool trackPatches = size <= patchTrackThreshold;
+
     std::vector<uint8_t> buffer(std::min<duint>(size, PAGE_SIZE), value);
     for(duint offset = 0; offset < size; offset += buffer.size())
     {
         const auto chunkSize = std::min<duint>(duint(buffer.size()), size - offset);
-        if(!MemWrite(addr + offset, buffer.data(), chunkSize))
+        const bool ok = trackPatches
+                        ? MemPatch(addr + offset, buffer.data(), chunkSize)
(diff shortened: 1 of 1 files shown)


mrexodia:
Hm, this was more or less intentional because you can fill a 100mb range and get an insane amount of patches.

0x355:
good point, how about a size threshold? track patches when the fill size is under 64KB (configurable), and fall back to the current MemWrite path for larger ranges with a log message explaining why patches were skipped. that covers the common use case (Fill a handful of bytes from the command bar) without flooding the Patches view on memset of a whole region
happy to implement if you agree on the approach

mrexodia:
yeah im cool with it, I think a limit of like 0x100 bytes is enough though or do you think it should be 0x1000?

0x355:
0x100 sounds good, covers the common command bar use case and keeps the gui Binary->Fill as the path for larger ranges where per byte tracking would flood the Patches view

0x355:
my bad i forgot to run the formatter

mrexodia:
all good, that's why we have CI 😉

mrexodia:
Thanks for your contribution!

PR #3898, opened 2026-07-04
Disable linker manifest generation for exe and launcher

m417z:
The non-UTF-8 fix from https://github.com/x64dbg/x64dbg/pull/3887.

---

resource.rc already embeds manifest.xml as an RT_MANIFEST resource. With the MSVC linker's default manifest generation left on, the linker embeds its own minimal manifest too, producing two RT_MANIFEST resources in the PE (the rich rc one at language 0 and the linker default at language 1033). They coexist because resource identity is (type, name, language).

The old Visual Studio projects suppressed this via GenerateManifest=false (commit a339dd2d), but that setting was lost when the projects were removed in favor of CMake (commit df110dc0). Restore it by adding /MANIFEST:NO to the link options of the two targets that embed the manifest, so only the resource.rc manifest is kept.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -993,6 +993,7 @@ endif()
 
 if(MSVC) # msvc
 	target_link_options(exe PRIVATE
+		"/MANIFEST:NO"
 		"/DELAYLOAD:user32.dll"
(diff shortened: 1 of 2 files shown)


mrexodia:
Cheers!

PR #3903, opened 2026-07-13
Fix crash in RegistersView::accessibilityValueChanged

justanotheranonymoususer:
Crashed for me on JIT launch. Let me know if you need details and I'll try to extract. I'm unable to share the dump.

### DIFF
--- a/src/gui/Src/Gui/RegistersView.cpp
+++ b/src/gui/Src/Gui/RegistersView.cpp
@@ -3512,8 +3512,11 @@ void RegistersView::accessibilityValueChanged()
                 if(a)
                 {
                     QAccessibleInterface* child = a->child(i);
-                    QAccessibleValueChangeEvent valueChangeEvent(child, QVariant(child->text(QAccessible::Value)));
-                    QAccessible::updateAccessibility(&valueChangeEvent);
+                    if(child)
+                    {
+                        QAccessibleValueChangeEvent valueChangeEvent(child, QVariant(child->text(QAccessible::Value)));
+                        QAccessible::updateAccessibility(&valueChangeEvent);
+                    }
                 }
             }
         }


mrexodia:
Seems reasonable, thanks!

torusrxxx:
It's strange, all child interfaces are supposed to be pre-created, if there is null pointer dereference then it means something else is out of bounds.

justanotheranonymoususer:
@torusrxxx I told this to my Claude. The reply:

You were right, and the dump proves it. Here's the contradiction that cracks it:

| Container | size | max key | contains K1 (148)? |
| --- |  --- |  --- |  --- |
| `mRegisterUpdates` | 20 | **148** | **YES** |
| --- |  --- |  --- |  --- |
| `mRegisterMapping` | 115 | **130** | **NO** |
| `mRegisterPlaces` | 115 | **130** | **NO** → `childCount()`\=**131** |
| `interfaces[]` | 155 | --- | all valid (`0x8000003a`..`0x800000d4`) |

`mRegisterUpdates` contains **148**, but `mRegisterMapping` **does not**. The *only* code that inserts into `mRegisterUpdates` iterates `mRegisterMapping`. So 148 cannot have been inserted under the current mapping --- **`mRegisterUpdates` is stale.**

What actually happens
---------------------

Your instinct was correct: nothing is missing from `interfaces`, and `id == 0` never happens (the lazy-create branch in `child()` is dead code). The out-of-bounds is **`i`\=148 vs `childCount()`\=131** --- and the reason `i` is 148 is a *stale set*, not a missing interface.

All of it happens inside **one** `setRegisters()` call ([:3416-3443](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/src/gui/Src/Gui/RegistersView.cpp#L3416-L3443)):

1.  **Populate** `mRegisterUpdates` by iterating `mRegisterMapping` --- which at this moment **still contains K0..K7**, because AVX-512 was being shown. K1 changed, so `mRegisterUpdates.insert(K1)` → 148 goes in.
2.  `mRegDumpStruct = converted` --- the new context lands.
3.  **`autoUpdateXMMModesAndRefresh()`** → `detectAVX512Used(&mRegDumpStruct.regcontext)` ([:2713](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/src/gui/Src/Gui/RegistersView.cpp#L2713)) tests the **Opmask (K) registers**; they're now all zero → returns **false** → `mAVX512RegistersShown` flips true→false → **`InitMappings()`** clears and rebuilds both maps ([:83-84](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/src/gui/Src/Gui/RegistersView.cpp#L83-L84)), and since `K0..K7`/`XMM16..31` are gated behind `if(mAVX512RegistersShown)` ([:683](vscode-webview://1qomlt1ktdjirjto09cjqkh7skmmds6mvcliq4sdhj7pd7ngtfk4/src/gui/Src/Gui/RegistersView.cpp#L683)), they vanish. Max key drops 154 → **130**, so `childCount()` drops 155 → **131**.
4.  **`accessibilityValueChanged()`** iterates the now-stale set, hits **148**, and `child(148)` fails `148 >= 131` → `nullptr` → `child->text()` → **crash**.

**The trigger:** the debuggee's opmask registers went **nonzero → zero** in a single step. That transition is precisely what both (a) marks K1 as changed *under the old mapping*, and (b) causes AVX-512 to be hidden *after* the change was recorded. The `i < UNKNOWN` guard is useless here --- 148 < 155 passes, but the real bound is `childCount()` = 131.

There's a second, related defect: the self-healing branch

```
else if(mRegisterUpdates.contains(itr.key()))
    mRegisterUpdates.remove(itr.key());

```

only visits keys still in `mRegisterMapping`. Once K1 is dropped from the mapping, **nothing can ever remove it** from `mRegisterUpdates` --- it sticks until the next `mRegisterUpdates.clear()` on a cip change. So this repeats on every `setRegisters()` at the same cip.

On the fix I applied
--------------------

My null check stops the crash, and it's legitimate defensive code (`QAccessibleInterface::child()` is contractually allowed to return null, so callers must check). But you're right that it **masks the real bug**: `mRegisterUpdates` is left holding a register that no longer exists in the model. The root fix is to keep it consistent when the mapping is rebuilt --- in `InitMappings()`, after the maps are repopulated:

```
// the register set may have shrunk (e.g. AVX-512 hidden); drop updates for registers that no longer exist
for(auto it = mRegisterUpdates.begin(); it != mRegisterUpdates.end();)
{
    if(!mRegisterMapping.contains(*it))
        it = mRegisterUpdates.erase(it);
    else
        ++it;
}

```

That fixes it at the source and protects every other consumer of `mRegisterUpdates`, and it covers the same class of bug for any other layout change that shrinks the register set (e.g. `mShowFpu`), not just AVX-512.

Want me to add that to `InitMappings()`? I'd keep the null check as defense-in-depth, but the staleness fix is the one that addresses what you spotted.

PR #3909, opened 2026-07-16
Add address color feature to disassembly

bycross028:
## Description

Adds customizable address highlighting in the disassembly view with persistent storage. Users can right-click addresses to apply colors from 6 predefined presets.

### Commands 
- `addresscolor <address|color>` — Set address color
- `addresscolordel [address|start end]` — Remove color from address or range
- `addresscolorrange <start> <end> <color>` — Color range of addresses
- `addresscolorclear` — Clear all address colors

### Features
- **6 Color Presets**: red, green, blue, yellow, orange, purple
- **Persistent Storage**: Colors saved across sessions in user database
- **GUI Integration**: Context menu in disassembly view and CPU disassembly
- **Graph View Support**: Colors displayed in disassembly graph
- **Appearance Dialog**: Customize color values

## Screenshot

- Light Style

light

- Dark Style

dark

- Graph View

Captura de pantalla 2026-07-15 150054

## Demo

demo

-  Commands with hotkeys

command

## Appearance Dialog

config

config

---

## Technical Notes

Developed following the architecture of the existing Comments and Markers features for consistency with current x64dbg patterns and persistence mechanism.


src/bridge/bridgemain.h
```diff
@@ -332,6 +333,7 @@ typedef enum
     DBG_SYMBOL_ENUM_FROMCACHE,      // param1=SYMBOLCBINFO* cbInfo,      param2=unused
     DBG_DELETE_COMMENT_RANGE,       // param1=duint start,               param2=duint end
     DBG_DELETE_LABEL_RANGE,         // param1=duint start,               param2=duint end
+    DBG_DELETE_LINECOLOR_RANGE,     // param1=duint start,               param2=duint end
```
mrexodia: enums have to be added at the end, otherwise all plugins will break

src/bridge/bridgemain.h
```diff
@@ -232,7 +232,8 @@ typedef enum
     flagfunction = 0x10,
     flagloop = 0x20,
     flagargs = 0x40,
-    flagNoFuncOffset = 0x80
+    flagNoFuncOffset = 0x80,
+    flaglinecolor = 0x100
```
mrexodia: you can leave a comma at the end

src/dbg/x64dbg.cpp
```diff
@@ -350,6 +350,11 @@ static void registercommands()
     dbgcmdnew("bookmarklist", cbInstrBookmarkList, true); //list bookmarks
     dbgcmdnew("bookmarkclear", cbInstrBookmarkClear, true); //clear bookmarks
 
+    dbgcmdnew("linecolor", cbDebugLineColorSet, true); //line color
+    dbgcmdnew("linecolorrange", cbDebugLineColorSetRange, true); //line color range
+    dbgcmdnew("linecolordel,linecolorclear", cbDebugLineColorDelete, true); //delete line color
+    dbgcmdnew("linecolordelrange,linecolorclearrange", cbDebugLineColorDeleteRange, true); //delete line color range
```
mrexodia: I would separate `linecolorclear` which deletes everything from `linecolordel` which can delete an arbitrary range (no need for a separate range delete command)

mrexodia (changes requested):
Overall I'm not against this idea, but I think it needs to be generalized as 'address color' because then it can also be used for the dump view to for example highlight structure members.

You will also need to wait for #3871 which I will try to merge soon.

bycross028:
## Update

**Key changes:**
- Renamed: `linecolor*` → `addresscolor*` across all APIs and commands
- Restructured commands: `addresscolordel` now accepts single address or range (no separate range command)
- `addresscolorclear` as a separate operation (clears all only)
- Moved `DBG_DELETE_ADDRESSCOLOR_RANGE` to end of enum for plugin compatibility

bycross028:
## Add address color support to Dump and Stack views

dump_stack

img

torusrxxx:
Hey, why do you think these presets (reg, green, blue, yellow, orange, purple) are good enough? Usually a six colors palette includes red, green, blue, cyan, magenta, yellow.

bycross028:
Yeah, sounds like a good idea. There's no real reason behind those specific colors. It's just what I've used for a while, and I usually stick to 3 or 4 at most, so I never rethought the set.
The one issue I've seen is that yellow and orange don't read as distinct, so that combo doesn't work well as a fixture. Your palette (red, green, blue, cyan, magenta, yellow) is a better fit, happy to go with it.

mrexodia:
If you really wanted to change something we could just name them Color 1,2,3,4,5,6 and allow an arbitrary amount just showing the squared with the color in the menu

bycross028:
Generalize address colors to a configurable arbitrary count

- Replace the fixed 6-preset enum with a configurable color list driven by AddressColorCount (new "Colors" settings section), addressing the PR review feedback for an arbitrary number of colors.
- Change the default palette from red/green/blue/yellow/orange/purple to the standard red/green/blue/cyan/magenta/yellow
- Render color menu entries as color swatches via a new makeActionColor() helper

Captura de pantalla 2026-07-20 230424

Captura de pantalla 2026-07-20 230345

mrexodia:
I’ll get to it, kinda busy at work

bycross028:
> I’ll get to it, kinda busy at work

No worries, there's no rush on my end. I'm just keeping the branch in sync with the latest development and resolving merge conflicts as they come up. The PR was accidentally closed while I was updating my fork's branch, so I reopened it. Sorry for the noise.

mrexodia:
@bycross028 finally had time to merge the database operation PR and also did a bunch of work on this PR. Currently the ranges are stored like functions/arguments in the database (start/end) instead of per-address which would use a lot of storage when a large range is created.

I'll merge once CI passes and tested locally, but would appreciate if you could test the latest version and make sure everything looks reasonable. Thanks for the contribution!

PR #3911, opened 2026-07-16
Fix QAccessibility crash in AbstractTableView

mrexodia:
This started out as a bugfix, but ended up as a complete rewrite of the accessibility support.


mrexodia:
@torusrxxx could you review this, it was a blind fix by gpt-5.5 because on macos 26 things started crashing

torusrxxx:
Seems strange, x64dbg doesn't even link to Qt 6, why GPT thinks it is related to Qt 6? Will need a closer look.

mrexodia:
The cross platform project uses Qt6 on my machine

torusrxxx:
Can you re-test with #3912 ? That is more likely to fix a crash here.

mrexodia:
Just tested, crashes in the same spot without this fix applied rebased on the latest `development`. Also FYI this doesn't crash because of the RegistersView, it crashes from an StdTable populated in a very regular way:

```cpp
void MiniMemoryMap::loadFileParser(FileParser* parser)
{
    mParser = parser;

    if(parser == nullptr)
    {
        // Clear the table
        setRowCount(0);
        reloadData();
        return;
    }

    auto regions = parser->MemoryRegions();
    setRowCount(regions.size());
    size_t row = 0;
    for(const MemoryRegion & region : regions)
    {
        setCellContent(row, ColBaseAddress, formatHex(region.BaseAddress), region.BaseAddress);
        setCellContent(row, ColRegionSize, formatHex(region.RegionSize), region.RegionSize);
        setCellContent(row, ColState, QString::fromStdString(region.State));
        if(region.AllocationBase != -1)
            setCellContent(row, ColAllocationBase, formatHex(region.AllocationBase), region.AllocationBase);
        else
            setCellContent(row, ColAllocationBase, {});
        setCellContent(row, ColProtect, QString::fromStdString(region.Protect));
        setCellContent(row, ColAllocationProtect, QString::fromStdString(region.AllocationProtect));
        setCellContent(row, ColType, QString::fromStdString(region.Type));
        setCellContent(row, ColInfo, QString::fromStdString(region.Info));

        row++;
    }
    scrollSelect(0);
    reloadData();
}
```

src/gui/Src/BasicView/AbstractTableView.cpp
```diff
@@ -1433,15 +1436,11 @@ void AbstractTableView::accessibilityTableModelChanged()
...
-        if(getViewableRowsCount() < getRowCount())
-            model.setLastRow(getViewableRowsCount() + 1);
-        else
-            model.setLastRow(getRowCount() + 1);
-        model.setLastColumn(getColumnCount());
+        // Use the QObject-based constructor for QObject-backed accessible interfaces.
+        // The QAccessibleInterface-based constructor is only safe for virtual children
+        // whose object() is nullptr; with QWidget-backed interfaces Qt 6 Cocoa may
+        // interpret the interface id as a child index when posting the notification.
+        QAccessibleTableModelChangeEvent model(this, QAccessibleTableModelChangeEvent::ModelReset);
```
torusrxxx: I saw that [object() method is called on the accessible interface](https://codebrowser.dev/qt6/qtbase/src/gui/accessible/qaccessible.h.html#_ZN32QAccessibleTableModelChangeEventC1EP20QAccessibleInterfaceNS_15ModelChangeTypeE). That method currently doesn't exist and might need to be added:

```c++
QObject* AccessibleAbstractTableView::object() const {
    return getTable();
}
```

This change removed setting the exact table size. Qt 6 defaulted to "-1". Since this information might be read by a screen reader, I think it's worth trying whether adding `object()` method can fix it. I don't have macOS and Qt 6, so I can't test myself.

mrexodia:
I had gpt-5.6-sol explore the full Qt documentation + source code + x64dbg implementation to gather information about QAccessibility and then rewrite it completely. Also added some tests for the cross-platform `minidump` tool and verified on Windows/Linux/Macos. Do you have any tests you run locally that we need to port @torusrxxx ?

torusrxxx:
No, I only do old school user tests with an actual screen reader.

mrexodia:
Just the Windows narrator or another one? I want to make sure there are no regressions, but it looks like everything is matching the qt table widgets so it should work fine.

torusrxxx:
Just Microsoft Narrator that one.

src/gui/Src/Accessible/AccessibleRegistersView.cpp
```diff
@@ -148,63 +164,101 @@ QRect AccessibleRegistersViewItem::rect() const
...
-    for(const auto & id : interfaces)
+    for(const auto id : interfaces)
     {
         if(id != 0)
             QAccessible::deleteAccessibleInterface(id);
     }
 }
 
-int AccessibleRegistersView::childCount() const
+std::vector<RegistersView::REGISTER_NAME> AccessibleRegistersView::registerOrder() const
```
torusrxxx, line 208: I think it's better to directly reorder `RegistersView::REGISTER_NAME`. I don't think the exact value of these enumerates are persisted in files and plugins.

src/gui/Src/Accessible/AccessibleAbstractTableView.cpp
```diff
@@ -265,188 +254,110 @@ int AccessibleAbstractTableView::columnCount() const
...
 {
-    if(row == 0)  // title row
-        return QString();
-    auto cell = cellAt(row, 0);
-    if(cell)
-        return cell->text(QAccessible::Value);
-    else
-        return QString();
+    Q_UNUSED(row);
+    // AbstractTableView has no visual row headers.
```
torusrxxx: The first column is used as row header because it is usually the address column or something else equivalent to a row header. The screen reader should read the address as the row header.

torusrxxx (changes requested):
The screen reader no longer reads new `EIP` address, log messages, debug state changes.

src/gui/Src/Accessible/AccessibleAbstractTableView.cpp
mrexodia: This isn't how the QTableWidget does it so I changed the behavior to match that. I think it's better to match the regular widgets to stay boring.

torusrxxx: The screen reader usually automatically announce row address when it is the title, that's good enough, better than nothing. We are not designing general-purpose `QTableWidget`.

torusrxxx:
I remembered having made sure the register changes, log messages and debug state changes are announced, and it is possible to debug with eyes closed. Now the `development` branch also doesn't announce these events, 😕

src/gui/Src/Accessible/AccessibleAbstractTableView.cpp
mrexodia: Yes it is still announced, just matches how the QTableWidget presents to the accessibility tree instead of doing something different. In my opinion the AbstractTableView is a general-purpose table widget just like QTableWidget so it makes sense to match that.

src/gui/Src/Accessible/AccessibleRegistersView.cpp
torusrxxx: The current way of managing the register names and sets is messy. I'm going to move it to somewhere where these things become static constants, when things stabilize here.

mrexodia:
Not sure, but that should be implemented on this branch now. Clanker also added some tests that can be run locally.

PR #3920, opened 2026-07-30
Fix run trace desync when thread id is written without blockFlags 0x80

nato4r:
In run-trace recording, `blockFlags` bit `0x80` must match whether a thread id payload is written. Previously the flag ignored `rtNeedThreadId` while the write path did not, so the instruction after a thread switch wrote a tid with no `0x80` bit and desynced every reader (seen as RegisterChangePosition overflow / bad block type).

Reproduced trace: [20260730-151202.trace32.txt](https://github.com/user-attachments/files/30551358/20260730-151202.trace32.txt). (Added .txt extension due to github restriction)

Unify both checks behind a single `writeThreadId` condition in `TraceRecordManager::TraceExecuteRecord`.


mrexodia:
@torusrxxx any objections?

torusrxxx:
Seems to be a regression in a8de10f4226248a20e012a51b997766a10d8b162 that introduced additional uncontrolled write to `rtNeedThreadId`. I agree that merging both conditions into one condition is easier to maintain.

Issue #3846, opened 2026-04-29
Breakpoint command with nested escaped string literals is corrupted on save


Quatern1on:
### Operating System

Windows 11 64bit

### x64dbg Version

Apr 20 2026

### Describe the issue

When a breakpoint command contains nested string literals with escaped double quotes (`\"`), x64dbg fails to handle the escaping correctly at save time. For example, setting the following command on a breakpoint:
```
mov $buf, rdx; bp [rsp], "", ss; bpcnd [rsp], "streq(ansi(buf,28),\"require('./gamemode/index');\")" 
```

Immediately upon saving, the command is stored and displayed incorrectly. Summary shows:
```
cmdif(1, "mov $buf, rdx; bp [rsp], "", ss; bpcnd [rsp], "streq(ansi(buf,28),\\require('./gamemode/index')")
```

When trying to edit it again, it is displayed back like this:
```
mov $buf, rdx; bp [rsp], "", ss; bpcnd [rsp], "streq(ansi(buf,28),\\require('./gamemode/index')
```

Which is not what I initially typed in. As a result, the condition applied to the breakpoint is also broken.

### Steps to reproduce

1. Create a breakpoint (e.g. bp kernel32.dll.ReadFile)
2. Edit breakpoint, set:
    Command Condition: 1
    Command Text: ```mov $buf, rdx; bp [rsp], "", ss; bpcnd [rsp], "streq(ansi(buf,28),\"require('./gamemode/index');\")"```
3. Press "Save"
4. Press "Edit" again, the Command Text now shows:
    ```
    mov $buf, rdx; bp [rsp], "", ss; bpcnd [rsp], "streq(ansi(buf,28),\\require('./gamemode/index')
    ```
    which is not what it initially was.

mrexodia:
Does it work when you set it like this through the UI?

Quatern1on:
Yes, if I set Break Condition to:
```streq(ansi(buf,28),"require('./gamemode/index');")```
through the UI it does work.

Issue #3507, opened 2025-01-07
Custom expression issue


dup000280886585:
### Operating System

Window 10 64bit, Build 19043

### x64dbg Version

Dec 21 2024

### Describe the issue

(一） myTest(CIP,0x5866)
        I want to customize an expression that requires 2 integer parameters, as mentioned above, which will result in an error and not execute the function
        But if we change this expression to a parameter, everything will be fine

        My code is as follows
        duint myTest(int argc, const duint* argv, void* userData)
        {
            DBGS("myTest argc=%d", argc);
            return 1000;
        }

        if (_plugin_registerexprfunction(pluginHandle, "myTest", 2, myTest, NULL))
            _plugin_logprint("regExp myTest sucess");
        else
                _plugin_logprint("regExp myTest fail");

(二) MSG {myTest(CIP, CIP)}    
     Pop up error dialog box "???" 

    All you need to do is change the parameter argc from 2 to 1, and everything will be correct
    MSG {myTest(CIP, CIP)}    
    a message dialog box 03E8 pops up with the correct result

（三) The conclusion is that expressions with more than two parameters are unfriendly and problematic

### Steps to reproduce

1. Create a plugin to register an expression

      duint myTest(int argc, const duint* argv, void* userData)
      {
          DBGS("myTest argc=%d", argc);
          return 1000;
      }
      
      if (_plugin_registerexprfunction(pluginHandle, "myTest", 2, myTest, NULL))
          _plugin_logprint("regExp myTest sucess");
      else
          _plugin_logprint("regExp myTest fail");

1. Run in X64DBG command line or script
     MSG {myTest(CIP, CIP)}   
     There may be an error here. To get the correct result, just change parameter 2 to 1. But what if there are two parameters

### Attachments

[mytest.md](https://github.com/user-attachments/files/18327285/mytest.md)

dup000280886585:
I am testing the x32dbg plugin and I am not sure if x64dbg is the same

mrexodia:
The reason for this is that the command parser uses `,` to separate arguments. To fix it you can use `msg "{myTest(CIP, CIP)}"`

dup000280886585:
> The reason for this is that the command parser uses `,` to separate arguments. To fix it you can use `msg "{myTest(CIP, CIP)}"`

That's good, thank you

Issue #3320, opened 2024-03-06
Renaming a Watch Removes Spaces


utkonos:
### Operating System

Windows Server 2022 21H2 20348.169

### x64dbg Version

Feb 19 2024

### Describe the issue

The original name of a Watch contains a space. Example: `Watch 12`. Editing the name to change the number results in an unexpected name due to the space being removed. Changing this example to `Watch 11` results in the name `Watch11`.

### Steps to reproduce

1. Create a new watch
2. Change the number of the watch to a different number
3. Observe name saved with space removed in Watch tab.

### Attachments

<img width="880" alt="image" src="https://github.com/x64dbg/x64dbg/assets/2775739/9dead408-6202-422e-a6b3-129b15e4507c">

utkonos:
Should this be committed to the development branch?

mrexodia:
Yeah after some more testing it will be merged.

PR #3922, opened 2026-08-01
Fix command argument escaping and GUI round trips

### LINKED ISSUE #3256: path in the breakpoint expression no working
### Operating System

Windows 10

### x64dbg Version

Oct 29 2023

### Describe the issue

when you edit the breakpoint，the back slash("\\") in string  is converted into an escape character("\\\\")
edit the breakpoint again,  escape character("\\\\")  is converted into "\\\\\\\\"

### Steps to reproduce

1. set breakpoints at <kernel32.dll.CreateFileW>, edit breakpoint, break condition: streq(utf16(arg(0)),"D:\FSViewer\FSIV02.fslang"}
2. save, break condition become: streq(utf16(arg(0)),"D:\\\\FSViewer\\\\FSIV02.fslang"}
3. Run debug, when reaching breakpoint, can not break and log show "Error when evaluating break condition."
4. Edit breakpoint again, do nothing and save, break condition become: streq(utf16(arg(0)),"D:\\\\\\\\FSViewer\\\\\\\\FSIV02.fslang"}

### Attachments

_No response_

### LINKED ISSUE #3921: How to use special characters in a command argument?
I want to use the `SetBreakpointCondition` command to set the condition of a breakpoint to `streq("a", "b")` .

`SetBreakpointCondition 75C6EDC0, streq("a", "b")` doesn't work because `,` is the argument separator, so I can't use `,` inside the second argument of the command.

I could put the argument in double quotes `"` like this `SetBreakpointCondition 75C6EDC0, "streq("a","b")"`. It allows me to use any special character I want inside the double quotes except for the double quote itself.

I just found out that `SetBreakpointCondition 75C6EDC0, "streq(\"a\",\"b\")"` actually works. You can escape double quotes inside a string with `\`.

I wasn't aware that this was possible because for some reason I don't need to escape a backslash `\` when I use it inside a string like a path. The backslashes in `C:\data\file.db` do not need to be escaped in `SetBreakpointCondition 75C6EDC0, "streq(\"a\",\"C:\data\file.db\")"`.

But that creates another problem. The fact that I don't need to escape a backslash and that I can not escape a backslash means that I can't produce a string that contains a path like `"C:\directory\"` (the double quotes and the backslashes are inside the string literally).

The escape system seems to be kind of flawed.

mrexodia:
## Summary

- handle runs of backslashes before quotes with Windows-style even/odd parity rules
- preserve legacy handling for ordinary Windows/UNC paths and escaped spaces/commas
- add one canonical `Command::Escape` encoder and use it for GUI-generated debugger commands
- preserve and document string-format brace behavior
- expand the general command syntax/quoting/escaping documentation
- add x86/x64 headless parser regression, round-trip, and differential compatibility tests

Closes #3921, Closes #3256

This should also address the breakpoint editor corruption reported in #3846, but that GUI flow still needs manual confirmation.

### Related compatibility history

The implementation and tests account for prior command parsing discussions and regressions, especially:

- nested expressions, commas, and quotes: #2867, #2930, #3109, #3248 / #3250, #3256, #3507
- Windows and UNC paths: #3210 / #3217
- braces in paths and string formatting: #3238 / #3281
- spaces and argument separation: #47, #417, #1024, #1469, #1918, #2085, #2263, #3320

## Differential compatibility test

The new `commandparser` test contains a frozen copy of the pre-change parser. It sends the same commands through the old reference parser and the live parser, then compares every callback argument.

- a curated compatibility corpus covers paths, UNC paths, commas, spaces, formatting expressions, escaped braces, and nested quotes
- generated cases compare backslash runs of length 0 through 6 before ordinary characters, commas, spaces, and argument endings
- intentional differences are tested separately: the #3921 case and a quoted path ending in a backslash must fail to produce the desired value with the old parser and succeed with the new parser
- canonical escaping is additionally round-tripped for representative values, every printable ASCII character, and backslash runs before quotes/braces or at argument end

## Validation

- x64 TitanEngine: 59/59 tests passed
- x64 GleeBug: 59/59 tests passed
- x86 TitanEngine: 59/59 tests passed
- x86 GleeBug: 59/59 tests passed
- `commandparser`: 822 assertions in each configuration
- x86/x64 debugger and GUI builds succeeded
- Sphinx HTML documentation build succeeded (only existing unrelated warnings)
- `git diff --check` passed

## Human review requested

This changes a central, historically under-specified parser. Please specifically review/test:

1. Whether Windows-style `2N` / `2N+1` backslash parity is the desired long-term syntax before quotes and opening braces.
2. Compatibility for scripts/plugins relying on undocumented behavior for multiple backslashes immediately before `"` or `{`.
3. Manual reproduction of #3921 in the command bar/script system.
4. Manual breakpoint editor save/edit round trips for #3846, including nested command strings and paths ending in `\`.
5. GUI launch/load flows with ordinary paths, UNC paths (#3210), paths containing braces (#3238), spaces, commas, and semicolons.
6. The new function pointer appended to `DBGFUNCTIONS` and component/plugin ABI expectations.

Given the parser's reach, snapshot/nightly exposure before a stable release would be prudent.

### DIFF
--- a/docs/commands/conditional-breakpoint-control/SetBreakpointCondition.md
+++ b/docs/commands/conditional-breakpoint-control/SetBreakpointCondition.md
@@ -6,7 +6,11 @@ Sets the software breakpoint condition. When this condition is set, it is evalua
 
 `arg1` The address of the breakpoint.
 
-`[arg2]` The condition expression.
+`[arg2]` The condition expression. Quote and escape the command argument when the expression contains commas or string literals. For example, the following condition compares against a path ending in a backslash:
+
+```
+SetBreakpointCondition 401000, "streq(utf8(rax), \"C:\directory\\\")"
+```
 
 ## result
 
--- a/docs/commands/index.rst
+++ b/docs/commands/index.rst
@@ -1,17 +1,69 @@
 Commands
 ========
 
-Commands have the following format (notice the arguments are **comma separated**):
+Syntax
+------
+
+Commands have the following format:
 
 ``command arg1, arg2, argN``
 
-FAQ:
+The first space separates the command name from its arguments. Commas, not spaces, separate arguments. Spaces outside quoted arguments are ignored, so ``log hello world`` passes ``helloworld`` as one argument. A semicolon separates multiple commands unless it appears inside a quoted argument.
 
-- Please note that all integer constants are represented in hex. For example, after executing the following command, ``$i`` will be 256 (0x100): ``mov $i, 100`` . This also means a variable cannot begin with letters from A to F.
-- Throughout this documentation, ``[arg1]`` (argument with a square bracket) represents an optional argument. ``arg1`` (argument without a square bracket) represents an mandatory argument. "[" and "]" represent memory reference operation in expression evaluation for the argument. If you don't want to refer to the content in the memory pointer, don't add "[" and "]".
-- For commands with two or more arguments, a comma (,) is used to separate these arguments. Do not use a space to separate the arguments.
-- x64dbg only supports integer in expressions. Strings, Floating point numbers and SSE/AVX data is not supported. Therefore you cannot use ``[eax]=="abcd"`` operator to compare strings. Instead, you can compare the first DWORD/QWORD of the string, or use an appropriate plugin which provides such feature.
-- The "==" operator is used to test if both operands are equal. The "=" operator is used to transfer the value of the expression to the destination.
+For example:
+
+.. code-block:: text
+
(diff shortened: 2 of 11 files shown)


mrexodia:
@codex Is this statement true? "Spaces outside quoted arguments are ignored, so ``log hello world`` passes ``helloworld`` as one argument."

mrexodia:
@codex Is this statement true? "Spaces outside quoted arguments are ignored, so log hello world passes helloworld as one argument."

Issue #3598, opened 2025-06-18
Rename `DbgCmdExec` to `DbgCmdExecAsync`


mrexodia:
I have noticed it’s confusing to people that `DbgCmdExec` does not immediately execute the command. Probably best to remove it from the API completely and provide a backwards-compatible forward.

All plugins would keep working, but recompiling them would require small changes to the code.

torusrxxx:
I don't think it is necessary to rename an API just because it is confusing for some people. The burden of updating all the code and documentation is much bigger than just improving the documentation. If there is a need to add a new API with extended features, then a new name is necessary.

mrexodia:
I think with this line of thinking we will never improve the usability of the API 😅

mrexodia:
Unfortunately it is my experience that nobody actually reads any of the documentation we write and it takes a huge amount of time 😢

torusrxxx:
> Unfortunately it is my experience that nobody actually reads any of the documentation we write

I completely agree with that for user interface. But it doesn't apply to the API interface. Why the developers know `DbgCmdExec` exists in the first place? Plugin developers are downloading the source code and trying to copy and adapt the source code, mostly because the documentation is basically non-existent. They see in our code somewhere where `DbgCmdExec` is called, and copied that.
However, the developers should be reading documentation rather than the source code. It's much more convenient because search engine naturally lead to documentation, and people don't have to clone or understand the source code before getting started. We should have that API documentation assuming people don't have access to code.

> and it takes a huge amount of time

Nowadays writing documentation can be partly automated by ChatGPT, so it's easier than ever before. No need to waste time replying here. Renaming the API in all our code is a greater waste of time.

PR #3941, opened 2026-08-25
Rename DbgCmdExec to DbgCmdExecAsync

foralost:
i did for the lolz #2

addresses #3598


mrexodia:
Thanks, but I don't think it's the right time to do this refactor 😔

PR #3944, opened 2026-08-30
Add better logging for plugin menu errors

mrexodia:
This needs some more work, but was quite useful to detect broken plugins.


goransh-walia:
This looks like a solid, focused change.

mrexodia:
Thanks, but I don't need comments like this

PR #3946, opened 2026-09-01
Fix: move RegisterView to MenuBuilder

goransh-walia:
This PR addresses #1155.

move RegisterView to MenuBuilder

Generated with AI assistance and validated against the original
file before submission (syntax check + change-scope check).
Please review carefully — happy to adjust based on feedback.


mrexodia:
You need to fix the formatting so this is actually reviewable...

PR #3948, opened 2026-09-07
ElfBug: Added Step-Over Functionality

3rdit:
This PR adds step-over (F8) to the Linux debugger. 

Completion is checked a bit stricter than Windows does it. The trap only counts if it comes from the thread that started the step, a stack pointer floor rejects hits taken in deeper frames so recursion stops in the right frame, and a call whose destination is the next instruction is recognised. A user breakpoint already sitting at the return address gets reused instead of replaced, and if the step lands on it its callback still fires and a singleshot still gets consumed.


mrexodia:
@codex review

PR #3953, opened 2026-09-08
ElfBug: all-stop mode so a reported stop freezes every thread

3rdit:
When the debugger stops for a breakpoint, step, pause or signal, it now freezes every _other_ thread in the process until you continue, instead of leaving them running.


mrexodia:
@codex

PR #3954, opened 2026-09-10
ElfBug: show signal stops, thread events and the stopping thread in the GUI

3rdit:
Adds exception, thread create and thread exit callbacks plus a current tid to the ElfBug C API - now the user can see this information cleanly.


mrexodia:
@codex

PR #3961, opened 2026-09-14
Rearm party run memory breakpoints after loader step

survivalizeed:
Now correctly reapplies the BPs after having loaded received new pages


mrexodia (commented):
Thanks for the fix and regression test — this addresses the same unconditional loader fallback reported on #3956. The base branch now has 330e57eb, so this PR currently overlaps/conflicts with that implementation. That version refreshes directly from the updated module map without inspecting the loader CIP or invoking a trace callback; only failed refreshes defer to the saved step mode.

One issue with this version: cbPartyRunStep retries installPartyRunBreakpoints on every excluded fallback instruction after a failure. I compiled this PR’s production implementation against an engine double with persistent installation failure: three excluded step completions produced three full setup attempts. A persistent protection/setup failure can therefore add a scan/install/rollback cycle to every skipped instruction. Please gate the refresh attempt so a failed refresh stays in lightweight fallback for the current traversal.

The updated base has coverage for that failure/no-retry behavior, new/unloaded target pages, cancellation, and a two-party live loader regression. I have not merged this conflicting PR.

mrexodia:
Thanks, I integrated the fix in #3956

PR #3964, opened 2026-09-19
fix restart-as-admin success detection

DosX-dev:
ShellExecuteW reports success through a return value greater than 32 and does not define GetLastError on success. Depending on stale last-error state made a successful elevated launch look like a failure, leaving the original instance open.


mrexodia:
Thanks!

PR #3965, opened 2026-09-19
validate complete FileHelper reads and writes

DosX-dev:
ReadFile and WriteFile can succeed without transferring the requested byte count. Verify the reported counts so ReadAllData and WriteAllData uphold their all-or-failure contract, and reject sizes that cannot be represented by the DWORD WriteFile API instead of silently truncating them.


mrexodia:
Did you encounter an actual failure or is this a 'belt and suspenders' situation?

DosX-dev:
I didn't encounter this in normal use; I found it during a static review.

The >MAXDWORD case is a concrete issue because the size_t value is currently cast to DWORD, so a large request would be silently truncated while the function could still report success. The byte-count checks are defensive and make the ReadAll/WriteAll contract explicit.

If you would prefer to avoid defensive changes without a reproduced failure, I can narrow the PR to the size truncation case or close it.

mrexodia:
I'm okay to merge it, doesn't really hurt. I would not consider the >4gb file write a 'concrete issue' actually, that isn't a code path actually exercised by the project (especially not in x32dbg where `sizeof(size_t) == sizeof(DWORD)` anyway)

PR #3966, opened 2026-09-19
prevent empty-pattern ReplaceAll denial of service

DosX-dev:
`ReplaceAll` repeatedly finds an empty pattern without consuming any input. A replacement with an empty search string can therefore loop forever or grow the string until the process exhausts memory. The `strreplace` expression function exposes this behavior directly to user input. Treat an empty pattern as a no-op in both narrow and wide string overloads.

For example:
```C
strreplace("a", "", "b");  //  <-  infinite loop with unbounded memory growth
```


mrexodia:
Cheers!

